PR #96, opened 2014-02-03
Removed unnecessary property with typo'd value.

Lukas-Stuehrk:
There was no space after the minus sign, so the value was never calculated correctly. Nevertheless, the unit of $gutter is 'em', so sass can't compute the value. The correct line would have been something like 'padding-left: $gutter - ((4px / $base-font-size) * 1em)'. But since nobody cared when the value was never used, I doubt that the line is necessary at all.


ericholscher:
Closing this. Seems like a small issue, and is old and out of date. If it still persists, feel free to re-open a new PR, and I will look at it.

PR #103, opened 2014-02-08
Only show Sphinx theme notice if show_sphinx is enabled

nicolaiarocci:
Make theme comply with the `html_show_sphinx` option in conf.py


ericholscher:
Merged! Thanks.

PR #128, opened 2014-06-11
made sticky nav actually stick to top and bottom even on long menus

swill:
This functionality is not 'perfect', but it is a lot better than it was.  Now the menu sticks to the left correctly.  Like before, you can only scroll the menu at the top of the page, but if you click on any menu items it moves to the top of the page usually anyway, so it works pretty well.

It was not trivial to implement exactly what I wanted to do, so this is the best I could come up with so far.  I will see if I can improve it when I have a little more time to think about it.


ericholscher:
Closed this with a manual merge.

PR #170, opened 2015-02-25
Allow github_host to be overriden

robyoung:
This is useful if you're running github enterprise on your own domain.

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -7,7 +7,7 @@
     <li>{{ title }}</li>
       <li class="wy-breadcrumbs-aside">
         {% if display_github %}
-          <a href="https://github.com/{{ github_user }}/{{ github_repo }}/blob/{{ github_version }}{{ conf_py_path }}{{ pagename }}{{ source_suffix }}" class="fa fa-github"> Edit on GitHub</a>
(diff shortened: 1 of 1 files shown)


ericholscher:
Hmm, how are you setting github_host? This should probably be documented in a more reasonable fashion, but it's a good idea.

robyoung:
I'm setting it in my `conf.py` along with the other github options (`github_user`, `github_repo` etc).

I based my approach for setting these on [this example `conf.py`](http://sphinxjpthemesbasicstrap.readthedocs.org/en/latest/options.html). Yes, documenting it somewhere would be good. I wasn't sure where best to do that. Any suggestions?

ericholscher:
Heh, funny. That's from the RTD codebase. Yea, that should work fine. I
guess the best place to doc the theme is in the README. We don't really
have official docs for it, and don't do a good job of documented non-RTD
use generally.

On Wed, Feb 25, 2015 at 10:02 AM, Rob Young <notifications@github.com>
wrote:

> I'm setting it in my conf.py along with the other github options (
> github_user, github_repo etc).
>
> I based my approach for setting these on this example conf.py
> <http://sphinxjpthemesbasicstrap.readthedocs.org/en/latest/options.html>.
> Yes, documenting it somewhere would be good. I wasn't sure where best to do
> that. Any suggestions?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/snide/sphinx_rtd_theme/pull/170#issuecomment-76019046>
> .
>



-- 
Eric Holscher
Maker of the internet residing in Portland, Or
http://ericholscher.com

Issue #172, opened 2015-03-05
Command displayed as block which makes it overflowed.


kxxoling:
Display error here: http://z42.readthedocs.org/zh/latest/mush.html#docker

![sphinx](https://cloud.githubusercontent.com/assets/1227139/6500760/08b30af4-c34b-11e4-814f-168baa8703c2.png)

It can be fix by changing tt element from block to inline-block -- change [these lines](https://github.com/snide/sphinx_rtd_theme/blob/abaf44fe0b0f10e69abb5688eaec8db80f8c7440/sass/_theme_rst.sass#L189) to:

``` sass
  tt
    @extend code
    color: $black
    display: inline-block
    big, em
      font-size: 100% !important
      line-height: normal
```

Warning: Not tested.

kxxoling:
@snide 
My code really fixed this issue, but while building multi files but sass are changed:

        modified:   sass/_theme_rst.sass
        modified:   sphinx_rtd_theme/static/css/theme.css
        modified:   sphinx_rtd_theme/static/css/theme.css.map

Should I add they into commit or not?

Blendify:
Fixed in cb3c08a

PR #173, opened 2015-03-11
Fix bug #172

kxxoling:
I fixed the bug #172 my self. It works fine in my env now.


ericholscher:
Thanks!

kxxoling:
@ericholscher You're welcome!

Issue #175, opened 2015-03-11
Compatibility with Sphinx 1.3


mhils:
Hi there,

thanks for the awesome template you're providing! I just upgrade to Sphinx 1.3 and noticed a few issues with the styling. Here's what I got on 1.2.3:
![](https://maximilianhils.com/upload/2015-03/2015-03-11_22-06-44.png)
![](https://maximilianhils.com/upload/2015-03/2015-03-11_22-07-31.png)

-------

And this is what I get on 1.3:
![](https://maximilianhils.com/upload/2015-03/2015-03-11_22-05-29.png)
![](https://maximilianhils.com/upload/2015-03/2015-03-11_22-06-03.png)

Apparently they changed the markup. It would be great if sphinx_rtd_theme would support the new one as well (or only the new one). :smiley:

ericholscher:
Thanks for the report. We'll try and get this fixed up soon.

bouzafr:
Is this patch on version 0.1.7 or not yet available?

agjohnson:
@bouzafr if you are referring to the referenced commits, it is now released as 0.1.8

bouzafr:
@agjohnson thx! all looks good now.

PR #181, opened 2015-03-27
Fix rtd theme for sphinx 1.3

danez:
In Sphinx 1.3 they changed ```<tt>``` to ```<code>``` everywhere. This PR adds ```<code>``` besides ```<tt>``` so both works. see #175


ericholscher:
Great, thanks! I was wondering if it was this simple. Have you tested the output?

danez:
I have tested it with the demo that is included in the source. Maybe there are more things that changed and need to be adapted for sphinx 1.3, but this commit fixes at least the transition they did from ```<tt>``` to ```<code>```

I need to say that SASS is generating a weird selectors after the change I did in the sass files. Like
```css
tt.descname, code.descname, tt.descclassname, code.descclassname, tt.descname, tt.descclassname {
    ...
}
```
(Not the double selectors)
Could not figure out why that happens, but the browser just ignores it.

ericholscher:
Great, thanks.

PR #182, opened 2015-03-30
Link to the external source of the page on arbitrary URL

xuhdev:
Currently only source code hosted on GitHub or BitBucket are supported to be linked. This PR adds the ability to link to an arbitrary source code host URL. To use it, html_show_sourcelink must be True, and source_url_prefix must be assigned to be the URL prefix.


ericholscher:
Thanks!

PR #183, opened 2015-04-02
update bower.json

muzuiget:
1. match the version in `sphinx_rtd_them/__init__.py` file
2. use `devDependencies` instead of `dependencies`, because we don't need wyrm in production install.


ericholscher:
Makes sense. Thanks.

PR #185, opened 2015-04-03
Remove broken "Edit on GitHub/Bitbucket" link from search pages

alexwlchan:
On a page of search results on ReadTheDocs, the "Edit on GitHub/Bitbucket" link tries to send the reader to a file "search.rst" or "search.txt" in the repository. Since the search results only exist as a template, this is a 404'd link.

This change to the template means that the "Edit on GitHub/Bitbucket" link is not shown on search pages, getting rid of this broken link.

---

The only change is the new if/endif block; the other changed lines are just matching indentation.


ericholscher:
Good point. Thanks for the PR.

PR #192, opened 2015-05-01
Change highlighted right-left padding to 1px

nikolas:
This was previously set to 6px, making it look like there were spaces between
the highlighted text and non-highlighted text. In my opinion, it looks better and
the highlight is less obtrusive when this is set to 1px.

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -141,7 +141,7 @@
     background: $yellow
     display: inline-block
     font-weight: bold
-    padding: 0 $base-line-height / 4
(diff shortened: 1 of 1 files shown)


ericholscher:
Sorry for the delay in getting back. Do you have a screenshot comparison that shows the difference here?

agjohnson:
I forgot to chime in on this. I tested this and I don't think reducing the padding looks good. Maybe there's something better to be done here.

ericholscher:
Closing this due to inactivity.

PR #194, opened 2015-05-04
Show caption elements in content as an h2 heading

agjohnson:
This just adds missing content styling for toc captions


jturner314:
This commit also affects figure captions, which makes them way too big and bold. This change should be restricted to elements in a `div` of class `toctree-wrapper`.

agjohnson:
@jturner314 good catch, could you make a ticket for this? I'll be following up on some more theme bugs this week or next

jturner314:
I created #197.

PR #195, opened 2015-05-05
Fix modernizr url

bport:
If we are in a subdirectory it's not working anymore because the path of modernizr is all time _static/...


ericholscher:
Thanks!

dghubble:
@ericholscher Is there a periodic release cycle with this fix or target date for v0.1.9?

agjohnson:
@dghubble not yet. I believe the plan at this point is to release a v1.0.0 backwards incompatible release, as per #215. I'll be working on this more this week.

PR #206, opened 2015-06-05
Don't show scrollbar if there's nothing to scroll

### LINKED ISSUE #200: Navbar scrollbar intermittently shows
Moving rtfd/readthedocs.org#1276 here

I've also noticed a scrollbar on the nav intermittently -- only on an external display 50% of the time on OSX + Chrome/Safari. Report above was on Linux + FF38 of the same.

2015-05-19-183004_310x784_scrot

untitaker:
Fix #200

Only tested with Firebug, didn't bother to install SASS.

### DIFF
--- a/sass/_theme_layout.sass
+++ b/sass/_theme_layout.sass
@@ -252,8 +252,7 @@
   left: 0
   padding-bottom: 2em
   width: $nav-desktop-width
-  overflow-x: hidden
-  overflow-y: scroll
+  overflow: auto
   min-height: 100%
   background: $menu-background-color
   z-index: $z-index-popover


untitaker:
cc @agjohnson

untitaker:
BTW this works with Chromium and Firefox on Linux.

agjohnson:
Thanks for submitting a patch! Unfortunately this isn't a full fix. As `overflow: auto` only shows a scrollbar if the content is clipped, this will still show a scrollbar on long nav content. Likewise `overflow: hidden` doesn't fix the issue for the opposite reason -- a long nav and short doc content makes a nav that doesn't scroll. I tested this on Linux FF and Chrome and reproduced the error I described with the suggested css edits.

untitaker:
I thought displaying a vertical scrollbar is actually desired when the content overflows.

agjohnson:
Ah okay -- no, that's a display bug new to 0.18. it should be hidden completely.

untitaker:
How else would you display larger menus then?

agjohnson:
It should display as-is, but with a hidden scrollbar that still allows for scrolling. The current CSS works correctly rather sporadically across browsers, and not at all on several browsers on Linux it seems.

untitaker:
I don't understand what you mean with a hidden scrollbar. Should it only show when the user hovers? Or should it scroll with the rest of the site?

agjohnson:
The scrollbar should never show.

PR #208, opened 2015-06-18
Add commit to the footer is it exists

ericholscher:
Refs https://github.com/rtfd/readthedocs.org/issues/1357


sphinx_rtd_theme/footer.html
```diff
@@ -32,5 +32,9 @@
   {% trans %}Built with <a href="http://sphinx-doc.org/">Sphinx</a> using a <a href="https://github.com/snide/sphinx_rtd_theme">theme</a> provided by <a href="https://readthedocs.org">Read the Docs</a>{% endtrans %}.
   {%- endif %}
 
+  {%- if commit %}
+  Build commit: <code>{{ commit }}</code>
```
agjohnson: So a couple issues with this seemingly small change:

 * Text placement at the end of the footer line looks awkward, I think "Build XXXX, built with Sphinx" or something more graceful would work better.
 * Is this the SHA short hash? We don't want the long hash here
 * Do we want to take the chance here to link back to the build output, as I mentioned in rtfd/readthedocs.org#1357?
 * Do we just make this the build id instead, and link back to the build page?

OR...

Put in this information in the flyout instead?

ericholscher: I'm kinda -1 on putting more info into the flyout. It's already pull of
relatively useless info, and we need to be pruning it, not adding more
debug info to it.

The main issue is that we need to support all VCS, not just git, so it's
whatever RTD is passing as the commit, which will vary by VCS. If we want a
short hash, we can either truncate it here, or make sure we're passing in
the proper hash we want.

Makes sense to link it back to the build, if we have the build id #. I
think having the commit is more generally useful than the build id, since
thats RTD specific, and the commit is used everywhere.

On Sat, Jun 20, 2015 at 3:38 PM, Anthony <notifications@github.com> wrote:

> In sphinx_rtd_theme/footer.html
> <https://github.com/snide/sphinx_rtd_theme/pull/208#discussion_r32887283>:
>
> > @@ -32,5 +32,9 @@
> >    {% trans %}Built with <a href="http://sphinx-doc.org/">Sphinx</a> using a <a href="https://github.com/snide/sphinx_rtd_theme">theme</a> provided by <a href="https://readthedocs.org">Read the Docs</a>{% endtrans %}.
> >    {%- endif %}
> >
> > +  {%- if commit %}
> > +  Build commit: <code>{{ commit }}</code>
>
> So a couple issues with this seemingly small change:
>
>    - Text placement at the end of the footer line looks awkward, I think
>    "Build XXXX, built with Sphinx" or something more graceful would work
>    better.
>    - Is this the SHA short hash? We don't want the long hash here
>    - Do we want to take the chance here to link back to the build output,
>    as I mentioned in rtfd/readthedocs.org#1357
>    <https://github.com/rtfd/readthedocs.org/issues/1357>?
>    - Do we just make this the build id instead, and link back to the
>    build page?
>
> OR...
>
> Put in this information in the flyout instead?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/snide/sphinx_rtd_theme/pull/208/files#r32887283>.
>



-- 
Eric Holscher
Maker of the internet residing in Portland, Oregon
http://ericholscher.com

agjohnson: I wouldn't want debug info in the flyout, but having a link to the build is what i was imagining in the flyout. That is, replace "Builds" with "Build", as "Builds" really isn't useful. Okay to put this in the footer though.

For VCS support, displaying the commit will only happen on RTD or maybe eventually those building docs with our extension, if we move that logic there. So, I'm happy if the truncating logic happens in RTD.

Because this patch is only useable with RTD at the moment, I'm fine linking to the build as well. If not on RTD, just degrade to no link, but display the revision if we have it.

Anyways, as for the display now, it looks hacky:
screen shot 2015-07-17 at 4 59 01 pm

I'd prefer to put this in the line above, where the last modified date would normally go, and make some changes to wording. Here's what I have when we have a build:

screen shot 2015-07-17 at 6 20 30 pm

Here's no build, and only a revision:
screen shot 2015-07-17 at 6 37 01 pm

agjohnson:
I've added #221 with my edits, to be merged into this branch.

ericholscher:
Edits look good.

ericholscher:
This is now updated with #221 & ready to merge.

PR #209, opened 2015-06-20
Fixed width in higher resolutions

rougin:
This is my fix to issue https://github.com/snide/sphinx_rtd_theme/issues/207.


ericholscher:
This was a specific design decision not to let the page width expand. It makes text harder to read as the lines get longer. If you want something that will actually fix this, at large resolutions we should likely increase the font size or something to make sure that the text line-width's stay the same proportion, not just allow the page to grow larger.

ericholscher:
More info: http://baymard.com/blog/line-length-readability

agjohnson:
Also -1 on making the content width expand, I agree the fix to this is expanding font size. High resolution displays could easily set a higher font size on `body` via media queries.

Closing this PR, feel free to submit another PR with the font size changes above.

rougin:
Thanks for this! I'll try to submit another PR for this next time.

PR #217, opened 2015-07-17
Fix citation code color, fixes #204

### LINKED ISSUE #204: ``code`` in footnotes are black while normal text is grey
In footnotes, code is displayed with a much darker color than the normal footnote text. I suggest the text color should be roughly the same. (The normal text may well be a bit darker in the footnotes, see  #110 )

agjohnson:
Mutes color on code elements in citations. This was black before and is now `$gray` -- `$gray-light` displays too faded with some fixed width fonts.

screen shot 2015-07-17 at 10 17 56 am


ericholscher:
Ugh @ conflicting checked in static assets.

PR #220, opened 2015-07-17
Fix Safari over-scroll bug, fixes #218

### LINKED ISSUE #218: Mac OS X Safari scrolling behaviour
The RTD sphinx template has problems with the Mac OS X Safari scroll behaviour. If you scroll past the top and the page bounces back the navigation is partly hidden (see the attached video).
rqmhq0idyg

agjohnson:
Protect against negative scroll values or scroll values past the end of the
document. Stop scrolling if this is the case.

### DIFF
--- a/sphinx_rtd_theme/static/js/theme.js
+++ b/sphinx_rtd_theme/static/js/theme.js
@@ -49,29 +49,30 @@ window.SphinxRtdTheme = (function (jquery) {
         var navBar,
             win,
             winScroll = false,
+            winResize = false,
             linkScroll = false,
             winPosition = 0,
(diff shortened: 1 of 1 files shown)


ericholscher:
This still feels hacky and like something we should be using a library or some supported thing for. Maintaining this with Browser changes, etc. is going to be a hassle. Perhaps we're a special case in some way, but it makes me sad to have to handle browser scrolling with javascript. Alas, browsers.

agjohnson:
yeah, if this presents more issues, maybe bringing in a library would ease some pain -- definitely a step towards bloat i would like to avoid though.

PR #212, opened 2015-07-17
Include version in the page

agjohnson:
Most doc users need to know what version of the project they are on. Add an optional show_version that will include this in the header under the main project name.


agjohnson:
Added an example. It adds a configurable version block, which will pull from conf.py or RTD context current_version. Here's what it looks like on this project:

screen shot 2015-07-17 at 4 40 17 pm

Do we want to add anything else to the version line? Is default on okay?

ericholscher:
:+1:

PR #221, opened 2015-07-18
Clean up display of revision and build on including commit

agjohnson:
This will require a patch on RTD side, to pass in `build_id` and `build_url`
in the template context.


agjohnson:
This depends on #208.

PR #226, opened 2015-07-25
Separated some sass variables

IKavanagh:
I've added some additional sass variables to _theme_variables.sass to make it easier to change some colours. The previous colours have been kept in all cases to leave the overall theme the same.


ericholscher:
This seems like a solid change, but will let @agjohnson or @snide weigh in.

snide:
I haven't tested it, but it's perfectly fine as long as it works. Specificity for coloring is harmless.

IKavanagh:
It still builds as normal with the grunt file and looks exactly the same before and after the change.

ericholscher:
:+1: on more configuration options. 

We need to document how to use all of our general configuration options. I've merged this in since we don't have an existing way of handling this, but I would love if we stole the ideas alabaster has here:

https://github.com/bitprophet/alabaster#theme-options

PR #228, opened 2015-07-28
Allow setting the nav header color via config

asakusuma:
If you have a collection of related, but distinct doc sites, it's really nice to be able to color code them without having a completely custom theme for each.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -85,7 +85,7 @@
 
     {# SIDE NAV, TOGGLES ON MOBILE #}
     <nav data-toggle="wy-nav-shift" class="wy-nav-side">
-      <div class="wy-side-nav-search">
+      <div class="wy-side-nav-search"{% if theme_nav_header_color %} style="background-color:{{ theme_nav_header_color }};"{% endif %}>
         {% block sidebartitle %}
 
         {% if logo and theme_logo_only %}


siminm:
Would love to see this, but I expected the appropriate change in the css file. Does it not work the same?

asakusuma:
Are you saying you want the config to be applied via stylesheet and not inline style?

siminm:
Right, that way it'd be applied to all places that use `#2980B9` and would be consistent. I don't know how the templating works here with css

asakusuma:
Well the purpose of this PR is not to allow people to globally change a color, but to specifically set the header color. It would be cool to allow a global change, but I'm not sure if that's possible since, as far as I'm aware, sphinx theme packages are comprised of css and not sass, and I'm not aware of a way to modify a theme css file based on the config.

asakusuma:
If it is in fact possbile and someone could explain how, I would be happy to update the PR.

snide:
So I haven't worked on this project in a bit, so I'll leave it to someone else one of the maintainers to make a decision but thought I'd chime in here.

The sass files give you essentially what you are looking for. An ability to theme the look and feel of a project using variables. Using python to change frontend is generally backwards way to do it and not the right tool for the job. Sphinx took this in the wrong direction by allowing the css_t stuff which essentially is a junk version of sass or any other preprocessor which are pretty standard tools at this point. Using config values to manage css is a very lazy (and dirty) way to solve a problem.

I generally think it's a bad idea for python code to modify the frontend when the frotend world has perfectly good ways to handle these problems.
--

To change colors in the proper way you want simply edit the sass files. Look to PR https://github.com/snide/sphinx_rtd_theme/pull/226 as an example of how coloring should be handled with the source files. But yeah, you want to edit sass files, then build css. Not edit python files to create style tags or run a secondary, un-used outside of sphinx css_t system.

asakusuma:
I agree that it's generally a bad idea for python code to modify the frontend.

However, I think changing just the header color via config is akin to setting the logo image via config. It's something super simple that allows you to individualize the site without having to do any dev work. This is how a lot of wordpress themes work, you can't just re-skin your theme through the config, but you can change a couple of key visual things, like the logo image and a header color.

Regardless, we are using my own fork to handle this, so it's not a big deal if other people don't think this is a good idea.

siminm:
@snide I'm not sure what you mean... Isn't this exactly what the [TODO](https://github.com/snide/sphinx_rtd_theme#todo) is describing?

@asakusuma I definitely want to see this, as color is the most obvious personalization on this theme

snide:
@siminm That todo actually describes what was done in this PR.

https://github.com/snide/sphinx_rtd_theme/pull/226

agjohnson:
I'm -1 on this too. This would be better left off to css overrides.

agjohnson:
Closing this out due to inactivity and because of the issues raised above -- I don't think this PR is mergable. I think this use case is better served with the other options that were outlined above and I wouldn't want to start implementing changes as inline CSS styles.

PR #241, opened 2015-09-08
Add block to add extra content to footer

leplatrem:
For example, in order to add Disqus comments.

(*I must admit I had trouble finding a name for the block...*)


agjohnson:
I'm +1 on more block structure, it makes customizing far easier.

The name `extra` is too generic though, I feel. Perhaps 'extrafooter', which is consistent with `extrahead` block.

leplatrem:
Thanks for your feedback! I updated accordingly.

agjohnson:
Great, thanks!

PR #243, opened 2015-09-09
Make theme into a consumable bundle for RTD

agjohnson:
This makes the theme javascript modular, so the Read the Docs doesn't need to fork the code or assets internally. In particular, this:

* Makes javascript a commonjs module
* Module exports everything as public members, instead of hiding the implementation. This allows RTD to override only the necessary pieces on the instance
* Cleans up implementation to be more composable
* Creates a javascript source file and a bundle output. RTD will not use the bundled output, but others may still. It should work the same as the module does now.
* Browserify creates the bundle. This is mostly for future design -- it was left in after an original test to move dependencies into the js bundle.
* jquery, underscore, and docutil javascript is still included via the normal script tags for now.
* Installs fonts from bower now. RTD installs directly from these bower packages as well.

See rtfd/readthedocs.org#1645 for more info on implementation there.


gregmuellegger:
Local test successful with a local installation of rtfd.org. JS is picked up successfully, sticky nav etc works. Code also looks good.

:+1: from me.

ericholscher:
:+1: -- Are we waiting on a release to do these merges with a version bump?

agjohnson:
I'll wrap up this milestone and get a release ready. Once this is out rtfd/readthedocs.org#1645 will need some final changes to reference the tagged release

PR #248, opened 2015-09-16
Remove modernizr.js, we are not using it

### LINKED ISSUE #245: Modernizr is not used
Hi, I wonder why does RTD theme use `modernizr.min.js` file.

I tried grepping JS files for usage of `Modernizr` object, and SASS files for typical Modernizr `.foo` and `.no-foo` classes, but could find neither of them.

Is Modernizr really used anywhere? If the answer is no, then I'll be happy to provide a pull request to get rid of it.

mitya57:
Also remove no longer needed classes from root `html` element.

Fixes #245.


agjohnson:
I'd prefer to see this tested thoroughly on ie8 and ie9. I don't think it's fair to say that we are not using modernizr here, as this seems to be pulling out the actual modernizr test classes that were in use -- that is `lt-ie9` and `no-js`.

mitya57:
Yes, I pulled out `lt-ie9` because it was only *defined*, but not actually *used*.

agjohnson:
The css classes aren't used, but I believe modernizr was added to support html5 in ie8, replacing html5shiv.

mitya57:
OK, I tested with IE8 and looks like my pull request breaks things. Let's return to it when IE8 is no longer supported :)

mitya57:
@agjohnson Now that Microsoft stopped supporting IE < 10 more than two months ago, will you be OK with merging this?

agjohnson:
Microsoft's support of IE doesn't shouldn't concern our usage. Looking at RTD traffic, IE < 10 makes up .4% of our traffic. What is the upside to removing support for modernizr? The focus of this theme is cross platform/device support, dropping this for no good reason will give these users a broken experience. For that, I'm still -1 on removing modernizr unless there is good reason.

mitya57:
If you wanted to look at the usage, you need to look only at IE 8, IE 9 works fine without modernizr. I'm the Ubuntu packager and not having the modernizr dependency will make life a bit easier for us. But having it is also fine, we can workaround that in our packaging.

agjohnson:
Roger. IE8 is a much smaller portion of our traffic. I'd want to compare html5 support against our traffic patterns before we call off support for modernizr. It's been a few years since the theme was created, so the html5 support landscape has certainly shifted. IE and mobile are probably my larger concerns, though browsers without html5 support there are also both a fraction of our traffic.

One option would be to deprecate modernizr here and move the dependency to RTD, where we're most concerned about cross-platform support.

PR #255, opened 2015-10-10
Make bar icon vertically centered in the nav menu in mobile view

zhanghongyi:
Currently, the bar icon of the top menu bar is not vertically centered on mobile devices. 
screen shot 2015-10-10 at 15 15 36

My change is intended to add the top paddings to it so that the element will be set in the middle. This is the easiest fix I can think of. And it is tested on a few mobile sizes. 

Btw, I am not sure if the compiled "map" files changes should be included in the pull request. Please let me know if this is not the right way and I will add them. :)


ericholscher:
Thanks!

PR #257, opened 2015-10-16
insert courtesy newline to <pre> elements

devel0:
I added a courtesy newline to pre elements when last line overflow the view with scrollbar.
courtesy-newline

I would to know if there are a way to enable a "copy to clipboard" button for generated pre elements. I'm not an expert in javascript but I suspect the select-all text or the copy to clipboard can work in some browser and in others not.

How it works:
- entering the pre element if last line overflow the view a pre space element is inserted to allow easy select
- when mouse goes out the pre is removed.

Optionally ( not yet implemented ) :
- enable of the feature in the layout.html through a config variable
- disable the auto-removal of the pre element if previously inserted ( to avoid too much movements of the page content ) when the mouse exits the pre content.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -192,5 +192,7 @@
 
   {%- block footer %} {% endblock %}
 
+  <script src="{{ pathto('_static/js/util.js', 1) }}"></script>
+  <script type="text/javascript">processPre();</script>
 </body>
 </html>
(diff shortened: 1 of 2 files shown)


agjohnson:
Correct, copy to clipboard doesn't work cross browser, most sites use flash (nopenopenope) to perform this. I don't like this implementation though, adding a line underneath is purely functional, doesn't look good, and is strange UX behavior.

If anyone has an other thoughts on how to implement this sanely, feel free to open a ticket to discuss.

devel0:
I noticed that using three-clicks entire text gets selected but not in all cases, expecially when it longer than one page.

PR #261, opened 2015-11-18
Allow for favicons generated by realfavicongenerator.net

JelteF:
This website generates favicons for all kinds of different platforms: http://realfavicongenerator.net/

I thought it would be nice to have support for it. What I did was use the html generated by the website when the favicon string is `"_static/realfavicongenerator.ico"`. In this case the favicons directory will be used as a base for these files. 

I guess some stuff should be added to the docs as well to explain this change, since it sort of hacks into a sphinx feature, which makes it very non obvious. But I first wanted to know if you guys would actually like this in the theme before I put any effort into that.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -19,7 +19,29 @@
   {% endblock %}
 
   {# FAVICON #}
-  {% if favicon %}
+  {% if favicon == 'realfavicongenerator.ico' %}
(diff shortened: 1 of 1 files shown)


agjohnson:
Thanks for the submission, but this seems like an edge case that no one will maintain. Is this not possible through theme overrides?

ericholscher:
You should be able to override this with a `layout.html` in `templates_path` -- http://www.sphinx-doc.org/en/stable/config.html#confval-templates_path

PR #268, opened 2015-11-27
Add a body_begin block at the start of <body>

leoshaw:
This makes it easy to add a cookie banner or similar at the top of the
page, particularly when JS is disabled and you can't insert the banner
dynamically.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -81,6 +81,8 @@
 
 <body class="wy-body-for-nav" role="document">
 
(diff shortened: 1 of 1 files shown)


ericholscher:
This looks good, however it shouldn't use the `_`, as none of the other blocks in the page use this.

agjohnson:
Also, for consistency, there is `extrahead`, maybe `extrabody` makes more sense.

ericholscher:
Yea, going to merge this, and fix the name manually.

PR #275, opened 2016-01-10
Drop dependency on Sphinx

### LINKED ISSUE #189: Relaxing sphinx requirement
I know this may come across as a strange request, but is there a way of relaxing and/or removing the sphinx requirement. This is causing some of us build trouble with sphinx ( https://github.com/sphinx-doc/sphinx/issues/1845 ).

mitya57:
As suggested by @birkenfeld in #189, this is the easiest way to solve the dependency loop that is creating problems for packagers. Fixes #189.


ericholscher:
Thanks!

jakirkham:
Thanks guys. And this just works for you without doing anything else?

mitya57:
@ericholscher: Thanks for merging it!

@jakirkham: It works for us in Debian (we have been shipping this patch for quite some time now), though I don't quite understand what your question is about.

jakirkham:
@mitya57, ok, I was just asking because I didn't know if there were imports that needed to be changed or other things. However, it doesn't look like `sphinx` was being imported anywhere.

mitya57:
Right, sphinx is not imported anywhere.

Issue #93, opened 2014-01-30
Tables are too wide


clee704:
I want my tables not to have a horizontal scrollbar. When I build the docs locally with sphinx-rtd-theme==0.1.5, it's fine. Contents in the table cells are properly wrapped. But [on the readthedocs.org](http://cpucoolerchart.readthedocs.org), it gets too wide and I think it hurts the UX.

You can build the docs locally by cloning [my project](https://github.com/clee704/cpucoolerchart) (assuming you are using Python 2.7):

```
$ git clone https://github.com/clee704/cpucoolerchart
$ cd cpucoolerchart
$ pip install -r dev-requirements.txt
$ make docs
```

PR #276, opened 2016-01-28
Fix tables scrolling

ClemPi:
This patch should fix scrolling issue with tables (#259) keeping mobile & tablets behavior.


agjohnson:
@snide has raised issue with this approach several times (#93 et al), though I'm not sure precisely what additional problems he ran into when allowing cells to wrap. This should be tested thoroughly, and against some test projects with wide tables.

sass/_theme_rst.sass
```diff
@@ -298,9 +298,25 @@
...
+      td
+        white-space: normal
+
+
+// Tablet specific
++media($tablet)
+  .wy-table-responsive
+    table
+      td
+        white-space: nowrap !important
```
agjohnson: We shouldn't need this block twice, and the `!important` flag should be avoided. Ordering this correctly will solve this issue.

ericholscher:
Closing this, as it never got any update from the intital feedback. Thanks for the patch, though!

PR #279, opened 2016-02-11
Add a styling for `:guilabel:` inline markup.

### LINKED ISSUE #262: guilabel role CSS style
The [guilabel role](http://sphinx-doc.org/markup/inline.html#role-guilabel) is displayed as italic by  the default Sphinx theme. Unfortunately, the RTD theme does not apply any emphasis, so it appears as normal text.

For instance, see "Owning group" in

http://docs.nethserver.org/en/latest/shared_folder.html#access-privileges 

```
<span class="guilabel">Owning group</span>
```

Is it possible to retain the default Sphinx style?

jonathanj:
Fixes #262.

`:guilabel:` is a convenient way to hint that a particular action in a user interface is available, unfortunately in the RTD theme there is no styling for it, so the markup appears as plain text.

screen shot 2016-02-11 at 11 16 30

[Review on Reviewable](https://reviewable.io/reviews/snide/sphinx_rtd_theme/279)


agjohnson:
Thanks for the addition! In terms of styling, here's a few things:

* The border box looks like it extends past the `line-height`, this will cause paragraph lines to overlap or look ragged.
* There should be a bit more margin on sides of the highlight
* I'm hesitant to force all caps, though if it's a pattern we're using on elements in the same way, I think it makes sense. We force caps on nav elements, I don't recall immediately where else we might

jonathanj:
Thanks for the feedback @agjohnson!

* The border box was basically on the edge of `line-height`, I've shrunk it down so that it fits more comfortably.
* I added a small horizontal margin. I'm not sure what variable / ratio to base this off, I used `base-line-height` but it seems like a letter spacing variable would be better. Is there one?
* I don't have a strong opinion on the capitals, it was mostly a personal style preference but I think they're a little clearer (there's no letter ambiguity like there can be in lowercase) and stand out more. I don't see a pattern in the use of all caps throughout the theme, why are the section captions in the sidebar capitals, for example?

Some screenshots again, on my particular resolution these two words appear right underneath one another, so it's a good illustration for the border box size.

screen shot 2016-02-15 at 23 20 19
screen shot 2016-02-15 at 23 19 42

Makc37:
Reviewed 1 of 5 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/snide/sphinx_rtd_theme/279)*

jonathanj:
@Makc37 I've resolved the conflicts with master.

ericholscher:
Seems like this can be merged. @agjohnson any objection?

sass/_theme_rst.sass
```diff
@@ -298,6 +298,15 @@
       @extend .fa-download
       &:before
         margin-right: 4px
+  .guilabel
+    border: 1px solid lighten($class-color, 25%)
+    background: lighten($class-color, 50%)
```
agjohnson: This shouldn't piggy back off of `class-color` and should set it's own color, or use something more appropriate than `class-color`.

ericholscher:
Fixed this in a new branch and merged.

PR #285, opened 2016-03-25
Add support for canonical URLs

lafrech:
Same as https://github.com/snide/sphinx_rtd_theme/pull/284 except this time I ran `grunt build` as requested in https://github.com/snide/sphinx_rtd_theme/blob/master/README.rst#before-you-send-a-pull-request.


sphinx_rtd_theme/layout.html
```diff
@@ -22,6 +22,10 @@
   {% if favicon %}
     <link rel="shortcut icon" href="{{ pathto('_static/' + favicon, 1) }}"/>
   {% endif %}
+  {# CANONICAL URL #}
+  {% if theme_canonical_url %}
+    <link rel="canonical" href="{{ theme_canonical_url }}{{ pagename }}.html"/>
```
ericholscher: This will break on the HTML Directory builder that doesn't use this specific page layout.

lafrech:
Thanks @ericholscher for the feedback.

I don't understand. (Consider me as a total noob, I'm pretty new to the subject.)

Anyone finds interest in this feature, hints appreciated.

Or maybe I should close this and open an enhancement request.

sphinx_rtd_theme/layout.html
rozza: @ericholscher what layout does that use?

ericholscher:
We need to add logic similar to https://github.com/rtfd/readthedocs-sphinx-ext/blob/master/readthedocs_ext/_templates/readthedocs-insert.html.tmpl#L4-L11 in this patch. We can't move that logic from RTD here, because we need to support other themes.

I'm not 100% sure we need to add this to the theme, as this is something people can pretty easily do in override templates. It does seem like a reasonable feature, if someone wants to add the above logicl

ericholscher:
Looks like `alabaster` is doing the same thing (https://github.com/bitprophet/alabaster/blob/42eaa8e8ebfa725ef5c79de95dd42f6a8fb6b6d0/alabaster/layout.html#L9). I'll go ahead and merge this

PR #297, opened 2016-05-10
Change README.rst code order of conf.d 

bdh1011:
Change the order of two lines.
`html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]`

`html_theme = "sphinx_rtd_theme"`

It could return error when .get_html_theme_path() method called later than theme assignment like below.

`Theme error:`

`no theme named 'sphinx_rtd_theme`' found (missing theme.conf?)`

`make: *** [html] Error 1`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,9 +45,9 @@ In your ``conf.py`` file:
 
     import sphinx_rtd_theme
 
-    html_theme = "sphinx_rtd_theme"
-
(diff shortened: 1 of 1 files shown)


ericholscher:
I can't reproduce this issue locally -- and in fact have seen this work in lots of projects. I'm closing this as I believe it's an issue with your local setup.

PR #299, opened 2016-05-30
Fix #258: Avoid large logo to overflow

### LINKED ISSUE #258: Extra large logo does not scale to fit inside logo box
If the logo is 3333x3333 for example, it will be displayed in full size, hiding behind the main body.

Changing [this line](https://github.com/snide/sphinx_rtd_theme/blob/8c880a64fa0f08785ffc560708628fc3108a2f3d/sass/_theme_layout.sass#L206) to `100%` should fix that.

romain-dartigues:
Hi,

I am having the same issue as in #258 (in my case the logo is in SVG format and used at multiple place, so it is not relevant to create a resized version of the file).

However I suggest a subtle different change than @vlttnv as setting `img.logo` to 100% will enlarge small images and make them ugly-looking.

### DIFF
--- a/sass/_theme_layout.sass
+++ b/sass/_theme_layout.sass
@@ -196,6 +196,7 @@
     display: inline-block
     padding: $base-line-height / 6 $base-line-height / 4
     margin-bottom: $gutter / 2
+    width: 100%
     +font-smooth
     &:hover
       background: rgba(255,255,255,.1)


ericholscher:
@agjohnson thoughts? Seems like a reasonable change. I'd also love to natively support putting a logo in the theme with the standard Sphinx options. 

@romain-dartigues sorry this has been open so long, but do you have an example of this failure or fix anywhere?

romain-dartigues:
@ericholscher < actually I found that sphinx-doc/sphinx@09544fc is supposed to prevent the issue; after digging in history to find what was going wrong at the time for @vlttnv and me I found I can not reproduce the problem starting Sphinx 1.2.1.

I'm not sure which version I was using at the time, since Sphinx 1.2.1 was released in January 2014 and at the time we reported our frustration your [`requirements.txt`](https://github.com/snide/sphinx_rtd_theme/blob/c4c681bc0e2a75035e100eae3d56d84d5f7562cd/requirements.txt) was requiring for Sphinx ≥ 1.3.

I suppose you can close #258 and #299; but I still advise favouring `max-width` instead of `width` when possible.

Thank you for your follow-up,
cheers.

ericholscher:
Great, thanks for following up as well! 🎆

PR #302, opened 2016-06-21
Fix how we check for meta and make sure it isn't None

ericholscher:


agjohnson:
Looks good. Targeting a 0.1.10 release with this.

PR #309, opened 2016-07-19
Add option for navigation depth

### LINKED ISSUE #277: Option to limit the depth in the left sidebar
Hi there,

We've recently ported our documentation over to Sphinx in the @godotengine community, and we're currently using the rtd theme: http://godot-wiki.readthedocs.org/

One minor issue that we're having is that the left sidebar will always show the headings of the currently opened page, while we would like to keep this left sidebar only limited to :toctree: elements (i.e. it should show only page names, not give links to each page's sections).

An example would be this page for example: http://godot-wiki.readthedocs.org/en/latest/reference/richtextlabel_bbcode.html#reference
We find that two levels of depths (heading 1 (page title) and heading 2 + 3) is a bit too much and might give users the feeling that there are many subtutorials, when those are actually just (relatively short) subsections.

Is there already an option to limit this behaviour to e.g. only 1st-level heading, or if not, would that be a feature proposal that you would find pertinent? And again if not, could you give us some pointers on how to customize this ourselves on our end?

Thanks in advance!

agjohnson:
Fixes #277

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -121,7 +121,7 @@
 
         <div class="wy-menu wy-menu-vertical" data-spy="affix" role="navigation" aria-label="main navigation">
           {% block menu %}
-            {% set toctree = toctree(maxdepth=4, collapse=theme_collapse_navigation, includehidden=True) %}
+            {% set toctree = toctree(maxdepth=theme_navigation_depth, collapse=theme_collapse_navigation, includehidden=True) %}
             {% if toctree %}
                 {{ toctree }}
             {% else %}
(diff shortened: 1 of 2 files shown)


akien-mga:
Thanks! Would this option be accessible from `conf.py` when using sphinx_rtd_theme installed system-wide, or does one need a custom copy of it to modify the `theme.conf` file?

agjohnson:
@akien-mga You can set options in conf.py, see http://www.sphinx-doc.org/en/stable/config.html#confval-html_theme_options

akien-mga:
@agjohnson Thanks, that was my understanding but I tried this for `collapse_navigation` recently but it did not seem to work: https://github.com/godotengine/godot-docs/commit/8976f4de32e615d1b6c6e8963f7d9e75b70ff121
Or do I have a syntax error?

akien-mga:
Hm even after [fixing my syntax error](https://github.com/godotengine/godot-docs/blob/master/conf.py#L57-L60) the option [doesn't seem to be taken into account](http://docs.godotengine.org/en/latest). Any clue as to what I'm doing wrong?

Issue #316, opened 2016-08-25
image size ignored if only height is set


r3mi:
image size is ignored when only height is set.
this works:

``` rst
.. image:: image.*
   :width: 4em
```

this doesn't work (image displayed at original size):

``` rst
.. image:: image.*
   :height: 4em
```

jessetan:
Closed by #505

PR #318, opened 2016-08-25
fix for #316 : image size ignored if only height is set

r3mi:
proposed fix for #316.
Not sure why the `!important` was needed : couldn't find any negative impact, is there a known problem ?
I added a test in the `demo_docs/` as well for #316 .


sass/_theme_rst.sass
```diff
@@ -17,7 +17,7 @@
   // Sphinx by default applies HxW style attributes to images. This fixes that oversite.
   img
     max-width: 100%
-    height: auto !important
+    height: auto
```
Blendify: Not sure here but maybe a better solution would be to set `width: auto !important`

Blendify:
@r3mi are you still around to update this to master. I think the changes here are good and should be merged.

Blendify:
Re-opened as https://github.com/rtfd/sphinx_rtd_theme/pull/505

PR #319, opened 2016-08-29
Fix #260: Add gettext functions to all messages

### LINKED ISSUE #260: Multiple not translated strings
I'd like to have my docs translated to Polish and now I have to translate displayed strings myself. Examples from current master:

https://github.com/snide/sphinx_rtd_theme/blob/012d42db6a2b00f799223963fb8b0aa3754630eb/sphinx_rtd_theme/breadcrumbs.html
"Docs", "Edit on GitHub", "Edit on Bitbucket", "View page source"

I expected the template to honor 'language' setting from the configuration file.

Mingun:
This pull request added missing `gettext` tags to messages, which should be translatable. Partially closes #260.

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -23,7 +23,7 @@
 <div role="navigation" aria-label="breadcrumbs navigation">
   <ul class="wy-breadcrumbs">
     {% block breadcrumbs %}
-      <li><a href="{{ pathto(master_doc) }}">Docs</a> &raquo;</li>
+      <li><a href="{{ pathto(master_doc) }}">{{ _('Docs') }}</a> &raquo;</li>
         {% for doc in parents %}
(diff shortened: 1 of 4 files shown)


ericholscher:
Seems like a good addition. @agjohnson any thoughts?

agjohnson:
We shouldn't translate ``Read the Docs``, as it's branding, not just a string. Looks good otherwise.

Mingun:
> We shouldn't translate  `Read the Docs`, as it's branding, not just a string.

Yes, I known. But this still UI string and in some cases may be useful to «translate» it. Furhermore, the presence feature better its absence.

agjohnson:
I would argue we *never* want to translate it. It's not just UI text, it's our brand, name of the site, and our domain name -- none of which transfers to other languages cleanly.

mitya57:
I think it can be closed now that #343 is merged?

Mingun:
Closing, because same PR #343 was already merged.

PR #331, opened 2016-09-30
Add Next/Previous next to top of each page.

Carreau:
This makes it much easier to navigate documentation, instead of having
to scroll to the bottom of the page, or use the Toctree on the side
which is hard to use on mobile; or if you are unsure what you are
looking for.

---- 

Animated gif/jif:

rtd-top-nav

CC: @willingc


sphinx_rtd_theme/breadcrumbs.html
```diff
@@ -55,5 +56,14 @@
       </li>
     {% endblock %}
   </ul>
+
+  <div class="rst-breadcrumbs-buttons" role="navigation" aria-label="breadcrumb navigation">
```
ericholscher: Hmm. Seems like this should be an option that's enabled with `html_theme_options` -- where you can set it to `[top, bottom, None]`.

ericholscher:
Generally +1 -- just needs a bit more configurability I think. Though I could also be convinced that it should be at the top & bottom on all themes. /cc @agjohnson

sphinx_rtd_theme/breadcrumbs.html
willingc: I like the idea of options: both, top, bottom, none ;-)

Carreau: > I like the idea of options: both, top, bottom, none ;-)

Good idea. Should be a  4-enum (top bottom both None) or 2 boolean top:Yes|No bottom:Yes|No.

Carreau:
> Though I could also be convinced that it should be at the top & bottom on all themes

I think that the current button might be a bit big/heavy to be at the top unconditionally. 
I would like to find a better placement, or maybe play with a simple css. 

I must say that when a theme has previous/next at the top is is _way_ easier to navigate (http://xon.sh/ for example). I keep my mouse on the `next` button, click it, read the introduction to this page, and decide whether or not to skip it. To skip I "just" have to click and I know on the next page my mouse will still be on the button. 


Though I can add the configurability and we can paint the bikeshed about whether it should be the default on another PR.

Carreau:
> Generally +1 -- just needs a bit more configurability I think. Though I could also be convinced that it should be at the top & bottom on all themes. /cc @agjohnson

Configurability added : 

```
    Make location of buttons configurable

    Add the `prev_next_buttons_location` variable which understand `top`,
    `bottom`, `both`.Any other value will not display the next and previous
    buttons.

    Not setting the value leave it to `bottom` by default.
```

Updated the readme as well.

ericholscher:
Thanks!

willingc:
Thanks @ericholscher for the merge. :sunflower: Yay @Carreau :tada:

Carreau:
Sweet ! Thanks !

ndameong:
moseslin4@gmail.com.

sphinx_rtd_theme/breadcrumbs.html
```diff
@@ -57,6 +57,7 @@
     {% endblock %}
   </ul>
```
popehatmatt: #279

PR #332, opened 2016-10-01
Add source for the minified modernizr.js file.

jmallach:
Add the source file for the minified version of modernizr.js, which although
is not required by the license, it is encouraged to make it easy to
check/debug, and is also recommended by downstream redistributors like
Debian. See https://wiki.debian.org/onlyjob/no-minification for some rationale.

### DIFF
--- a/sphinx_rtd_theme/static/js/modernizr.js
+++ b/sphinx_rtd_theme/static/js/modernizr.js
@@ -1,0 +1,1393 @@
+/*!
+ * Modernizr v2.6.2
+ * www.modernizr.com
+ *
+ * Copyright (c) Faruk Ates, Paul Irish, Alex Sexton
+ * Available under the BSD and MIT licenses: www.modernizr.com/license/
+ */
+
+/*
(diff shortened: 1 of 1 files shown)


jmallach:
Hi, as bower and grunt are something I haven't used before (I'm just the Debian maintainer of a package that embeds a copy of sphinx_rtd_theme), I can use some help here. Else I'll do my homework as soon as my very copious free time permits. Thanks for the pointers @agjohnson!

agjohnson:
Roger, happy to help here a bit. Is it fine to have grunt and bower be packaging dependencies? My interpretation of the packaging workflow would be that a pre-packaging step would run bower and grunt to get the files prepped for vendoring in the deb package. This leaves the sources out of our repository and should still provide package maintainers a method of vendoring our dependencies.

What package is this for? I've only done a bit of deb packaging, but might be of some use there -- or at very least, having the deb package source might make the packaging requirements more clear to me.

ericholscher:
Closing this, as it hasn't been updated in a few months. Feel free to reopen with requested changes.

PR #335, opened 2016-10-17
Cast maxdepth to an int.

### LINKED ISSUE #334: Maxdepth is a int, not a string.
https://github.com/snide/sphinx_rtd_theme/pull/309 seem to have make maxdepth a string ('4')  which make sphinx complain in python 3  that string and int are not-orderable.

Carreau:
closes #334

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -121,7 +121,7 @@
 
         <div class="wy-menu wy-menu-vertical" data-spy="affix" role="navigation" aria-label="main navigation">
           {% block menu %}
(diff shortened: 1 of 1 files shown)


Carreau:
I'm not sure how to do it differently. I looked at several other sphinx
theme to figure out how they were having default value that were integers.
That's the only way I found. I agree that I would prefer doing it through
an explicit python API but as far as I can tell, we need to patch Sphinx
for that.

Carreau:
Get we get that in to fix the theme ? Otherwise it is plainly broken for some projects.

Thanks !

ericholscher:
Looks good to me. We can try and fix this another way over time if need be.

Carreau:
Thanks, appreciate it.

PR #337, opened 2016-10-17
Collapse sidebar

Carreau:
Make sidebar collapsible even on desktop.

---

collapse

---

The small hamburger menu is not shown on narrower devices, and fixed on top to be visible even on page scroll.


agjohnson:
I'm not 100%  on this, i'll mark this WIP until I have a chance to play with this more.

I'm not sure, but perhaps hiding by default on narrow displays is better UX -- that is, find a better default width to set as the mobile `max-width`. I don't think the button added here is quite obvious enough to users, but i'm not sure currently what improvements i'd make -- this requires some time on my part with this change.

Carreau:
> I'm not sure, but perhaps hiding by default on narrow displays is better UX

It already hides by default on "Tablette" which I guess is narrow display. 

I'm fine changing the UI element to toggle that, whatever you want. But I think having the ability to hide on wide display is nice as well. I often use the docs on a background browser as you can scroll without bringing browser to front. having to resize the windows to make the navbar disappears would be much less useful maybe even annoying. 

[Python's doc](https://docs.python.org/3/library/warnings.html) use js with the 8-10 px stripe and the `<<`/`>>` symbol to collapse/expand. I wanted to avoid Js to do it, but I can if you think it's worth it.

alyjak:
+1, I would definitely appreciate this

ericholscher:
I'd like to get this merged. @agjohnson any thoughts?

agjohnson:
I'm still not +1 on this, and haven't had a chance to poke this. I'll take assignment on this.

I don't see this as a pressing feature, so I'm not inclined to push this through just yet.

indyragandy:
Get directly

1138-4EB:
@agjohnson you might find the approach in [portainer](https://github.com/portainer/portainer) interesting. A [glyphicon-transfer](https://www.w3schools.com/bootstrap/tryit.asp?filename=trybs_ref_glyph_transfer&stacked=h) button is added on the right of the home/title: https://github.com/portainer/portainer#demo

jessetan:
The placement of the icon in Portainer looks better: in the sidebar, rather than next to it when its shown. We can't put it in the sidebar when it is hidden because unlike Portainer, we don't have a list of icons we want to show.
Also I'm not too fond of the hamburger, as there is more than just a hamburger menu (e.g. version selection). Perhaps just an arrow or similar indicator.

SysEngDan:
Is there someone who can review this PR and merge? This is a feature our team needs. We have a lot of content in the body and require as much screen real estate as possible, yes even with 1920px wide monitors.

PR #342, opened 2016-11-17
Fix versions.html for non-rtfd instances.

sanscore:
I have a RTD server running internally and this theme has a problem navigating to different versions of the docs. I believe that this problem has existed for a while, but has gone unnoticed due to readthedocs.org gives each project a subdomain on readthedocs.io so it has gone unnoticed. More details in the commit message.

I did run `grunt` and `grunt build`, but nothing new was generated.

### Commit Message

```
It appears that the template variable `versions` must have changed at
some point (or else, this code was never "correct").

As is, the href produced from "url" is "/en/latest/". This works for
projects on readthedocs.org because the resulting url uses the project
subdomain http://project-name.readthedocs.io/en/some-ver/.

However, this does not work for privately managed rtfd servers where
projects do not have project-specific subdomains. So, the resulting href
produced is http://rtd.private-company.tld/en/some-ver/ when it
should be http://rtd.private-company.tld/docs/project-name/en/some-ver/.

I had a look in the rtfd/readthedocs.org project for a solution. I found
a few instances, like readthedocs/templates/core/project_details.html,
that use a `for version in versions` block with the href produced by
`version.get_absolute_url` and the slug comes from `version.slug`.
```

### DIFF
--- a/sphinx_rtd_theme/versions.html
+++ b/sphinx_rtd_theme/versions.html
@@ -9,8 +9,8 @@
     <div class="rst-other-versions">
       <dl>
         <dt>Versions</dt>
-        {% for slug, url in versions %}
-          <dd><a href="{{ url }}">{{ slug }}</a></dd>
+        {% for version in versions %}
+          <dd><a href="{{ version.get_absolute_url }}">{{ version.slug }}</a></dd>
         {% endfor %}
       </dl>
       <dl>


ericholscher:
This code is using the versions config object that RTD generates here: https://github.com/rtfd/readthedocs.org/blob/master/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl#L83-L85

It's definitely specific to builds of RTD, and could probably be extended in a more reasonable manner. I don't think we've specifically passing in the Version objects that this change seems to be using, unless I'm missing it somewhere?

sanscore:
Interesting. I'm at a loss on how I can debug this issue so I have been working mostly on intuition. @ericholscher  Do you have any suggestions? A solution that drops me into `pdb` would be perfect, but I'd settle for `print(...)` debugging at this point. 😄

sanscore:
Actually. On second review of `doc_builder/conf.py.tmpl`, I can see that this is expected behavior. I'll  look into other solutions. Potentially, it will just be an internal change to `doc_builder/conf.py.tmpl`.

sanscore:
I'll look at my issue as a problem with `readthedocs.org` and fix from there. If the solution for internal changes is portable with upstream, `readthedocs.io`, etc. then I'll push it there. Thanks, again.

PR #343, opened 2016-11-25
Add gettext functions to all messages (2)

mlutfy:
follow-up to PR #319, does not translate the "Read the Docs" branding, as per feedback from @agjohnson.

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -24,7 +24,7 @@
 
   <ul class="wy-breadcrumbs">
     {% block breadcrumbs %}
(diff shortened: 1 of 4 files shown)


ericholscher:
Looks great, thanks!

mitya57:
This is useful, thanks! We are now successfully using a translated version of the RTD theme in our project: http://docs.istina.msu.ru/.

OskarPersson:
How do I use this? I'm currently using sphinx-intl to translate all the content but I can't find where I actually translate the components created from this theme.

mitya57:
@OskarPersson Try this: https://github.com/rtfd/sphinx_rtd_theme/issues/403#issuecomment-289096940

OskarPersson:
@mitya57 Works great, thanks!

Issue #329, opened 2016-09-30
Odd vertical spaces in pre elements


skirpichev:
This really break 2D ASCII-art figures, as pasted from the console session.

C.f. 
![1](https://cloud.githubusercontent.com/assets/2155800/18990260/c3d6baec-8721-11e6-8c9e-55d8bca541db.png)

And console:
![2](https://cloud.githubusercontent.com/assets/2155800/18990351/40316dd0-8722-11e6-8362-bb150433906b.png)

Unsetting line-height for pre - solves problem for me.

PR #344, opened 2016-11-28
Fix spacing in pre

skirpichev:
Now it looks like this:

c.f. with #329.

Could you review also #335?  Theme is broken on py3 without this patch.


ericholscher:
Thanks!

njsmith:
fyi, this broke the display of line numbering: #381

PR #345, opened 2016-12-05
Add 'Edit on GitLab' support

denisvm:
Add "Edit on GitLab" support and upgrade Font-Awesome to 4.6 as the [GitLab icon](http://fontawesome.io/icon/gitlab/) was added only in this version.


agjohnson:
Thanks, this looks great! If there were any issues with the asset build, I can take care of rebuilding on next release.

PR #351, opened 2016-12-28
Add comments block which can be overriden

atodorov:
With this new block I was able to override the theme and add a Disqus comments section on http://mutation-testing-patterns.readthedocs.io/en/latest/. Scroll at the bottom to see how it looks.


ericholscher:
Thanks :)

PR #365, opened 2017-03-02
Remove forcing of image size on badge.

ericholscher:
This is a really weird thing to do in our badge CSS.
It makes badges show up 100% width on sites that use our badge,
which seems like a broken thing.


ericholscher:
screenshot 2017-03-02 15 14 58

agjohnson:
Ooo pretty. Yeah, this fix looks good then

PR #366, opened 2017-03-03
Use 100% width for rst-footer-buttons

Blendify:
This fixes the buttons from jumping around on small pages/ resizing


ericholscher:
Is there an example that this fixes? Would be good to understand a bit more.

Blendify:
broken page

Here is an example

Blendify:
To fix this we did:

```
.rst-footer-buttons::before , .rst-footer-buttons::after{
	width: 100%;
}
```
I assume this does the same thing; just started using sass today.

Blendify:
planing changes here

ericholscher:
Seems like a reasonable fix, thanks. I'll take a look and merge it next week.

Blendify:
See https://github.com/snide/sphinx_rtd_theme/pull/370

PR #372, opened 2017-03-05
Include fontawesome-webfont.woff2 in pip package

Minecrell:
`fontawesome-webfont.woff2` was added in f22bfd2, but it was never added to the white list for the pip package. Without the woff2 file, modern browsers are required to fall back to the older woff format which has a slightly larger file size.


agjohnson:
Looks good. Thanks for catching this!

astrofrog:
Would it be possible to release a new bugfix version of sphinx_rtd_theme to include this fix?

Blendify:
There is a beta out right now.

PR #373, opened 2017-03-06
Fix problem triggered by unconditional toctree call and empty toctree

agjohnson:
This bumps to v0.2.3

Refs rtfd/readthedocs.org#2398


agjohnson:
We should open a reminder ticket to remove once sphinx-doc/sphinx#3518 is merged.

PR #375, opened 2017-03-06
Adds testing and hopefully a final fix for the empty toctree issue

agjohnson:
This adds some more hacky fixes to our layout to avoid singlehtml builders.
Instead of trying to guess if this is working, I added testing!


sphinx_rtd_theme/layout.html
```diff
@@ -129,7 +129,7 @@
               The singlehtml builder doesn't handle this toctree call when the
               toctree is empty. Skip building this for now.
             #}
-            {% if builder != 'singlehtml' %}
+            {% if builder not in ['singlehtml', 'readthedocssinglehtml', 'readthedocssinglehtmllocalmedia'] %}
```
ericholscher: Any reason not to do `if 'singlehtml' in builder`?

tests/test_builders.py
```diff
@@ -0,0 +1,99 @@
...
+            search = (
+                '<div class="local-toc"><ul>\n'
+                '<li class="toctree-l1">'
+                '<a class="reference internal" href="index.html#document-foo">foo</a>'
+                '<ul>\n'
+                '<li class="toctree-l2">'
+                '<a class="reference internal" href="index.html#document-bar">bar</a>'
+                '</li>\n'
+                '</ul>'
+            )
```
ericholscher: This seems brittle, I wonder if we could just look for the class names? `local-toc` vs. `toctree-wrapper`?

ericholscher:
Ha, just happened :)

tests/test_builders.py
agjohnson: It shouldn't be all that brittle, but I do think it's important to test the full thing. Otherwise, we won't be testing if the toctrees have anything in them, which is the more important part.

sphinx_rtd_theme/layout.html
agjohnson: I debated it. It's probably a cleaner implementation, so 👍

tests/test_builders.py
ericholscher: Makes sense. :+1:

Issue #292, opened 2016-04-14
Captions on literal includes and code blocks are not styled


exhuma:
When using `:caption:` inside the `.. literalinclude:` directive, the caption is not properly styled.
- The `¶` character is visible
- The font does not fit visually well with the box containing the included text.

Input:

```
.. literalinclude:: ../pynet/test/test_case_invalid_vprn_sap_description.py                                                                                                 
     :language: python                                                                                                                                                       
     :caption: pynet/test/test_case_invalid_vprn_sap_description.py    
```

Output:

![image](https://cloud.githubusercontent.com/assets/65717/14532381/2a20b1cc-0261-11e6-8800-861231d01c29.png)

exhuma:
Hi again :)

I just came across this in another situation. The directive `.. code-block` can have captions as well and is rendered in the same way.

Might be useful for another test-case as with 8df9434a3ef5df74009c546d9f51f3bf4502e140?

PR #378, opened 2017-03-07
Caption rework

Blendify:
Before:
screenshot from 2017-03-07 12 59 34

After:

screenshot from 2017-03-07 12 58 45

fixes https://github.com/rtfd/sphinx_rtd_theme/issues/315


Blendify:
Ok this is ready for review the before and after pictures are above except the after picture does not have the vertical-align change.

Blendify:
> The decreased font size and decreased contrast reduce the readability of the captions. This is an negative change for accessability of our docs

This was done to better distinguish body text and captions.

> The table formatting appears to have regressed, I prefer the formatting before this change

I think I can still improve this. But mind being more specific here?

agjohnson:
> This was done to better distinguish body text and captions.

The theme isn't currently displaying any other text at this low contrast, so this is a bit of a stylistic deviation on top of being a fairly unreadable. There might be some other treatment to the captions that preserves readability and still distinguishes the elements.

> I think I can still improve this. But mind being more specific here?

The table inner cell borders are dark in the after screengrab.

Blendify:
This patch now also fixes #292

Blendify:
Reopened as #401

PR #380, opened 2017-03-08
Fix #273 line-block: set the left margin to 0px

### LINKED ISSUE #273: Line breaks causing indented text when using the theme
When using the vertical bars to force line breaks, the text is indented 

```
| this is line 1.
| this and the previous line should be unindented.
```

results in

```
  this is line 1.
  this and the previous line should be unindented.
```

I got around this by using the raw html <br/> for now but just wanted to find out why the rtd theme was rendering the line breaks differently.

Thanks
-Selim

Blendify:
this fixes https://github.com/rtfd/sphinx_rtd_theme/issues/273


Blendify:
This is actually removing intended indent and needs a different solution

davidfischer:
I co-opted this PR with some additional fixes. I hope that's ok. Essentially my changes treat top level and nested line blocks slightly differently. Top level line blocks have a bottom margin and no left margin while nested line blocks have a left margin but no bottom margin.

This also backports a [workaround from RTD.org](https://github.com/rtfd/readthedocs.org/blob/a7f5425/media/css/readthedocs-doc-embed.css#L23-L26) into the RTD theme.

screen shot 2018-01-15 at 11 31 00 am
(Before and after of this PR: specifically notice left and bottom margins on the line-blocks)

Blendify:
Changes look good to me. Needs aproval from someone else first.

Blendify:
I submitted a patch: https://github.com/rtfd/readthedocs.org/pull/3519 to remove this from the main RTD repo

PR #382, opened 2017-03-08
Amend #344, correct spacing on line-numbering

skirpichev:
@njsmith, does this works for you?

BTW, is there any chance that fix for #329 will be in rtd soon?

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -19,7 +19,7 @@
     max-width: 100%
     height: auto !important
 
-  .highlight > pre
+  .highlight > pre, .linenodiv > pre
     line-height: normal
 
   div.figure


njsmith:
I haven't figured out how to rebuild and test the theme or I'd have written a PR myself, but for a test case it should be enough to do something like:
```
.. code-block:: none
   :linenos:

   1
   2
   3
   4
```

(And regarding getting newer stuff onto RTD, obviously I'm not @ericholscher or anyone, but... I've also noticed in general that RTD tends to lag way behind on upgrading to new versions, which makes sense, I guess I don't want them to spend all their time dealing with version churn either :-). BUT if you specifically request a new version be installed via requirements.txt or environment.yml then it does seem to use that? For my docs I need Python 3.5 and an updated sphinx_rtd_theme, and by [turning on conda](https://github.com/njsmith/trio/blob/master/.readthedocs.yml) and then [requiring those in my conda environment.yml](https://github.com/njsmith/trio/blob/master/ci/rtd-conda-env.yml) I was able to get them working.)

skirpichev:
It seems that latest version is installed (i.e. 0.2.4):
http://readthedocs.org/projects/diofant/builds/5120623/

Do you suggest add requirement for git version of the theme?

njsmith:
Oh, that's cool, I guess they must have just updated the versions? Beyond that I can't really advise... you certainly could pull in a git version of the theme if that works for you.

ericholscher:
This looks like a solid change, but could use a test case or an example -- or pointing to somewhere it's broken so I can verify it is a fix for it.

skirpichev:
@ericholscher, it seems - installed via pip version of the theme is not used on rtd?  c.f. [this](http://diofant.readthedocs.io/en/latest/tutorial/intro.html) - there are same odd spacing as for original problem, #329.  (See matrix example at the bottom.)  But as you can see, the latest release [was installed](http://readthedocs.org/projects/diofant/builds/5124436/).

skirpichev:
@njsmith, @ericholscher, hmm - I think this doesn't work.  Somehow, line-height=1.5 inherited...

@ericholscher, you can reproduce problem with
```rst
.. code-block:: python
   :linenos:

    def mrv_max(f, g, x):
        """Computes the maximum of two MRV sets."""
        if not f or not g or f & g:
            return f | g

        c = compare(list(f)[0], list(g)[0], x)
        if c:
            return f if c > 0 else g
        else:
            return f | g
```
There are 10 lines.  Here is the problem:
before

skirpichev:
@ericholscher, oh, no, everything works.  Just forgot to run grunt...

After this change:
after

skirpichev:
@Blendify, maybe you can answer question in [this comment](https://github.com/rtfd/sphinx_rtd_theme/pull/382#issuecomment-285269624)?

Blendify:
@skirpichev we will release a new version of the theme some time in the next week. This will also mean that we will bump the version used on RTD.

Blendify:
@skirpichev we will release a new version of the theme some time in the next week. This will also mean that we will bump the version used on RTD.

skirpichev:
@Blendify, but right now it's not up to date (ie. 0.2.4), correct?

Blendify:
Currently, I think RTD is using 0.2.4

Aaron Carlisle

Picture taker | Bit cruncher | Pixel pusher | Document writer |
Project administrator for the Blender 3D Documentation Project

skirpichev:
> Currently, I think RTD is using 0.2.4

Then I don't understand why my fix from #329 doesn't work.  See [this](https://github.com/rtfd/sphinx_rtd_theme/pull/382#issuecomment-285269624).  Locally - everything fine (after installing 0.2.4).

Should I fill bugreport?

Blendify:
Sure, I do not fully understand the issue so a new report my clear things up.

skirpichev:
The issue is that fix for #329 is included in 0.2.4 and it works for me locally, if I install that release.  But I still see issue #329 on RTD.

If RTD uses older release, there is no issue, of course.  Why do you believe that it uses latest version?

Blendify:
Ok I was wrong Read the Docs was using 0.2.0 version of the theme but I
just bumped it to 0.2.4

Aaron Carlisle

Picture taker | Bit cruncher | Pixel pusher | Document writer |
Project administrator for the Blender 3D Documentation Project

skirpichev:
Hmm, I triggered [the build](http://readthedocs.org/projects/diofant/builds/5230429/), but nothing has changed.  See matrix example at the bottom of the [introduction](http://diofant.readthedocs.io/en/latest/tutorial/intro.html) page, it's just as in #329.

njsmith:
I'd appreciate that release if you have a chance, because this bug is currently causing my RTD docs to be mis-rendered :-)

njsmith:
Also, thanks to an old pip bug, it's not actually possible to force rtd to use an unreleased version of the theme :-(: https://github.com/pypa/pip/issues/536#issuecomment-294406115

ericholscher:
Hmm. We're not ready to release a new version of the theme and deal w/ the possible fallout this week, but what we can do is release a beta version to pypi, so you can install it properly w/ a requirements file, and then we can roll it to a real release after some more testing :+1:

ericholscher:
I've uploaded `0.2.5b1` -- let me know if that fixes this.

PR #384, opened 2017-03-08
Data display as a table

### LINKED ISSUE #90: ..data indented constants should be separated by \n
Assuming an rst file like:

```
.. data:: PRIO_PROCESS
          PRIO_PGRP
          PRIO_USER
```

The resulting html file will look this:
http://docs.python.org/3/library/os.html#os.PRIO_PROCESS

RDD hower, fail to separate every entry with a new line, see for example:
http://psutil.readthedocs.org/en/latest/#constants

Blendify:
This fixes #90

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -256,6 +256,9 @@
         font-size: 100% !important
     dt:first-child
       margin-top: 0
+    // And this would be the ..data::
+    dl.data dt
+      display: table
     // Since dts get the callout style, we treat this less as callouts.
     tt, code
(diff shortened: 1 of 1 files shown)


agjohnson:
Can you give an example of what this does? It's not clear the implications of this change.

Blendify:
Before:
screenshot from 2017-03-09 16 38 54

After:

screenshot from 2017-03-09 16 47 38

agjohnson:
Great, this looks like a good addition. Other signatures like this are already displayed like this, good to have parity here.

PR #386, opened 2017-03-09
Convert Sass to Scss

Blendify:
This Converts Sass to Scss at the downside off loss of git history


ericholscher:
What's the benefit of this change?

Blendify:
It allows use to us jinga markup within the scss.

ericholscher:
Talked a bit about this in IRC. If we can use this as a way of passing Jinja variables into the outputted CSS, it would be a huge win for users and their ability to configure the theme. I'd like to see an example of this working though, as I haven't been able to get sass to do it previously.

agjohnson:
I'm -1 on this change if #387 still isn't possible. I think we can find another way of providing a css template file though.

PR #387, opened 2017-03-09
Add the ability to configure the width

Blendify:
This PR relies on https://github.com/rtfd/sphinx_rtd_theme/pull/386

### DIFF
--- a/demo_docs/source/conf.py
+++ b/demo_docs/source/conf.py
@@ -108,8 +108,7 @@ html_theme = 'sphinx_rtd_theme'
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    # 'sticky_navigation': True  # Set to False to disable the sticky nav while scrolling.
-    # 'logo_only': True,  # if we have a html_logo below, this shows /only/ the logo with no title text
+    "content_width": "auto"
 }
 
 # Add any paths that contain custom themes here, relative to this directory.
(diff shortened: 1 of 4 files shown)


Blendify:
This pr build and work but I cannot seem to access the new theme option. the theme should use `auto` for the width but it does not.

demo_docs/source/conf.py
```diff
@@ -108,8 +108,7 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    # 'sticky_navigation': True  # Set to False to disable the sticky nav while scrolling.
-    # 'logo_only': True,  # if we have a html_logo below, this shows /only/ the logo with no title text
+    "content_width": "auto"
```
Blendify: Just a note that this is only meant for testing purposes

agjohnson:
So does this still not work with scss then? In the output, do we have `{{ theme_content_width }}` where expected? If this method doesn't work, I think #386 doesn't buy us much. We can explore options for making this into a css template file, or supporting a secondary template file that is vanilla css.

Blendify:
No I thought it was working but I forgot to tell grunt to use scss instead of sass

agjohnson:
Roger. If you feel this is a dead end, we can close out these issueus and open up a ticket to discuss possible solutions to either converting this to a css template, or perhaps think about adding a secondary css template file with minimal overrides.

PR #389, opened 2017-03-10
Set the before & after width of the nav buttons to 100%

Blendify:
This solves the issue where the buttons might appear in weird locations when the is a trailing element aligned to one side (e.g. an image).

See below for an example of this broken result:

broken page


agjohnson:
Looks good!

PR #390, opened 2017-03-12
Para block

Blendify:
Before:
blockbefore

After:
blockafter

Take note of the comment in the code:
```
  // For the most part, its safe to assume that sphinx wants you to use a blockquote as an indent. It gets
  // used in many different ways, so don't assume you can apply some fancy style, just leave it be.
```

Looking at our demo files the only affects it has is on block quotes and line blocks. I can make it only apply to block quotes. AFAIK this currenty would cause issues on my website but would like to get some feed back on this.

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -47,8 +47,9 @@
     margin-left: $base-line-height
     line-height: $base-line-height
     margin-bottom: $base-line-height
-    background-color: darken($section-background-color, 5%)
-    border-left: $base-line-height / 4 solid darken($section-background-color, 10%)
+    padding-left: $base-line-height / 4
+    background-color: darken($section-background-color, 3%)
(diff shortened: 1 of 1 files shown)


Blendify:
My personal preference would be to not add the styling to line blocks

Apus21:
13°48 39.3"N  100°29"48.7

Apus21:
eqnyte connect pong22077file Serve your trial ended please subseribe online or call us at 1.877.734.6983

jessetan:
There may be cases where blockquote is used for layout purposes, e.g. [sphinxcontrib-httpdomain](https://pythonhosted.org/sphinxcontrib-httpdomain/#basic-usage). This is of course a wrong use of blockquote, but it demonstrates that changing the rendering may have unintended side effects in specific cases.
I like the look though.

PR #391, opened 2017-03-12
Fix #256 include lato italic font

### LINKED ISSUE #256: Include Lato Italic font
I was wondering if it is possible to get the Lato Italic fonts included. I noticed that this has been discussed previously in issue rtfd/readthedocs.org#824 but Lato Italic does not seem to be included still. it seems that this could be easily fixed by adding the Lato Italic font file and a corresponding `@font-face` block in [sass/_theme_font_local.sass](https://github.com/snide/sphinx_rtd_theme/blob/master/sass/_theme_font_local.sass).

(I reported this issue in rtfd/readthedocs.org#1755 but was directed to this repository instead.)

Blendify:
Adds the lato italic font locally fixes #256


Blendify:
Does the font file also need to be add? Or does grunt build copy is from bower? If so do I just add the font file to static/fonts?

Blendify:
Should be fine now.

agjohnson:
Yup looks good!

PR #393, opened 2017-03-14
Match change done in readthedocs fix #252

### LINKED ISSUE #252: Hide Edit links on autocreated pages
We should implement something similar to https://github.com/rtfd/readthedocs.org/pull/1643 on the theme.

Blendify:
merge the change from https://github.com/rtfd/readthedocs.org/pull/1643 and fixes #252


ericholscher:
LGTM. :+1:

ericholscher:
@Blendify This probably warrants a change in the changelog, as it's an oft requested feature. Also we should probably have a way to track changes that should result in a release, as this seems like a broadly useful change. Something to think about. /cc @agjohnson

PR #394, opened 2017-03-14
Fix #149 Fix padding for field list

### LINKED ISSUE #149: Field lists have weird padding
Source:

```
.. class: Test
    .. method: test()
        :param int point: Point in the view whose scope the selector should be
                          matched against.
        :param str selector: A scope selector.
        :returns bool: Whether the selector matches or not.
```

Current css:

``` css
.rst-content table.field-list td {
    border: medium none;
    padding-top: 5px;
}
```

Without padding:

``` css
.rst-content table.field-list td {
    border: medium none;
}
```

Blendify:
Before:

listbefore
After:

listafter

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -192,7 +192,7 @@
       white-space: nowrap
     .field-body
       text-align: left
-      padding-left: 0
+      padding: 8px 0px 0px 0px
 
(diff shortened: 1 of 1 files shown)


Blendify:
Someone may want to ensure that they do in act line up.

PR #396, opened 2017-03-14
Fix #305 Move 1400px media-query to 1100px 

### LINKED ISSUE #305: Move 1400px media-query to 1100px ?
The `.wy-nav-content` wraps at 1100px width when above 1400px+  wide, and has a linear growth until this value, which make the website "jump" when you reach the 1400px width threshold. 

Which is then weid as a 1401px wide window has narrower content than a 1399 one...

I think the theme should either move the 1400px transition to 1100px. THen the content just "stop growing", or change the max-width of `wy-nav-content` to be higher by 300px. 

Both looks fine. I can likely make a PR depending on the one you prefer.

Blendify:
This makes the media-query always equal to the sidebar width + the content width. Fixes #305


Blendify:
Ok, I found the issue with the content taking the full width, you can not use `calc()` inside a media query. We can either hack our way around this or use javascript and get the width of the sidebar (`wy-nav-side`) then get the with of the content (`wy-nav-content`). After getting these two values we can pass the value to the media query. This would be the proper solution however, I do not know javascript so someone else will have to do this.

Blendify:
Poke.

Blendify:
Can we please merge this?

davidfischer:
I am going to defer to @agjohnson for final approval but I think this is good to go.

Here's a screenshot to help just visualize the changes:
fix-305

The big difference is that currently in master there is a left margin on the main content block (`wy-nav-content`) up until width 1400px at which point it drops to zero. The main content block is always the same size (800px) so don't let looks deceive you.

Blendify:
With two reviews approving this and Anthony not being away I will just merge this.

PR #399, opened 2017-03-21
Fix #298 Use different styling for literals

### LINKED ISSUE #298: monospace font without background change
I would like to use a monospace font to format literals or command line arguments, but feel that the default rendering of `--my-monospace` is distracting to reading because of the change in color and background (as seen in the [rtd docs](http://read-the-docs.readthedocs.io/en/latest/theme.html)).

Is it possible to have simply change the font or some other way that makes the text stand out but does not distract so much?  As an example of what I'm after, you can open up just about any old O'Reilly book and see.

I do realize that this is highly personal preference.

Blendify:
This uses a grey styling and matchs the styling of `guilables`

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -207,7 +207,12 @@
       line-height: normal
 
     &.literal
-      color: $text-code-color
+      color: inherit
+      background-color: lighten($text-light, 20%)
+      border: 1px solid darken($text-light, 5%)
+      border-radius: $base-line-height / 6
+      margin: auto $base-line-height / 12
+      padding: $base-line-height / 10 $base-line-height / 4
     &.xref, a &
       font-weight: bold
       color: $text-codexref-color


Blendify:
After picture: 
https://github.com/rtfd/sphinx_rtd_theme/pull/399

sass/_theme_rst.sass
```diff
@@ -207,7 +207,12 @@
       line-height: normal
 
     &.literal
-      color: $text-code-color
+      color: inherit
+      background-color: lighten($text-light, 20%)
+      border: 1px solid darken($text-light, 5%)
+      border-radius: $base-line-height / 6
```
agjohnson: I'm -1 on making this look like `guilabel`. I don't like rounded borders, and rounded borders aren't used in theme anywhere except the search box and guilabel -- which is a rare role anyways.

Blendify:
Reverted the changes and instead made the literal inherit the color

Blendify:
Poke.

jessetan:
There are many ways to distinguish literals from body text. Right now the theme uses

1. font (monospaced)
2. color (red)
3. border
4. background (white)

Of these 4, number 1. is the minimum difference I feel we should have. This does not make it stand out so much, so adding some more would be good.

Use on other documentation websites:

- Docker: 1, 2, 4 (monospaced blue on light blue bg)
- GitHub: 1, 4 (monospaced black on gray bg)
- Sphinx: 1 (monospaced black)
- Stripe: 1, (2), 4 (semibold monospaced black on light gray bg)
- Bootstrap: 1, 2 (monospaced pink/red)
- GitLab: 1,2,4 (monospaced pink/red on light pink bg)
- Facebook Developers: 1, 2 (monospaced green)

My preference would be to follow to have a monospaced font, with some color and bg that fits in with the overall colors (so somewhere in the gray/blueish spectrum).

PR #400, opened 2017-03-21
Add language to the JS output variable

tobiasHeinke:
Makes the current language accessible as a JS variable.
Would be nice to have for the version & language switch I'm writing (to avoid the workaround of extracting it from the url).


Blendify:
LGTM

PR #401, opened 2017-03-21
Wrap table contents

Blendify:
Before:
screenshot from 2017-03-07 12 59 34

After:

screenshot from 2017-03-07 12 58 45

fixes https://github.com/rtfd/sphinx_rtd_theme/issues/315 and #292

### DIFF
--- a/sass/_theme_layout.sass
+++ b/sass/_theme_layout.sass
@@ -358,6 +358,8 @@ footer
     color: $text-medium
     font-size: 90%
 
+.wy-table-responsive table td, .wy-table-responsive table th
+  white-space: normal
 
 +media($tablet)
   .wy-body-for-nav
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -26,9 +26,29 @@
     margin-bottom: $base-line-height
     p.caption
       font-style: italic
+      font-family: inherit !important
+      font-size: inherit !important
+      color: $caption-color !important
+    
 
   div.figure.align-center
     text-align: center
+
+  caption
+    font-style: italic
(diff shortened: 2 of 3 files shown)


sass/_theme_layout.sass
```diff
@@ -358,6 +358,8 @@ footer
     color: $text-medium
     font-size: 90%
 
+.wy-table-responsive table td, .wy-table-responsive table th
+  white-space: normal
```
agjohnson: This sneaks in table wrapping, we should address this separately.

sass/_theme_rst.sass
```diff
@@ -26,10 +26,30 @@
     margin-bottom: $base-line-height
     p.caption
       font-style: italic
+      font-family: inherit !important
+      font-size: inherit !important
+      color: $caption-color !important
```
agjohnson: We should never have to use `!important`. Specifying a selector with great specificity makes it more explicit what we're changing, and reduces the problems with overriding that `!important` introduces. A number of folks extend the theme and this would cause problems.

sass/_theme_rst.sass
```diff
@@ -26,10 +26,30 @@
...
 
   div.figure.align-center
     text-align: center
 
+  caption
+    font-style: italic
+    // workarond a wyrm overide
+    font-family: inherit !important
+    font-size: inherit !important
+    color: $caption-color !important
```
agjohnson: Same with `!important` here. We can target this selector with greater specificity to override wyrm rules.

agjohnson:
This has been WIP for several years now, I'll close this for now. Feel free to reopen when it's ready to continue

PR #404, opened 2017-03-31
Add the ability to change the nav search background color

Blendify:
There will be changes down in master after this is accepted but for right now I left them out to avoid merge conflicts.


Blendify:
See for example https://borgbackup.readthedocs.io/en/stable/

CrazyPython:
@ericholscher Can you review this? It's been over two months now...

Blendify:
I started working a new docs branch but for now, I am going to add this to the readme

README.rst
```diff
@@ -87,13 +87,16 @@ file of this repository, and can be defined in your project's ``conf.py`` via
...
+        # Base Options
         'typekit_id': '',
         'canonical_url': '',
         'analytics_id': '',
         'logo_only': False,
         'display_version': True,
         'prev_next_buttons_location': bottom,
         'style_external_links': False,
+         # CSS Options
+        'nav_search_background_color': #2980B9,
```
Blendify: I guess the `#` will need to be escaped somehow?

davidfischer: I think it's just a string in the `conf.py` directory. Wouldn't it simply be quoted?

davidfischer:
I'm having some issues testing this and perhaps you could point me in the right direction.

Firstly, I don't think you can simply add `nav_search_background_color` to `html_theme_options` dict unless it is mentioned under `[options]` of `theme.conf`, right? At least I couldn't get it working without that. I don't see any mention of `[colors]`. Here's the [relevant docs](http://www.sphinx-doc.org/en/stable/theming.html#creating-themes).

Based on what I can see since most of the actual overrides are to the saas files wouldn't somebody wanting to use this have to rebuild the theme (with Grunt rather than just making an entry in `html_theme_options`?

Blendify:
After rethinking this I do not like this hack of passing strings around our sass. I think we should remove this sass styling and move it over the a "configurable' CSS file. Does this sound like a better idea?

davidfischer:
> Does this sound like a better idea?

To me, yes. I don't think things that can be configured easily with an additional CSS file should be a theme option.

Blendify:
This is now possible with https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-style_nav_header_background

In the future, I would like to revisit this to see if its possible to make theme variables affect sass directly.

PR #406, opened 2017-04-01
Set html language metadata to {{ language }}

Blendify:
I am not sure how valid this but this seems logical to me.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -8,8 +8,8 @@
 {%- endif %}
 
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


Minecrell:
While this is valid if you use one of the languages with a two letter locale code (e.g. `de`) it won't work for languages like `pt_BR` or `pt_PT`. It's possible that we'd just need to replace the underscore with a hyphen but I'm not entirely sure either. (See https://www.w3.org/International/articles/language-tags/)

agjohnson:
@Minecrell ah interesting, good catch. We should be able to alter this with a template transform.

Blendify:
https://github.com/rtfd/sphinx_rtd_theme/pull/443 is a better fix for this.

PR #411, opened 2017-04-10
Add support for Google Analytics to theme.

ericholscher:
This setting has actually existed in the default config since the original version,
but doesn't currently work.
This fixes that and supports GA if it's set.


Blendify:
+1 This will simplify the code for my project.

sphinx_rtd_theme/layout.html
```diff
@@ -207,6 +207,20 @@
...
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+  ga('create', '{{ theme_analytics_id }}', 'auto');
+  ga('send', 'pageview');
+  </script>
+
+  {% endif %}
```
agjohnson: This conflicts with our insertion of the GA tracker:
https://github.com/rtfd/readthedocs.org/blob/499a74076f0debae09acc1d4c4eeeab03f8f557e/readthedocs/doc_builder/templates/doc_builder/include.js.tmpl#L7-L12

Also, i don't think we want to be loading this more than once.

ericholscher: So what would be the proper solution? Seems like we just wouldn't set this on RTD, and the theme can support it if it wants?

ericholscher: I guess we'd have to forcible unset it on RTD, but that's true already if people are overriding layout to insert their own GA.

Blendify:
Side note that there is a sphinx contrib extension for this: https://pypi.python.org/pypi/sphinxcontrib-googleanalytics/

sphinx_rtd_theme/layout.html
agjohnson: I think RTD is the proper place for this. Removing it from RTD will then remove it from non RTD theme projects, which we don't want to do. Isn't the official mechanism the "Analytics Code" on advanced project settings?

Perhaps there is a way to do this conditionally -- that is, if running on RTD, the analytics block above doesn't load the script or set up GA. Instead, it either sets the GA user code or the template below is changed to look for this user GA in `conf.py` settings.

https://github.com/rtfd/readthedocs.org/blob/499a74076f0debae09acc1d4c4eeeab03f8f557e/readthedocs/doc_builder/templates/doc_builder/include.js.tmpl#L9

agjohnson:
@Blendify good point. I think that is the most correct solution. I don't know what that means for our override, because I'd still like to avoid loading the script twice. At least it provides similar functionality in both cases though.

sphinx_rtd_theme/layout.html
ericholscher: It won't load if the `analytics_id` isn't set in the `html_theme_options`.

ericholscher:
The extension above hacks it into the `meta` field, which isn't the correct solution either. This definitely seems like something the theme should support.

ericholscher:
Alabaster does similar, so I think it makes sense for us to support it also: https://github.com/bitprophet/alabaster/blob/master/alabaster/layout.html#L65-L81

sphinx_rtd_theme/layout.html
ericholscher: I think you're conflating the RTD theme with RTD -- lots of people use this theme outside of our implementation, so we shouldn't be handicapping those folks. Having a standard override mechanism gives us _more_ control over how people are integrating GA. The current status quo has them overriding templates, and there's no way for us to unset it. This way if they're using the standard mechanism, we can *unset* it from RTD's side, and avoid loading it twice. This seems like a win/win, instead of telling people to override templates to do it.

jessetan:
Isn't this already covered by readthedocs-sphinx-ext? Also see rtfd/readthedocs-sphinx-ext#29

Blendify:
Yes, but this allows people outside RTD to use it.

Aaron Carlisle

Picture taker | Bit cruncher | Pixel pusher | Document writer |
Project administrator for the Blender 3D Documentation Project

jessetan:
That extension can also take a user-supplied analytics id in a projects "advanced" settings.
Putting this functionality in the theme seems fine as well, but there will be an overlap.

Blendify:
What I think should be done is leave it up to the theme and phase out the
option on RTD.
I am planning to make a PR to include it in the default template with that
and this PR most themes will have support for this.
The other major theme, Alabaster already does this:
https://github.com/bitprophet/alabaster/blob/master/alabaster/layout.html#L65

Aaron Carlisle

Picture taker | Bit cruncher | Pixel pusher | Document writer |
Project administrator for the Blender 3D Documentation Project

README.rst
```diff
@@ -88,6 +88,7 @@ file of this repository, and can be defined in your project's ``conf.py`` via
 .. code:: python
 
     html_theme_options = {
+        'analytics_id': 'UA-3463466-1', #  Provided by Google in your dashboard
```
davidfischer: This keys exists 3 lines below here already.

sphinx_rtd_theme/layout.html
```diff
@@ -219,6 +219,23 @@
       });
   </script>
 
+  {# Do not conflict with RTD insertion of analytics script #}
+  {% if not READTHEDOCS %}
+    {% if theme_analytics_id %}
```
davidfischer: Couldn't these two if statements be combined? It works but it does seem a bit awkward.

davidfischer: Actually, the more I think about it while these can be combined they are sort of logically separate. I'm ok with it.

nsoranzo:
That's a very nice addition, thanks! Would it be possible to have a new PyPI release with it, please? :pray:

ericholscher:
Yea, I will try and get a new theme release out this week.

ericholscher:
0.4.3 should be released 👍

nsoranzo:
Thanks a lot @ericholscher !

PR #413, opened 2017-04-18
Set margin-bottom to 0 for captions in figures

Blendify:
The issue here is that figures have a margin of 24px and `p` tags also have a margin of 24px so figure caption get huge 48px margins on the bottom.


Blendify:
@agjohnson this a simple one to review, do you mind taking a quick look?

jessetan:
@Blendify Is there an example of where this is an issue? If `p.caption` is the last child of `div.figure`, then the margins should collapse. See example from demo doc (built from master):
figure caption margin collapse

Blendify:
Check the example on this page: https://github.com/rtfd/sphinx_rtd_theme/blob/master/demo_docs/source/structure.rst#document-subsection-1

jessetan:
Ah, I see. What about changing the selector to `p:last-child.caption` so you only target captions at the end of a figure? That will retain the same space between the caption and legend currently in https://github.com/rtfd/sphinx_rtd_theme/blob/master/demo_docs/source/demo.rst#figures

davidfischer:
👍 for @jessetan's proposal to use `p:last-child.caption`.

Here's a screenshot with the legend since the above links are dead due to repo structure changes:

screen shot 2018-01-15 at 4 36 01 pm
(I think there should be space there)

PR #422, opened 2017-05-20
Fix the line numbers for literalinclude.

cprogrammer1994:
- removed `line-height: normal` from `.highlight > pre, .linenodiv > pre`
because only `.highlight > pre` gets changed.


cprogrammer1994:
Otherwise you get line numbers like [this](https://moderngl.github.io/Examples/01_hello_world.html).

Blendify:
I think the lines can just be  removed

Blendify:
I will take a look Monday.

Blendify:
While this fixes the issue, I think a proper solution would be to replace those lines with:

```
pre
    line-height: normal
```

Can you confirm that this works for you? This should fix the issue and not change the spacing.

cprogrammer1994:
I'll check it tomorrow please verify [my css file](https://github.com/cprogrammer1994/ModernGL/blob/master/docs/static/css/custom.css#L21). I guess lines 21->23 should be removed before testing.

Blendify:
The real issue comes from: https://github.com/snide/wyrm/blob/fd41b56978f009e8c33cb26f384dd0dfaf430a7d/sass/wyrm_addons/pygments/_pygments.sass#L49 which is getting more priority than our css. So maybe this needs a `!important`.

TonyFlury:
Any idea when this fix will be published ?

Blendify:
We are trying to release an update of the the theme however, things have
been slow lately. I will try to get things moving again.

agjohnson:
You can avoid `!important` by using more specificity in your selector.

This issue this is referencing is not clear to me, what is the problem this is solving?

cprogrammer1994:
Line numbers and lines were not aligned in code snippets.
Currently I am not able to reproduce the bug with the latest version.
Closing.

TonyFlury:
Current version on pypi is still 0.2.4 - (i.e. not the master) therefore the bug still exists in the current live version. When will the master being release to pypi ?

PR #425, opened 2017-05-30
Added setuptools entry point for Sphinx

segevfiner:
This allows using `sphinx_rtd_theme` as Sphinx `html_theme` directly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,7 +46,7 @@ In your ``conf.py`` file:
     html_theme = "sphinx_rtd_theme"
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
-or (since master):
+or (since v0.2.5):
 
 .. code:: python
 
--- a/README.rst
+++ b/README.rst
@@ -48,6 +48,12 @@ In your ``conf.py`` file:
 
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
+or (since master):
+
+.. code:: python
+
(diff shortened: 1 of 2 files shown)


ericholscher:
Thanks. Is this the same code that was used in Sphinx previously? It would be good to have a comment or something that points to an explanation of how this works.

segevfiner:
I added a link to the section in Sphinx's documentation about this. In essence, this is just a classic usage of the setuptools/pkg_resources entry points mechanism. Documented [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins) and [here](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points). But there might be more explanations elsewhere too 😛

README.rst
```diff
@@ -48,6 +48,12 @@ In your ``conf.py`` file:
 
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
+or (since master):
```
Blendify: I think this text should be handled differently maybe use `.. versionchanged::` or `.. versionadded::` referencing the next major release of the theme (0.2.5)

segevfiner: I think GitHub's reStructuredText renderer doesn't know about `.. versionadded::`/`.. versionchanged::` which are Sphinx things.

Blendify: I would just say as of 0.2.5... then

ericholscher:
Thanks!

PR #426, opened 2017-06-02
Swagger plugin

txomon:
Basically, the swagger ui requires not only the plugin but also files in the theme. Suggestions accepted!


txomon:
You probably want to tweak something there... A quick preview just in case you are wondering what this looks like...

image

ericholscher:
This looks really neat! It seems like this could live as an extension to Sphinx itself, instead of shipping with our theme. Is there a reason you're adding it to the theme instead of as a standalone extension?

txomon:
Yeah, it requires resources in the theme and I have no idea on how to
provide them other way.

Blendify:
I agree, I would much rather see something like this be added to sphinxcontrib what exactly is forcing you to only bundle this with the theme?

txomon:
Hello, I don't understand the question. @Blendify How is this different question from the one I already answered?

If you check the PR, you will see that there are come css/js resources that need to be added to the theme for it to work.

Blendify:
Is it not possible to make the css/js work in a sphinx extension?

txomon:
The problem I have is that I haven't found a way to do that :/, I can specify how to render a node, but AFAIK I cannot inject references to resources... I haven't found any example similar to this one, as they all injected everything together, so no idea.

Also I understand there should be css work to adapt it to the rtd theme, but I have no knowledge on fronted.

Blendify:
Please read the sphinx application docs (sphinx.app) which has some tools for injecting css and js script files

PR #418, opened 2017-04-28
Cleanup README.rst:

NickVolynkin:
* Refactor list items.
* Use autonumbering in lists to prevent errors.
* Indent code inside list items to make it belong to those items.
* Use inline-code style for filenames.
* Update link to a pull request with new repo path.
* Fix a few typos and punctuation.
* Use .. code:: bash for proper highlighting
* Use # for comments in bash

PR #433, opened 2017-06-16
Finalized #418

benjaoming:
This finalizes comments regarding indents by @agjohnson on #418

In an ideal world where everyone knows exactly what's the most efficient thing to do, #418 would have just had PRs against @NickVolynkin branch rather than all the comments about indentation back and forth :)

(please don't regard the above as some smart-ass comment, I'm trying also to teach myself to spend less time commenting and more time PRing in other reviews, and I know it's hard because I fear getting sucked in and ending up spending half the evening with someone elses PR)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -42,23 +42,21 @@ In your ``conf.py`` file:
 
 .. code:: python
 
-    import sphinx_rtd_theme
-
-    html_theme = "sphinx_rtd_theme"
-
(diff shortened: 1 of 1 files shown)


benjaoming:
Oh yeah, this also fixes the merge issue in #418, that's why I didn't PR against the original PR branch :)

PR #438, opened 2017-06-29
Add a version selector dropdown to the main menu

agjohnson:
This adds a version selection dropdown to the menu. It is used to redirect on
change to the version URL. The ``versions`` theme option is added and is used to
add a mapping of version slugs to URLs.

Currently, this is going to be used in pydoc.io. The implementation will change
the javascript slightly, to accept json data instead of a blob of HTML.

This will currently be disabled on Read the Docs, as we are rethinking what the
version selection menu looks like.
version-select


Blendify:
For simplicity can we get rid of the unrelated changes? Opening these as separate PRs if necessary. All in all, I can see this being helpful rather than leaving it up to users to write there own. But maybe instead of putting it in the `wy-side-nav-search` maybe it should go into `rst-versions`?

ericholscher:
@Blendify can you clarify the `But maybe instead of putting it in the wy-side-nav-search maybe it should go into rst-versions?` part of your comment?

ericholscher:
Oh, I see -- it's in that block in the sidebar. +1 on adding it outside of the search div.

agjohnson:
This isn't a problem of the version selector being placed in the wrong element, but the element was poorly named initially. `.wy-side-nav-search` contains not just the search, but the logo, project name, and version. So -1 from me on changing this here, I think it's out of scope for this feature.

Blendify:
For myself, I prefer it down there. This is where we wrote our custom version switch for my other project.

Blendify:
> This isn't a problem of the version selector being placed in the wrong element, but the element was poorly named initially. `.wy-side-nav-search` contains not just the search, but the logo, project name, and version. So -1 from me on changing this here, I think it's out of scope for this feature.

The feature itself is fine and we should merge this. Although I think after we merge this we need to make a final decision on how this will work on readthedocs and where in the html template it should go. We should also add support for a language drop down. Also it seems you ran grunt before committing to the branch so those changes need to be reverted before merging.

VSEphpbb:
Yes please 👍

Blendify:
Ok, so they are getting back from a month or so of vacation but after other important task get done we can have a look at this.

baylessj:
Just dropped these changes into some docs that I'm building locally, and I'm running into a few issues. They are likely due to me not configuring my project correctly, but I'm not really sure how this is supposed to work after reading this thread and the changes to the docs. For reference, I am used to using Sphinx-Versioning (https://robpol86.github.io/sphinxcontrib-versioning/index.html) to achieve this same goal.

In Sphinx-Versioning, each version is a completely separate project, with separate conf.py files and such. Is the intent to do the same with this feature, where each project has the same versions dictionary and then routes to the other's url? That seems to be a proper way of handling it, with the exception of building docs locally. If there was any way to specify relative links instead of urls for the versions that would make it much easier to test.

Additionally, the dropdown box looks much different and generic compared to the example in the original post here. Granted, I have modified a decent bit of the theme's color scheme so that's a very likely source of the disparity, but I have not touched any other parameters or characteristics of the theme besides the colors. Any ideas as to what could cause the differences, or is it more than likely due to my own meddling? A picture for reference:
differentdropdownappearance

agjohnson:
I'd like to rethink how this could be used for RTD. We have some thoughts on what a new menu is going to be like, but i do like having an additional pattern like this for switching versions on RTD.

Eventually, the menu is going to at least be API driven and not just a blob of HTML handed out by RTD.

billsacks:
I'm curious if there are any plans for this or similar functionality to be merged to the master version of this theme. I'm maintaining a few projects where we are hosting our documentation on github.io (rather than readthedocs), but want versioning functionality. We are currently using this branch to do so, and I'm wondering if that is what we'll need to do long-term, or if there is some hope that we'll be able to use master for this functionality at some point.

agjohnson:
No, I don't think this feature will be merged as-is. The problem with this approach so far is that this is a build-time only feature, meaning as soon as the version list is updated, this menu is inaccurate. We don't want to continue adding features like this, however our API is not where it needs to be to support a feature like this.

I'm going to close this for now.

billsacks:
@agjohnson thanks for your reply - and thank you for all of your work on this project.

FYI, and in case it happens to be useful to you: We use this version of the theme with a bit of added JavaScript code that sets the versions dynamically. To add a version, the only thing needed is to add a line to a json file in the build directory – no rebuild is needed. For details of what we do, see https://github.com/ESMCI/cime/pull/3439 - both the files changed and the initial comment in the PR. (The json file I'm referring to looks like this: https://github.com/ESMCI/cime/blob/f8dba01bbdfbf0c86e132be08490ae001815c64d/versions/versions.json .) (I'll be honest, though: I am far from an expert in web development and don't totally understand how this works: I inherited it from others.)

I'll also note, in case it's useful to others, that we're maintaining this branch in a fork (https://github.com/ESMCI/sphinx_rtd_theme/tree/version-dropdown-with-fixes), so that we can cherry-pick any critical bug fixes into this branch (currently just https://github.com/ESMCI/sphinx_rtd_theme/commit/05a46f571ecc868dd0176ae89642b9e6a7924fc2).

kapooramit:
Hi @agjohnson 

I am trying to do this and copied all the changes you mentioned from the commits to my local fork of the theme. However, the only thing I see is name of the theme version I am using ( 1.0.1alpha1 ). I was expecting to see a drop-down with names of versions added in the conf.py versions option - maybe the links that won't work because I haven't setup the versions.

My question is - do I need the versions folder or things setup like @billsacks mentioned in his fork of the theme or is there another way of doing this that you originally intended to work?

Thank you!

PR #443, opened 2017-07-21
Reflect language in lang attribute.

halfninja:
The basic Sphinx templates check `language` and use it to set `<html lang>`, so I think it makes sense to do this here. A few sites like Blender documentation have multiple languages, and at the moment Google will return results from all the languages even when asking for English results. With a proper `lang` I think it will be able to return appropriate results.

I'm not a Jinja ninja so let me know if there's a more concise way to do the if/else. I've built the demo site to check that it works (defaults to `en`, and changes when `-A language=fr` is passed)

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -8,8 +8,8 @@
 {%- endif %}
 
 <!DOCTYPE html>
-<!--[if IE 8]><html class="no-js lt-ie9" lang="en" > <![endif]-->
-<!--[if gt IE 8]><!--> <html class="no-js" lang="en" > <!--<![endif]-->
+<!--[if IE 8]><html class="no-js lt-ie9" lang="{%if language == None %}en{% else %}{{ language }}{% endif %}" > <![endif]-->
+<!--[if gt IE 8]><!--> <html class="no-js" lang="{%if language == None %}en{% else %}{{ language }}{% endif %}" > <!--<![endif]-->
 <head>
   <meta charset="utf-8">
   {{ metatags }}
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -6,10 +6,11 @@
 {%- else %}
   {%- set titlesuffix = "" %}
 {%- endif %}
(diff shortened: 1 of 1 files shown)


Blendify:
Hi thanks for the patch I opened a PR a while ago with the same idea (#406) However @Minecrell pointed this out:

>While this is valid if you use one of the languages with a two letter locale code (e.g. `de`) it won't work for locales like `pt_BR` or `pt_PT`. It's possible that we'd just need to replace the underscore with a hyphen but I'm not entirely sure either. (See https://www.w3.org/International/articles/language-tags/)

As a solution, we found that you can use http://jinja.pocoo.org/docs/2.9/templates/#replace but I never got around to updating the PR. If you want to talk you can poke me in IRC (sorry I missed your message).

halfninja:
Aha, don't know how I missed that one. Thanks, I'll experiment with replacing the underscore.

halfninja:
This change does the job. Also tidied up the syntax a little bit.

sphinx_rtd_theme/layout.html
```diff
@@ -6,10 +6,11 @@
 {%- else %}
   {%- set titlesuffix = "" %}
 {%- endif %}
+{%- set lang_attr = 'en' if language == None else (language | replace('_', '-')) %}
```
ericholscher: I wonder if this catches the proper language in most cases. In particular, what happens if `language` is undefined? I believe jinja has a check for this. Have you tried building sets of docs with this patch and different language variables defined?

halfninja: I've checked the output of

    sphinx-build demo_docs/source demo_docs/build
    sphinx-build demo_docs/source demo_docs/build -A language=pt_BR

In the first case language is None so `en` is used. I can update it to handle things like empty string if we're expecting people to pass `-A language=`. I was taking a cue from the official basic theme which just checks None-ness (though it uses `is none` which equivalent but looks like the more "correct" way to do it, so happy to update the syntax there).

https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html#L120

ericholscher: Ah, makes sense :+1:

ericholscher:
Thanks!

Issue #441, opened 2017-07-17
Better JavaScript syntax highlighting


TimKam:
The Sphinx Read the Docs theme could do better in highlighting JavaScript code:
![js_syntax_highlighting](https://user-images.githubusercontent.com/5434565/28287477-23a1d1e2-6b3c-11e7-84d3-3398c35036ad.png)

Sphinx/Pygments already provides the necessary classes:
![classes_highlight](https://user-images.githubusercontent.com/5434565/28287530-5189e522-6b3c-11e7-92f8-f96b6d11ecf2.png)

So this should be merely a matter of adding additional styles.

If this makes sense to you, I could try to find the time for creating a PR.

Blendify:
PRs are welcome anytime.

ericholscher:
Yep, this would be a great addition!

ItsPugle:
@TimKam What Atom UI/Syntax do you use?

Blendify:
https://github.com/rtfd/sphinx_rtd_theme/pull/448 should fix this.

skaller:
Could I have some feedback on this please? In my case, it isn't JS that's the problem, but other Pygments lexers. With "classic" html_theme the colours are strong, the bolding good, and the sizes slightly bigger, and the background different from RTD. I don't mind the size change or background difference much, but the Pygments colours are basically clobbered.

The bolding is unnoticeable and almost all the colour is bled away, except the disgusting bright red used for string literals.

This is running the online version of RTD. Shouldn't the theme respect the Pygments colours?

The problem I have is that the tokenisation scheme used doesn't really fit my language or what I want to see, so I have to fudge a bit to get the appearance right, but nothing works on RTD. 

My language (Felix) isn't the only one impacted. C++ code is similarly impacted. Comments get faded to almost unreadable status, colour is bled away.

Blendify:
The reason this does not work is that the HTML is not aware of the Pygments CSS #448 was going to fix this but this change is now included in a larger set of changes that can be found in this PR: #467. There are two temp fixes:

1. Use `pip install git+https://github.com/rtfd/sphinx_rtd_theme.git@6da7eb974754582353e80eada1151751745e185c`

2. Add `http://www.sphinx-doc.org/en/stable/extdev/appapi.html#sphinx.application.Sphinx.add_stylesheet` to the conf.py to link the css file manually.

skaller:
Ok thanks!

[...]

PR #444, opened 2017-07-27
#441 add initial version of styles for JS highlighting

TimKam:
I added a first proposal for JavaScript highlighting. It's based on the default syntax highlighter of the Atom text editor.
However, I identified a set of issues with the JavaScript lexer of [Pygments](http://pygments.org/):
* ``const`` should be class ``.kd``
* arrow function`` =>`` and spread operator ``...`` should be class ``.o``
* functions should get own class
* all global variables and their properties should be added to class ``.nb``
* difference between names after ``var`` + ``let`` vs. ``const``

I plan to create a fix for Pygments eventually.
The question is: Should we wait for a Pygments fix or merge the best we can get until then as soon as possible?


Blendify:
> I plan to create a fix for Pygments eventually.
The question is: Should we wait for a Pygments fix or merge the best we can get until then as soon as possible?

I would rather see this fixed in Pygments itself rather complicating our theme. Plus this will allow others that do not use our theme to use this syntax highlighting style.

Note that currently, Payments does not work at all with the theme, #448 fixes this.

TimKam:
Ok, the missing style sheet link explains a lot ;-)
I'll try to fix the remaining issues in Pygments then and close this one for now.

PR #446, opened 2017-07-28
Refactor layout.html

Blendify:
- Make the template more similar to the basic sphinx template
  - Use macros for css & js
- Set lang for main html tag
- Give sphinx conf control over things like page incoding
- add vars needed for other builders like epub
- and more + cleanup


Blendify:
> introduces a bunch of variables that aren't user configurable or documented. This definitely needs more docs

Some of these (maybe all) are used internally by sphinx

Blendify:
Planning to do something different here

PR #447, opened 2017-07-29
Fix html head tag

Blendify:
- Add html_tag (used by epub builder: http://www.sphinx-doc.org/en/stable/_modules/sphinx/builders/epub3.html?highlight=html_tag
- Add missing html tag? Not sure what is going on here. ALso add lang_attr to that tag.


Blendify:
Question do we even need the IE conditionals? The `.no-js` class is not used in WYRM or our theme.
None of the other sphinx templates do this either.

Blendify:
See #467

PR #449, opened 2017-07-30
Make charset user configurable

Blendify:
- `use_meta_charset` is used by epub3 builder


Blendify:
Planning to do something different here

PR #450, opened 2017-07-30
Remove 'extra_css_files'

Blendify:
`extra_css_files` was add in https://github.com/rtfd/readthedocs.org/commit/e759c50697c5c9b210f5b7ea2978e24b25aef901 but serves no purpose (not sure why it was added intially) users should use `Sphinx.add_stylesheet` instead.


Blendify:
Actually looking at this there are several projects that use this. Is there a way to mark this as deprecated?

agjohnson:
Yeah, this is still in use by forks, so I wouldn't want to deprecate it either.

htgoebel:
Well, this is used for add style sheets, which might be required to fine-tune the appearance where rst is coming to its limits.

Blendify:
This should still be removed to conform sphinx's new method for adding css files.

https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_css_files

This can be removed when we require sphinx2, there does not seem the be a way to deprecate this first.

humitos:
I think we can target this for 3.0 and start raising a warning on 2.x.

humitos:
I just wanted to note that we had some issues when releasing 2.0 because of this chunk of code. I'm fine removing this since it adds no value over the built-in Sphinx feature.

Blendify:
Updated the branch and waiting on approval for 2.0

humitos:
Thanks for the update @Blendify. I don't think we will include this in 2.0 since we are on `rc4` already and this will break projects already using this config. However, I'm 👍🏼 to remove this on 3.0 together with other breaking changes we are planning to do as well.

agjohnson:
We definitely need to do a deprecation warning before removing an option like this.

Though this is a configuration we shouldn't have added, I also personally don't see much value in a deprecation like this given it's such a sideways change -- the code here isn't really presenting a problem for our maintenance and the user is not really better off being forced to change either.

If we do a deprecation in 3.0, we should emit a deprecation warning starting in 2.0

Issue #312, opened 2016-07-20
Option not to include headings in the toctree


akien-mga:
This is a follow-up to #277 which was just fixed by adding the `navigation_depth` option. I've noticed that though this option is quite useful to have, it doesn't fully address my use case when using `collapse_navigation = False` (i.e. tree-like structure).

What I'd like to have as an option is to be able to limit the table of contents sidebar to sections and pages, without showing the various levels of headings/subsections for pages.

Basically I'd like the toctree to only display, where `+` shows an expandable item, and `o` shows a non-expandable item.

```
+ Section 1
   + Subsection 1.1
      o Page 1.1.1
      o Page 1.1.2
    o Page 1.2
```

instead of:

```
+ Section 1
   + Subsection 1.1
      + Page 1.1.1
         o Heading2 1.1.1.1
         + Heading2 1.1.1.2
            o Heading3 1.1.1.1.1
         o Heading2 1.1.1.3
      o Page 1.1.2
    + Page 1.2
       o Heading 1.2.1
```

Some examples to be more explicit:
- This looks good in the sidebar, as what is uncollapsed shows only subsections: http://docs.godotengine.org/en/latest/tutorials/engine/_engine.html
- This is not good, as what is uncollapsed shows individual pages as uncollapsable, which gives the impression that they are subsections (and thus that there are an overwhelming amount of tutorials to read): http://docs.godotengine.org/en/latest/tutorials/step_by_step/_step_by_step.html
- Another example where the verbosity can be in our opinion quite misleading to users; all the uncollapsed items here are pages or h2/h3, even though they look like subsections: http://docs.godotengine.org/en/latest/reference/compiling_for_universal_windows_apps.html#compiling

Using `collapse_navigation = True` works this around thanks to the properly defined `:maxdepth:`, but we really want to use this tree-like navigation. What we'd need for our use case is an option that ignores page headings (h2 and later) when generating the toctree, but only takes into account the page and section titles (h1).

The `navigation_depth` parameter does not fully address our use case, as we need variable depths depending on the section (e.g. our "Learning step by step" section needs a depth of 2, while our "Engine" section with its subsections needs a depth of 3).

Hope this makes sense, and thanks for your work on this theme :)

reduz:
To put a little more context, a single page such as this:

![image](https://cloud.githubusercontent.com/assets/6265307/17005514/0d868f72-4eb1-11e6-8a24-c98d90a96af7.png)

[...]

PR #456, opened 2017-09-01
Add titles_only option to the toc

pcvonz:
I believe this fixes issue: #312 

It makes it so the table of contents only generates entries for the titles, instead of each heading within a page. From the sphinx doc: "If you want only the titles of documents in the tree to show up, not other headings of the same level, you can use the titlesonly option:"[1]

[[1] Sphinx doc on toctree directive](http://www.sphinx-doc.org/en/stable/markup/toctree.html)


sphinx_rtd_theme/layout.html
```diff
@@ -131,7 +131,7 @@
               toctree is empty. Skip building this for now.
             #}
             {% if 'singlehtml' not in builder %}
-              {% set global_toc = toctree(maxdepth=theme_navigation_depth|int, collapse=theme_collapse_navigation, includehidden=True) %}
+            {% set global_toc = toctree(maxdepth=theme_navigation_depth|int, collapse=theme_collapse_navigation, includehidden=True, titles_only=theme_titles_only) %}
```
akien-mga: I think the two-space indentation was mistakenly removed here.

sphinx_rtd_theme/theme.conf
```diff
@@ -12,3 +12,4 @@ display_version = True
 navigation_depth = 4
 prev_next_buttons_location = bottom
 canonical_url =
+titles_only = True
```
akien-mga: I guess the default value should likely be kept to `False` as it used to be?

pcvonz: Ok, I looked into this and that option is being read in as a string. So, it doesn't matter if you put "True" or "False", it'll still read True since they're both strings. An empty string will register as False though. 

See here: https://github.com/sphinx-doc/sphinx/blob/cb0220de1a0d8acd00292b898d2ef03b588179b0/sphinx/environment/adapters/toctree.py#L180

The True / False setting in conf.py works because those are actually read in as booleans.

pcvonz:
I did some digging and found that the way the config parser in sphinx work is based on strings, booleans don't regsiter. From what I understand, all the boolean values in the theme.conf file that are set to a value are going to be interpreted as True. Unless the string is empty, that would return false. So, the titles_only option I've added in the theme.conf file wont interfere with the current style since by default it is set to an empty string which will return false.

All the other values that are set to False in the theme.conf file I believe are actually True.

pcvonz:
I did some digging and found that the way the config parser in sphinx work is based on strings, booleans don't regsiter. From what I understand, all the boolean values in the theme.conf file that are set to a value are going to be interpreted as True. Unless the string is empty, that would return false. So, the titles_only option I've added in the theme.conf file wont interfere with the current style since by default it is set to an empty string which will return false.

All the other values that are set to False in the theme.conf file I believe are actually True.

PR #466, opened 2017-09-18
Fix error in setup entry points

Blendify:
This was missing from b3225c556684a57e091ded887a80a9e72211004b

### DIFF
--- a/sphinx_rtd_theme/__init__.py
+++ b/sphinx_rtd_theme/__init__.py
@@ -3,7 +3,7 @@
 From https://github.com/ryan-roemer/sphinx-bootstrap-theme.
 
 """
-import os
+from os import path
 
 __version__ = '0.2.5b1'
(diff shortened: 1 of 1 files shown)


ericholscher:
Could use a comment explaining what it is doing.

sphinx_rtd_theme/__init__.py
```diff
@@ -3,13 +3,17 @@
...
+from os import path
 
 __version__ = '0.2.5b1'
 __version_full__ = __version__
 
 
 def get_html_theme_path():
     """Return list of HTML theme paths."""
-    cur_dir = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
+    cur_dir = path.abspath(os.path.dirname(os.path.dirname(__file__)))
```
ericholscher: This is still using `os`.

Issue #166, opened 2015-01-24
Choosing pygments monokai style


lifeBalance:
Following the instructions given in [the sphinx docs](http://sphinx.readthedocs.org/en/latest/config.html#confval-pygments_style), in my `conf.py` I set:

```
# The name of the Pygments (syntax highlighting) style to use.
pygments_style = 'monokai'
```

But it doesn't work. Any ideas why? Thanks.

patricksnape:
Can confirm that this doesn't work (Sphinx==1.3b2, sphinx-rtd-theme==0.1.6, Pygments==2.0.2) using the Napolean extension, on Ubuntu 14.04 x64.

frol:
It is because RTD packages pygment styles inside its theme.css:
https://github.com/snide/sphinx_rtd_theme/blob/master/sass/theme.sass#L45
Here is the file (https://github.com/snide/wyrm/blob/master/sass/wyrm_addons/pygments/_pygments_light.sass)

lifeBalance:
Thanks frol. But I don't get it. How can I change the pygments style?

frol:
@lifeBalance While it has to be fixed in RTD theme, there is a workaround.
Create `source/_templates/layout.html` tempate with the following content:

```html
{# layout.html #}
{# Import the theme's layout. #}
{% extends "!layout.html" %}

{% set css_files = css_files + ['_static/pygments.css'] %}
```

Then your `pygments_style` variable in the `conf.py` will take effect.

oliver-sanders:
Note that the workaround https://github.com/rtfd/sphinx_rtd_theme/issues/166#issuecomment-73278320 results in a mixed style. Attributes like font weight / style will likely leak from the RTD style into the pygments style.

PR #467, opened 2017-09-19
Refactor HTML templates

Blendify:
This extends the default sphinx "basic" theme anything that is not there is added to extra head.

Fixes:  #166:


Blendify:
Note: this is untested code and I need to make sure it works still.

Blendify:
@ericholscher @agjohnson can you have a look at this PR

ericholscher:
Is the idea that the deleted bits are already in the basic theme, and we inherit them?

ericholscher:
This seems like a really large change, tying ourselves to any additions in the Sphinx base theme. I imagine there could be many unindented outcomes.

Blendify:
> Is the idea that the deleted bits are already in the basic theme, and we inherit them?

Yes

Possible but doubtful, sphinx tries to keep this compatible since other theme do it to see: https://github.com/bitprophet/alabaster/blob/master/alabaster/layout.html

Blendify:
@agjohnson @ericholscher what is the status here? I really think that this is the better way to go as the old way has already caused several errors that this could prevent.

Juanlu001:
This pull request fixes the issues we're having with adding extra JavaScript libraries to display Plotly notebooks in Read the Docs, see https://github.com/spatialaudio/nbsphinx/issues/128. There is only one issue with vertical space in the header:

| master | pr |
|-----------------|---|
| screenshot-2017-11-18 poliastro - astrodynamics in python poliastro 0 7 dev0 documentation 4 | screenshot-2017-11-18 poliastro - astrodynamics in python poliastro 0 7 dev0 documentation 5 |

The rest looks exactly the same and it would be great to have this merged :+1:

Blendify:
I will have a look as soon as I can.

oliver-sanders:
js files in the `script_files` jinja2 list seem to be linked in twice, once in the header and once at the end of the document body.

Blendify:
Note that to get exact behavior as before https://github.com/sphinx-doc/sphinx/pull/4245 needs to get merged. But this is a minor thing and should not stop this PR.

sphinx_rtd_theme/layout.html
```diff
@@ -1,43 +1,13 @@
...
-  {% block htmltitle %}
-  <title>{{ title|striptags|e }}{{ titlesuffix }}</title>
-  {% endblock %}
 
-  {# FAVICON #}
-  {% if favicon %}
-    <link rel="shortcut icon" href="{{ pathto('_static/' + favicon, 1) }}"/>
-  {% endif %}
-  {# CANONICAL URL #}
   {% if theme_canonical_url %}
```
Blendify: Note that this can be removed also if https://github.com/sphinx-doc/sphinx/pull/4193 gets merged

Blendify:
@Juanlu001 do you mind testing again?

Juanlu001:
@Blendify the background gray stripe leveled up with the white body, but the whole block is still displaced. I attach current capture.

screenshot-2017-11-27 poliastro - astrodynamics in python poliastro 0 9 dev0 documentation

Juanlu001:
You can test it by yourself by building these docs: https://github.com/poliastro/poliastro/tree/master/docs

jessetan:
The gray stripe was two blocks of "related" stuff that was inherited from the basic theme, even more visible on less wide screens:
inherited blocks
Since this functionality is already covered by the sidebar and the prev/next buttons, I've removed them.

sphinx_rtd_theme/footer.html
```diff
@@ -8,44 +8,29 @@
...
-    {%- elif last_updated %}
-      {% trans last_updated=last_updated|e %}Last updated on {{ last_updated }}.{% endtrans %}
+      {%- endtrans %}
     {%- endif %}
-
     </p>
   </div>
 
-  {%- if show_sphinx %}
-  {% trans %}Built with <a href="http://sphinx-doc.org/">Sphinx</a> using a <a href="https://github.com/snide/sphinx_rtd_theme">theme</a> provided by <a href="https://readthedocs.org">Read the Docs</a>{% endtrans %}.
```
jessetan: @Blendify why was this removed? When building the demo docs, this does not show up anymore, so I don't think it is inserted elsewhere.

jessetan:
@Blendify One thing that this changes is using XHTML Transitional doctype instead of HTML5 doctype. I can't fully oversee the effects, but it might influence the way older browsers (IE) render things. We can change the doctype by redefining the `doctype` block, but changing the`<html>` element is only possible starting from Sphinx 1.6 (after [this commit](https://github.com/sphinx-doc/sphinx/pull/3651/commits/76ff2c01351a02b49b2941ab7c0984b31d051531#diff-a58d9f292d96b91ce96d485163c52d90R114))

jessetan:
The search page has some extra text, currently it looks like this:
current search
Refactored, it uses the same content as the Sphinx theme, which has a search box on the results page:
search refactored
Given that the RTF theme always has the search box in the navbar, maybe the old way is better?

sphinx_rtd_theme/footer.html
Blendify: I thought it was inherited from the main layout file but I think that is not working. I can add it back.

Blendify:
@ericholscher @agjohnson what do you think about making the search page match one that looks like https://alabaster.readthedocs.io/en/latest/search.html I am thinking about making this a separate PR because it would mean that we get rid of the licence issue of having Sphinx code in our repository. See https://github.com/rtfd/sphinx_rtd_theme/issues/283

ericholscher:
That would def. be a new PR. Our search stuff is complex, and I believe @agjohnson has plans to kill our dependence on the Sphinx search logic.

Blendify:
Would be interesting to hear more about because we still need to support
people using the theme not on readthedocs.

On Dec 18, 2017 6:04 PM, "Eric Holscher" <notifications@github.com> wrote:

That would def. be a new PR. Our search stuff is complex, and I believe
@agjohnson <https://github.com/agjohnson> has plans to kill our dependence
on the Sphinx search logic.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/rtfd/sphinx_rtd_theme/pull/467#issuecomment-352586096>,
or mute the thread

jessetan:
I've made the change to use the same DOCTYPE as the theme does now, which should prevent rendering differences (if any) from Sphinx preferring XHTML.
We are still missing some IE-specific classes on `<html>`, but since there is no CSS specifically targeting IE8, I think its safe to leave out.
```html
<!--[if IE 8]><html class="no-js lt-ie9" lang="{{ lang_attr }}" > <![endif]-->		
<!--[if gt IE 8]><!--> <html class="no-js" lang="{{ lang_attr }}" > <!--<![endif]-->
```

ericholscher:
Just as a note, I think this is a _major_ change to the theme, and will require a lot of work in order to release properly. In particular, it ties us to the underlying Sphinx version in a lot of ways, and it will need to be tested across a bunch of different scenarios.

I'm imagining that this will be a **2.0** release of the theme.  It will need to be tested across all supported Sphinx versions. It will also need to be tested across those Sphinx versions paired with RTD includes. Specifically, we have a lot of custom JS & CSS that could be changed by the underlying Sphinx HTML/CSS changes that come through, and we'd also be tying ourselves to supporting all new Sphinx features in the theme as soon as they are released in Sphinx itself. 

As I was before without a lot of explanation, there is a huge amount of complexity that comes from this somewhat simple change. This PR should definitely _reduce_ the amount going on, including not changing how we're doing search, so that we can properly test this and make sure what is coming across from Sphinx fully works in all the use cases that we need to support.

jessetan:
This PR does fix some issues, in addition to delegating some stuff to the Sphinx theme.
Perhaps we can separate the fixes into other PRs (I think these mostly relate to where script tags are included), so this PR will not have functional effects, only remove things that we can inherit from the base Sphinx theme. This will give us some time to come up with a proper strategy to handle this PR, since it affects more than just the theme.

Blendify:
Yes, I think this should be split up. It has become quite large and would be easier to split up. Also with Sphinx 2.0 coming in the next couple of years with some major changes likely. (we already know that `css_files` will be removed) maybe it would be better to drop version of sphinx lower than 2.0 at that time. We can save all the major (backward compatibility breaking) changes to the theme until then.

Issue #397, opened 2017-03-15
Sidebar does not reliably open from L3/L4 links


mattklein123:
Possibly related to https://github.com/rtfd/sphinx_rtd_theme/issues/290

In the [Envoy docs](https://envoyproxy.github.io/envoy/index.html) if a user clicks on any :ref: links that point to L3 or L4 headings, the sidebar does not reliably open to the L3 or L4 heading.

Repro:
1) On this page: https://envoyproxy.github.io/envoy/intro/arch_overview/listeners.html
2) Click on "sniffing" link: https://envoyproxy.github.io/envoy/intro/arch_overview/mongo.html#arch-overview-mongo

Expected: Open left side sidebar down to MongoDB

Actual: Only opened to L2.

NOTE: This bug happens all over the documentation. Sources are here: https://github.com/envoyproxy/envoy

petermd:
I think afaf0dc9bddf19e34658e661b3c3c2a7d4dfa265 adds a second way to search for the matching anchor, but these anchors (`arch-overview-mongo`) don't exist? 

even if the `link` array is still empty the code in `nav.reset` still removes the `current` style from L2/L3/L4 but never re-adds it (causing the side-bar to close)

only removing `current` if there is a link located _should_ fix the issue eg something like 42e90cdd64983633733eb6f636819992dee65e55

mattklein123:
Thank you @petermd! Hero!

mattklein123:
@petermd I tried out your commit here: https://github.com/envoyproxy/envoy/commit/1abdeb25081d1a5267482469ad422b5b8cd71c7e

And it still seems to be not working. Did I setup requirements.txt the right way? It seemed to clone your branch.

petermd:
@mattklein123 right, so I think a full change needs a `grunt` run to create the static export, not sure what the process is for a PR (there are several re-generated files but only the `theme.js` is needed is this case)

I added 09da15bb9b0cf26543f789c0c661dce7924765b4 that you can use to test the fix from the envoy side.

mattklein123:
@petermd just tried it and it definitely seems to fix our issue! 🎉 Thank you so much!

@Blendify any chance of getting this reviewed / merged? Otherwise I can setup a fork. Thank you!

Blendify:
I can take a look Wednesday.

mattklein123:
Awesome thank you @Blendify!

mattklein123:
[...]

PR #472, opened 2017-10-04
Only change current if a matching link is found

petermd:
Fix for #397 to only reset `current` style and re-apply if a new candidate link has been found.

Envoy docs have links with unresolved anchors, this fix avoids collapsing the L3/L4 headers if not matching link for the anchor is found.

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -108,13 +108,15 @@ function ThemeNav () {
                   // Try again with the closest section entry.
                   link = $('.wy-menu-vertical')
                     .find('[href="#' + closest_section.attr("id") + '"]');
-
                 }
-                $('.wy-menu-vertical li.toctree-l1 li.current')
-                    .removeClass('current');
-                link.closest('li.toctree-l2').addClass('current');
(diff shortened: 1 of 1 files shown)


Blendify:
Lgtm but I want to leave it up to @agjohnson because I do not do much Javascript.

mattklein123:
@Blendify @agjohnson I'm fine going ahead and forking, however I'm a little confused about the "grunt" aspect of this. When I tried this patch without whatever extra stuff @petermd did it did not work for me. Can you advise so I can do the right thing in the fork? Thank you.

Blendify:
The correct thing to do is run `grunt build` the reason we say not to do that with the Pull Request is this auto-generates files which can be an issue with multiple pull requests. Although because it works for you and you have been using it for a while I will just merge it. However, I will ask for a favor, can you submit a new pr based off after the merge that just runs `grunt build`. I would do it but I only have a chromebook atm and wont be able to do it.

mattklein123:
Thanks. @petermd any chance you can do the extra PR?

PR #476, opened 2017-10-13
Follow-up grunt for #472

petermd:
Ran `grunt build` to re-generate the static js as per #472.

Please note: I _did not_
- update any version number
- include the re-generated css as there's a non-obvious diff (formatting of generated css?) in the `css` and `css.map` files but since the change has no CSS in it, I figured it was best to omit it

Let me know if thats ok


Blendify:
No, we need the css too from other PRs

petermd:
K, figured out my issue, had `sass 3.5.2` installed which changes the default precision hence the large diffs. Tried`bundle exec grunt build` instead but the `Gemfile.lock` is pretty old so not sure there's a simple way to get a repeatable env? (if you have newer versions of the dependencies)

In any case, re-installing the required gems and re-running has picked up the CSS changes to `theme.css` - let me know if you need anything else.

Blendify:
Thanks

mattklein123:
Thank you @petermd !!! FYI https://github.com/envoyproxy/envoy/pull/1860 💯

PR #487, opened 2017-11-22
Apply clearfix on breadcrumbs

### LINKED ISSUE #486: Long section header flows "View page source" link to ugly position
If a section header has a relatively normal length, the "View page source" link (could also be "edit on GitHub" etc, whatever is in `li.wy-breadcrumbs-aside`) floats to the right of the page at the same vertical position:
normal

If the section header is very long, the "View page source" link floats over the horizontal rule:
long

If the Prev and Next buttons are also present (by setting `prev_next_buttons_location` to `top` or `both` as implemented in #331), the "View page source" link floats to the right of these buttons:
long buttons

Root cause is lack of CSS clear on the `<div>` containing the buttons and the `<hr>`. More elegant solution is to apply a clearfix on the `<div>` containing the breadcrumbs.

jessetan:
Fixes #486

Old behaviour: 

 long
long buttons

New behaviour:
long-fixed
long buttons-fixed

### DIFF
--- a/sass/_theme_breadcrumbs.sass
+++ b/sass/_theme_breadcrumbs.sass
@@ -1,3 +1,6 @@
+.wy-breadcrumbs
+  +clearfix
+
 .wy-breadcrumbs li
   display: inline-block
   &.wy-breadcrumbs-aside


Blendify:
LGTM would want one more review from either @ericholscher or @agjohnson before merging

ericholscher:
Also looks good to me, but will cede to @agjohnson who knows more about how we do our CSS.

ericholscher:
Alternately, to unblock this on @agjohnson, I'd be curious to hear why clearfix is the right solution to this. I'm happy to merge it, I just want to know it's the correct solution, and don't know enough to judge.

jessetan:
@ericholscher as the author of the PR, my reasoning for the clearfix is as follows:
The "View page source" link is placed to the right of the document title by floating it to the right. As the title becomes longer, its size will take precedence over the float, i.e. the title will be allowed to expand in width until it spans the entire page width, pushing the float downwards.
Any element below the title that does not have a CSS clear statement will be allowed to wrap around the float. This causes the buttons or the horizontal rule to appear next to the float, which is undesirable.
The clearfix inserts a `clear: both` after the containing div, which seemed more elegant to me than just adding `clear: both` to any element that can possibly follow the breadcrumb section.

An alternative to the clearfix would be to use absolute positioning with `right: 0` on the "View page source" link, which would probably cause the section title to linewrap earlier.
I'm fine waiting for the final say from @agjohnson (if this happens before the next release), since I'm not familiar with the reasoning behind the current implementation.

ericholscher:
@jessetan thanks for the explainer! This sounds like a good approach, and I'd be happy to merge it. I'll let @agjohnson comment here soon if he has feelings, otherwise I'll get it merged before next release :+1:

agjohnson:
Clearing floats here should be fine. Testing the normal case of short breadcrumb and link floating to the right for GitHub/Bitbucket/View source shows clearing div has no effect on positioning of button.

👍

ericholscher:
🎆

PR #489, opened 2017-11-22
Fix #485 Add Includehidden theme option

### LINKED ISSUE #485: includehidden option should be exposed in conf.py
The theme includes hidden index entries in the navigation bar by default due to `includehidden=True` in [layout.html#L134:116](https://github.com/rtfd/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L134:116)

In order to override that option, one has to put the entire layout file in the `_templates` directory just to change that one value to False.

There should be a way to set `includehidden` in conf.py

Blendify:
Fix: #485 Also, reorder the theme option to group them by similarity.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,9 +87,16 @@ file of this repository, and can be defined in your project's ``conf.py`` via
 .. code:: python
 
     html_theme_options = {
-        'collapse_navigation': False,
-        'display_version': False,
-        'navigation_depth': 3,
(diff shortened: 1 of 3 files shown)


ericholscher:
Looks great! Thanks for the doc update. We should really put this in RST & throw it on RTD, so we can just `.. include` the actual config options :)

Issue #491, opened 2017-12-01
Navigation sidebar scroll is janky when scrolling main content


jessetan:
The navigation sidebar scrolls along with the main content when it is scrolled. This scrolling is not as smooth as the scrolling motion of the main content. I can't easily reproduce this on video, but it can be reproduced like this:

1. Open a page that has a sidebar and main text content that does not fit on one screen (e.g. http://docs.readthedocs.io/en/latest/getting_started.html)
2. Scroll sidebar, observe smooth scrolling 
3. Scroll main content, observe smooth scrolling, but not in sidebar. This lags behind and is a bit janky

## Root cause
The scroll position of the sidebar is updated using a continuous check with an interval of 25 ms  for debouncing purposes (https://github.com/rtfd/sphinx_rtd_theme/blob/master/js/theme.js#L35).
This interval results in a max scrolling framerate of 40Hz, which is less than what most modern browsers target for scrolling.

## Solution
Use requestAnimationFrame to debounce the scroll event. This allows the browser to scroll the sidebar as fast as it is able to render. See #492.

## Side-effects
May be incompatible with browsers without support for requestAnimationFrame (IE < 10 ?), but [shims are widely available](https://gist.github.com/paulirish/1579671)

PR #492, opened 2017-12-01
Debounce scroll and resize through rAF

jessetan:
Use `requestAnimationFrame` instead of an interval to debounce scrolling and resizing events.

Makes sidebar scroll smoother when main content is scrolled, fixing #491.
Also removes the need for the interval firing every 25ms when nothing is scrolled.
I've added also changed handling of resizing the window for the sake of consistency, although I see no difference in behavior.

Uses [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) and [`Function.prototype.bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind). Both of these are supported by modern browsers. If backward compatibility (mostly IE < 10) is needed, shims are widely available for both.


ericholscher:
This looks lovely, that is one of my biggest pet peeves currently. Will give it a test!

agjohnson:
This looks like a great addition.

IE10 is not horribly important, it's only about 1% of our traffic. I'd like to know what the fallback is on <= IE10 is though. This seems it would just break? I'm fine with choppy scrolling there, no need to shim, unless its easier.

jessetan:
@agjohnson this PR will work for the following browsers (assuming MDN is correct):

- Chrome >= 24
- Firefox >= 23
- IE >= 10
- Opera >= 15
- Safari >= 6.1

For older browsers it will just break (i.e. no scrolling of the sidebar if you scroll main content), but support for IE 9 can be added by including a [rAF polyfill](https://gist.github.com/paulirish/1579671#file-raf-js) somewhere in the `<head>`. This replaces `requestAnimationFrame` with `setTimeout` on browsers where rAF is not available.

For even older browsers (IE < 9, Safari < 5.1), a [workaround for the lack of `bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind#Polyfill) can also be added to `<head>`.

I think that adding the rAF polyfill to head is worth it to prevent breaking IE 9, since it only adds 31 lines of code if not minified. I can add that to the PR if you want.
Do you care about even older browsers working or not?

agjohnson:
If there is anyway to write this in a conditional way, I think backwards compatibility is important for us. Not in that this feature needs to work, but scrolling can't break on older browsers. If this limits our use of bind, then that's fine.

I don't want to introduce another JS request, so if there is a way to shim inline with our current code, that is more attractive.

jessetan:
@agjohnson I've removed the use of `bind()` and added the polyfill inline. I don't have IE9 or lower here to test, but I expect it to work.

ericholscher:
Sounds good to me!

ericholscher:
Thanks for your work lately @jessetan & @Blendify -- y'all have been doing great stuff here :)

ericholscher:
Hmm. On a local test instance I have _no_ sidebar scrolling currently on latest Chrome. They are totally detached from either other :/

ericholscher:
Seems like this depends on the `sticky_nav` setting to be True (defaults to False) in order for the `enable()` call to actually happen.

PR #493, opened 2017-12-02
Bump to 0.2.5b2, for QA testing before release

ericholscher:


Blendify:
Looks good but there are two PRs I would like to get in before 0.2.5: #492 and #448 I need to do a little testing on #448 but I will try to have it done today or sometime this week.

PR #496, opened 2017-12-08
Fix theme logic and rebuild prod assets

ericholscher:


sphinx_rtd_theme/layout.html
```diff
@@ -159,7 +159,7 @@
 
       {# PAGE CONTENT #}
       <div class="wy-nav-content">
-        {% if (theme_style_external_links == 'True') %}
+        {% if theme_style_external_links %}
```
jessetan: @ericholscher I think this change won't work for all cases, just like mine was wrong.
There is a difference in the type of `theme_style_external_links` depending on if it is set in `html_theme_options` in `conf.py` or if the default is taken from `theme.conf`. In the first case it is a Python Boolean, in the second case it is a String with value of True or False.
Unfortunately I only tested with `theme.conf`.

It would work if this line is
```py
{% if (theme_style_external_links == 'True' or theme_style_external_links == True) %}
```

Your suggested change evaluates to True if the value in `html_theme_options` is `True` but also if it is not set in `html_theme_options`, because the default from `theme.conf` is "False".

ericholscher: Interesting. I'll go ahead and try that. Thanks!

ericholscher:
The JS changes were already reviewed, just not "built". :+1:

jessetan:
Sorry, my bad. I assumed building the JS was only needed when the theme is released, not when filing or merging the PR.

PR #499, opened 2017-12-12
Default stickynav to on properly.

ericholscher:
This is documented as being on,
but previously we weren't running `tobool` on it to check it properly.
This had the effect of always returning True unless the user set it to False in Python.

You can see the commit that changed it here: https://github.com/rtfd/sphinx_rtd_theme/pull/496/files#diff-1336f80431f476a232e3c48abb277c51R210

This properly defaults it to True,
so that navigation works :)


ericholscher:
@jessetan also ^^

PR #500, opened 2017-12-13
Move DOCUMENTATION_OPTIONS to external JS for purposes of CSP

remyabel:
Ref: https://github.com/sphinx-doc/sphinx/issues/3620
https://github.com/rtfd/sphinx_rtd_theme/issues/414

Currently the only thing that needs to be fixed for [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) support is removing this bit of inline javascript. This is probably not the best way to do things, but I'm opening this PR for feedback. Even if it doesn't get approved, suggestions would still be useful so I can use them for my own site.

### DIFF
--- a/js/documentation_options.js
+++ b/js/documentation_options.js
@@ -1,0 +1,9 @@
+var urlParams = new URLSearchParams(window.location.search);
+var DOCUMENTATION_OPTIONS = {
+    URL_ROOT: urlParams.get('URL_ROOT'),
+    VERSION: urlParams.get('VERSION'),
+    COLLAPSE_INDEX: urlParams.get('COLLAPSE_INDEX'),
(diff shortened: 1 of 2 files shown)


Blendify:
I would like to see sphinx themselves add support for this first see: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html#L89

Blendify:
Also, I think it should be possible to have the [jinja](http://jinja.pocoo.org/) parts fully in that linked js file by renaming it to a `js_t` file. For example, https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/static/searchtools.js_t

remyabel:
Interesting, did not know about `js_t`, thanks. Will close this for now.

PR #504, opened 2017-12-16
Remove wyrm_addons/pygments

### LINKED ISSUE #503: Remove use of wyrm codeblock
To better support Pygments we need to remove the use of `.codeblock` from wyrm and style code blocks ourselves.

Blendify:
Fixes #503 By removing wyrm_addons/pygments we get full support of using native pygments to style code blocks


Blendify:
This is not ready for review because I still need to fix things like text and spacing in our own sass

Blendify:
This is ready for review now

jessetan:
Looks pretty good as far as I can see, although the `line-height` feels a bit too much considering the font size and shape (I'm using macOS, which has Consolas). Previously the `line-height` was `normal` (== 1.2), now it is 1.5.
Screenshots below of possible other values. 1.5 looks better to me if the fontsize is a bit larger (14px?), but maybe I'm just nitpicking.

- `line-height: 1.2` (current value in 0.2.4)
  
- `line-height: 1.3`
  
- `line-height: 1.4`
  
- `line-height: 1.5` (new value)

sass/_theme_rst.sass
```diff
@@ -53,8 +50,41 @@
...
+  // Needs 100% width for line highlighting to work properly
+  div[class^='highlight'] td.code
+    width: 100%
+  .linenodiv pre
+    border-right: solid 1px lighten($table-border-color, 2%)
+    margin: 0
+    padding: $base-line-height / 2 $base-line-height / 2
+    font-family: $code-font-family
+    font-size: 12px
+    line-height: normal
```
jessetan: Perhaps `font-size` and `line-height` should be moved to a new CSS rule for line numbering and actual code. That will prevent them from going out of sync in future.

``` sass
pre.literal-block, div[class^='highlight'] pre, .linenodiv pre
    font-size: 12px
    line-height: normal
```

Adding `pre.literal-block` to this rule also aligns the font size for literal blocks, which look out of place now:
literal code block

sass/_theme_rst.sass
```diff
@@ -53,8 +50,41 @@
...
+      margin: 0
+  // Needs 100% width for line highlighting to work properly
+  div[class^='highlight'] td.code
+    width: 100%
+  .linenodiv pre
+    border-right: solid 1px lighten($table-border-color, 2%)
+    margin: 0
+    padding: $base-line-height / 2 $base-line-height / 2
+    font-family: $code-font-family
+    font-size: 12px
```
jessetan: Perhaps `font-size: $base-font-size * 0.75` ?
(Also in other places in this PR)

Blendify: This would be a new behavior correct? Seems fine but maybe this is better for another PR?

sass/_theme_rst.sass
Blendify: Done

sass/_theme_rst.sass
jessetan: The behavior would be the same (`$base-font-size` is 16px), but indeed perhaps better to move to a new PR that adds a `$code-font-size` in `_theme_variables.sass` so we can be sure we have consistent font sizes.

PR #512, opened 2017-12-18
Prevent footnote and citation references changing line spacing

### LINKED ISSUE #511: Footnote references increase line spacing
See attached image and note inconsistent line spacing: whenever a footnote reference is present, line spacing is larger than it normally is.

footnote references

jessetan:
Fixes #511.

Before:

footnote references

After:

footnote references pr 512


jessetan:
I have left the positioning and sizing as it was, so it looks the same as before the PR. This is inconsistent with the positioning and sizing of `<sup>` from wyrm, which uses `font-size: 75%` and `top: -0.5em`, would be good to align these values.

PR #513, opened 2017-12-18
Do not add border to active links with level >1

### LINKED ISSUE #459: Toggling current ("on") item in sidebar changes height
## Issue

When a toctree item is "current" in the sidebar, the `<a>` element is bold and has an extra `border-top` and `border-bottom` of 1 px. This causes the containing `<li class="toctree-l#">` to grow in height by 2 px.
Example:
menuitem-border

## Effect

Every time an item is clicked in the toctree, the item itself and the items below will change vertical position slightly.

## Cause
[_theme_layout.sass:81-82](https://github.com/rtfd/sphinx_rtd_theme/blob/master/sass/_theme_layout.sass#L81-L82):
```css
    border-bottom: solid 1px darken($menu-vertical-background-color, 20%)
    border-top: solid 1px darken($menu-vertical-background-color, 20%)
```
My guess is that these styles were intended only to apply to the first `li.current > a` (since the comment hints at this and there are overrides below), but only `li.toctree-l4 > a` has settings for `border-top: none` and `border-bottom: none`

## Suggested fix
Add `border-top: none` and `border-bottom: none`  for `li.toctree-l#.current > a` (with `#` 1--4)
OR
Set a fixed height for toctree `<li>` elements

jessetan:
Fixes #459 

Before:
before

After:
after


davidfischer:
👍

PR #519, opened 2017-12-20
Always call nav.enable() to make nav collapsing work

### LINKED ISSUE #481: Disabled collapse_navigation stops working when sticky_navigation is also disabled
How come these two settings don't work when both are set to false?
```
html_theme_options = {
     'collapse_navigation': False,
     'sticky_navigation': False,
}
```
`'collapse_navigation': False` and `'sticky_navigation': True` allows the user to collapse/expand all items in the sidebar. Which is what I want.

However, I do not want the sticky nav bar. But when I set `'sticky_navigation': False`, all the little + signs that allow the user to collapse/expand items simply go away as if `'collapse_navigation'` was set to True.

Is there a way to have collapsible navigation items without a sticky nav bar?

jessetan:
This used to depend on the sticky nav being enabled.
Fixes #481

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -15,17 +15,22 @@ function ThemeNav () {
         isRunning: false
     };
 
-    nav.enable = function () {
+    nav.enable = function (withStickyNav) {
         var self = this;
 
-        if (!self.isRunning) {
-            self.isRunning = true;
-            jQuery(function ($) {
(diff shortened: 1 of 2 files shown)


ericholscher:
Yea, I ran into this one myself, thanks for fixing it!

PR #520, opened 2017-12-20
Docs 2.0

Blendify:
This splits the docs from the demo so that we can now actually document
the theme and its parts.


ericholscher:
Also needs a requirements file or some other way to build it on RTD, so we get the proper extensions.

ericholscher:
Cleaned things up :+1:

ericholscher:
🎆 http://sphinx-rtd-theme.readthedocs.io/en/latest/

PR #488, opened 2017-11-22
Footnote citation spacing

jessetan:
Old behaviour:
old

New behaviour:
new

This PR removes the `margin-bottom` from (the wrapper around) citations and footnotes, which makes them look better.
To preserve space between footnotes and whatever comes next, I've added a `margin-top` to any element that follows.

Fixes #440

Issue #379, opened 2017-03-07
Change how we do packaging


agjohnson:
Currently, PRs are expected to `grunt build` themselves. This leads to a merge conflict problem when there are multiple PRs open. We should instead be handling build artifacts on release.

The workflow here would be that when it makes sense, a maintainer will release the package, either as a development release or as a prod package. The maintainer will run `grunt build`, check those artifacts into the repository, the commit will be tagged, and a release to go out to PyPI.

We can't remove storing artifacts as a number of folks pin their theme dependency here. We can remove the requirement that PRs run `grunt build`, but perhaps find another mechanism for keeping our version on conflict, such as using `.gitattributes` to use ours on merge.

jessetan:
Removing the need for PRs to run `grunt build` is a good idea to reduce merge conflicts. We could add instructions in a [PR template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/)

If building and committing the assets to the repo is needed for people who use master instead of a package from Pypi, this can be a periodic task for maintainers or automated using a bot and/or git webhooks.

agjohnson:
I think a release time process by maintainers makes the most sense. We can think about automating this later perhaps.

We're hoping to shorten the release cycle and make more minor releases with the theme. With that, we can just release more often, generate build assets on release, and perhaps just disregard the use case of running the theme from `master` -- running from master would effectively just be the same as running from the latest release.

jessetan:
Keeping the built assets up to date can be done with Travis as following (roughly):

- An `after_success` block is added to `.travis.yml` to call a bash script
- The script checks if the commit is not a PR commit and the affected branch is master. This is effectively a check for a merged PR.
- The script installs any dependencies and runs `grunt build` to create the build artefacts (located in `sphinx_rtd_theme/static`)
- The script commits these files to master using a properly secured access token

The process for a PR then becomes:

1. Authors creating a PR will be presented with a PR template that instructs them not to include the built assets (although nothing will break if they do)
2. Comments and review
3. PR is merged to master
4. Travis commits built artefacts to master, making master up to date with the changes from the PR

Inspiration:

https://github.com/cvan/webvr/blob/ca0e903/scripts/deploy.sh
https://gist.github.com/domenic/ec8b0fc8ab45f39403dd
https://benlimmer.com/2013/12/26/automatically-publish-javadoc-to-gh-pages-with-travis-ci/

[...]

PR #522, opened 2017-12-21
Run grunt

Blendify:
- Update grunt file with changes done to docs
- Update grunt sass to 1.0.0 to avoid warning message


Gruntfile.js
```diff
@@ -133,17 +133,17 @@ module.exports = function(grunt) {
       },
       /* Changes in theme dir rebuild sphinx */
       sphinx: {
-        files: ['sphinx_rtd_theme/**/*', 'demo_docs/**/*.rst', 'demo_docs/**/*.py'],
+        files: ['sphinx_rtd_theme/**/*', 'docs/**/*.rst', 'docs/**/*.py'],
```
jessetan: Since README is now also built, `README.rst` should be added to wacthed files.

Blendify: Can we just check the full docs file? `docs/**/*` wont this catch everything?

jessetan: README.rst is not in the `docs` dir, it's in the root.

sphinx_rtd_theme/static/js/theme.js
```diff
@@ -16,17 +16,22 @@ function ThemeNav () {
         isRunning: false
     };
 
-    nav.enable = function () {
+    nav.enable = function (withStickyNav) {
```
agjohnson: What is the purpose of this addition? If configurable, this needs more information or documentation.

sphinx_rtd_theme/static/js/theme.js
```diff
@@ -80,9 +90,16 @@ function ThemeNav () {
...
-        $("table.docutils:not(.field-list)")
+        $("table.docutils:not(.field-list,.footnote,.citation)")
             .wrap("<div class='wy-table-responsive'></div>");
 
+        // Add extra class to responsive tables that contain
+        // footnotes or citations so that we can target them for styling
+        $("table.docutils.footnote")
+            .wrap("<div class='wy-table-responsive footnote'></div>");
+        $("table.docutils.citation")
+            .wrap("<div class='wy-table-responsive citation'></div>");
```
agjohnson: These changes are also unrelated to the PR. Is there background on these changes anywhere?

This pattern always seemed strange, especially when we have full control of the Sphinx writer as a Sphinx extension. The most correct is to do this on writing, not on each view I think. It might be best to split this work out into a separate PR if it isn't already.

sphinx_rtd_theme/static/js/theme.js
```diff
@@ -170,7 +187,7 @@ function ThemeNav () {
 module.exports.ThemeNav = ThemeNav();
 
 if (typeof(window) != 'undefined') {
-    window.SphinxRtdTheme = { StickyNav: module.exports.ThemeNav };
+    window.SphinxRtdTheme = { Navigation: module.exports.ThemeNav };
```
agjohnson: This likely breaks on RTD. Has this been tested?

Blendify:
All the changes done to `theme.js` were made in separate PRs run git blame to see the changes.

sphinx_rtd_theme/static/js/theme.js
jessetan: This comes from #519. `enable()` was only called to enable sticky nav, while it was needed to decorate the navbar with the correct toggle options.
The extra bool was added so that a single entry point could be used to start the navigation logic (with or without sticky nav).
Documenting (with JSDoc?) of this file is a good idea, perhaps something for a separate issue?

sphinx_rtd_theme/static/js/theme.js
jessetan: This comes from #488, see comments there.

sphinx_rtd_theme/static/js/theme.js
jessetan: This also comes from #519. The theme code that calls this has been [changed accordingly](https://github.com/rtfd/sphinx_rtd_theme/pull/519/files#diff-1336f80431f476a232e3c48abb277c51).
It was tested on a local build, but I don't know if it has been tested on RTD. I can't imagine this function being called from outside of the theme though. A [search in the RTD source](https://github.com/rtfd/readthedocs.org/search?utf8=%E2%9C%93&q=StickyNav&type=) does not turn up anything.

jessetan:
This is a good example of confusion caused by the built JS being included in PRs, as discussed in #379

sphinx_rtd_theme/static/js/theme.js
agjohnson: Yeah, not important here I think, but having our public API documented would make sense for folks extending or repurposing the theme.

sphinx_rtd_theme/static/js/theme.js
agjohnson: I've opened #530 to experiment with removing this pattern.

sphinx_rtd_theme/static/js/theme.js
agjohnson: At some point we were overriding or referencing `StickyNav` in RTD. So this would be worth testing.

agjohnson:
Agreed. Separating asset generation from feature PRs is a great help. In the interim, until we get more serious about our release process, separating asset generation into a changeset and noting the changeset in PR would make review slightly easier. We do something similar with linting review on RTD, but it's still difficult to review.

Changes here make more sense now, think this can just be merged.

PR #531, opened 2017-12-28
Uglify during build

### LINKED ISSUE #527: Minimize theme.js file
It seems silly that we just browserify `theme.js` and just copy it over. We should minimize it like the rest of the static files.

I am not sure what the best way to do this is though. Perhaps use [grunt-contrib-uglify](https://www.npmjs.com/package/grunt-contrib-uglify) But not sure how to use that and browserify.

jessetan:
Adds an extra uglify task when running `grunt build`.
The symbol `jQuery` is left as-is so other code may use this identifier if needed.
IE6-8 compatibility is turned on just in case.

Fixes #527

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "grunt-contrib-connect": "^1.0.2",
     "grunt-contrib-copy": "~1.0.0",
     "grunt-contrib-sass": "~1.0.0",
+    "grunt-contrib-uglify": "^3.3.0",
     "grunt-contrib-watch": "~1.0.0",
     "grunt-exec": "~1.0.1",
     "grunt-open": "0.2.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "grunt-contrib-connect": "^1.0.2",
     "grunt-contrib-copy": "~1.0.0",
     "grunt-contrib-sass": "~1.0.0",
-    "grunt-contrib-uglify": "^3.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
     "grunt-contrib-connect": "^1.0.2",
     "grunt-contrib-copy": "~1.0.0",
     "grunt-contrib-sass": "~1.0.0",
+    "grunt-contrib-uglify": "^3.3.0",
```
Blendify: Maybe use a Tilda to match most other deps? Not really sure on this though. We should come up with a better policy for versions. @agjohnson @ericholscher How does RTD handle this?

jessetan: I think npm switched from using caret instead of tilde by default a while ago if you add a dependency using `--save-dev`.
Basically `^3.3.0` will upgrade to any `3.x.y` but not to `4.0.0`. `~3.3.0` will upgrade to any `3.3.y`, but not to `3.4.y`.
Since we're talking about semver, caret should be safe enough, but I'm fine with tilde if the extra safety is desired.

Blendify: I am fine with whatever, it just would be nice to keep consistency.

jessetan: I've changed it to tilde for consistency.

Blendify:
Unless someone has something else to add this can be merged

davidfischer:
This looks good to me.

PR #532, opened 2017-12-28
Do not make local-toc contents always visible

### LINKED ISSUE #445: Toc not expanding
When I create the sphinx doc the table of contents is not expandable. When I click on the + sign it only overlays a gray colour on all the items below but does not expand anything.

jessetan:
Fixes #445, but the historic reason why this SASS statement was added is not known.
Created this PR (with built CSS) so it is easier to test. 
@ericholscher hinted this may come from #375, which was fixed by @agjohnson


Blendify:
I did some git blaming and to me, it seems like this added with some changes to local tocs on single html and looks like they have since been deleted. So I think this is safe to revert.

agjohnson:
I think this is fine. It sounds like a local toc was tested, and I have no historical context to add here. +1

PR #535, opened 2017-12-29
Readme sections in sidebar

### LINKED ISSUE #523: README sections not in sidebar
Since #520, the README is part of the built documentation, but the headers in README do not appear in the sidebar at http://sphinx-rtd-theme.readthedocs.io/en/latest/index.html.

jessetan:
This PR makes the README sections appear in the navbar.
It uses an intermediate file to indirectly include the README (`toctree` does not allow including files that are in a parent directory).
The location of the README remains unchanged (so it renders nicely on GitHub).

A JavaScript redirect is used to send viewers directly to the README page, rather than displaying a page containing only the table of contents.

Fixes #523.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,8 +24,9 @@
 Read the Docs Sphinx Theme
 **************************
 
-.. contents:: 
+.. contents::
    :backlinks: none
+   :local:
 
 View a working demo_ over on readthedocs.org_.
 
--- a/docs/README.rst
+++ b/docs/README.rst
@@ -1,0 +1,1 @@
+.. include:: ../README.rst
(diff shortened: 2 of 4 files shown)


Blendify:
Personally, I do not think the redirect is needed.

jessetan:
@Blendify do you think the landing page needs some more content then? Without the redirect,  readers will land on a page with only toctree that they need to click through.

Blendify:
We can include the first couple of lines from the readme me? Or just
manually duplicate the content.

Blendify:
Poke, I would just include an introductory statement in the index this should be adequate to describe the project.

jessetan:
I'm waiting to see what the outcome of #532 is, since it may affect the way we can get the README to show up in the sidebar.

Blendify:
The is obsolete now with 67bb54d450a6cd0c7c8fb452c13af2a50bab9b21

PR #537, opened 2018-01-07
Remove base64 background image

Blendify:
I am not sure why this is here but removing it does not change how the rendered HTML looks.


davidfischer:
👍

PR #551, opened 2018-01-18
Allow setting 'rel' and 'title' attributes for stylesheets

Blendify:
See https://github.com/sphinx-doc/sphinx/commit/618ef6492c89c50f263c4c6c5794e40e35f91be8


jessetan:
@Blendify I haven't tested this, but why does the new line use `pathto(css.filename, 1)` while the old uses `pathto(css, 1)`?

Blendify:
I think the way the API changed internally.

PR #552, opened 2018-01-18
Remove leftover compass files

Blendify:
We got rid of compass in 8af1437a38ad0d869cffd2e7bd728020e69b552c these are now dead


davidfischer:
Can I have a bit of background on this one? I'm coming in late and don't think I know exactly why compass was here or why we don't need it anymore?

Blendify:
I came in after we use stopped too. Seems we used it to compile sass? Then
switched to grunt? No idea.

PR #555, opened 2018-01-24
Only output div.articleComments if the block has content

### LINKED ISSUE #553: articleComments: Should be in if statement
Currently, this is included in every documentation even though only a few percent actually use it. This should be put in an if statement. The issue with this is how do we not break compatibility? Can we check if the comment block is being overridden somehow? If not this would have to wait for some future version where we can say it is ok to drop compatibility.

jessetan:
Fixes #553

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -171,9 +171,11 @@
            <div itemprop="articleBody">
             {% block body %}{% endblock %}
            </div>
+           {% if self.comments()|trim %}
            <div class="articleComments">
             {% block comments %}{% endblock %}
            </div>
+           {% endif%}
           </div>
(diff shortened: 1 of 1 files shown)


sphinx_rtd_theme/layout.html
```diff
@@ -171,9 +171,11 @@
            <div itemprop="articleBody">
             {% block body %}{% endblock %}
            </div>
+           {% if self.comments()|trim %}
```
Blendify: Why trim?

jessetan: Without trimming, the `comment` block will be printed if it is not exactly empty but contains newlines or spaces. I don't know if that is likely to happen, but I thought it would be cheap enough to call `trim`

PR #559, opened 2018-01-27
add VERSIONING (rtfd/sphinx_rtd_theme#543)

### LINKED ISSUE #543: Conflicting interpretation of READTHEDOCS
In [`sphinx_rtd_theme/layout.html#L208-L210`](https://github.com/rtfd/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L208-L210), `_static/js/theme.js` is included `if not READTHEDOCS`. At the same time, [`sphinx_rtd_theme/versions.html#L1-L36`](https://github.com/rtfd/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/versions.html#L1-L36) is included `if READTHEDOCS`. As a result, the following settings in `html_context` produce a valid "versions box/menu" in the bottom of the sidebar, but it is not expandable due to the missing js script:

```
html_context = {
    'READTHEDOCS': True,
	'current_version': 'builders',
    'versions': [ ['builders', '../builders'], ['v0.35', '../v0.35'], ['v0.34', '../v0.34'] ],
	'downloads': [ ['PDF', '/builders.pdf'], ['HTML', '/builders.tgz'] ]
}
```

In order to fix it, each artifact needs to be edited so that `<script type="text/javascript" src="/js/theme.js"></script>` (or the corresponding path) is added.

I'd suggest using a different envvar/mechanism in order to know whether the project is to be hosted on readthedocs.org, so that the "versions box/menu" is usable no matter what the host is.

1138-4EB:
close #543


jessetan:
Who is responsible for setting the `VERSIONING` variable? Is this documented?

1138-4EB:
In ReadTheDocs.org, no one, because it is not required. For builds outside of ReadTheDocs, the same agent that would decide whether to set `READTHEDOCS`.

It is not documented anywhere. However, the usage is explained in #543. Furthermore I didn't find `READTHEDOCS` to be documented. Is it? If so, I'll add a comment in the same place.

1138-4EB:
Both `READTHEDOCS` and `versioning` are now added to `theme.conf` and `README.rst`.

sphinx_rtd_theme/layout.html
```diff
@@ -110,7 +109,7 @@
 
           {% if theme_display_version %}
             {%- set nav_version = version %}
-            {% if READTHEDOCS and current_version %}
+            {% if (READTHEDOCS or versioning) and current_version %}
```
ericholscher: This is not going to work. All variables set in the theme are prepended with `theme_`.

README.rst
```diff
@@ -94,6 +94,8 @@ file of this repository, and can be defined in your project's ``conf.py`` via
         'display_version': True,
         'prev_next_buttons_location': bottom,
         'style_external_links': False,
+        'versioning' : False,
+        'READTHEDOCS' : False,
```
ericholscher: This isn't the same thing as RTD is doing. This will set a theme variable, and not the actual HTML context.

1138-4EB: Yet, `versioning` will require almost the same context data as `READTHEDOCS`: `current_version` (string), `versions` (dictionary), and `downloads` (dictionary). It seems not coherent to handle `READTHEDOCS` as a context field, along with auxiliary fields; but make `versioning` be a theme option, provided that the same context fields are required.

sphinx_rtd_theme/layout.html
1138-4EB: Fixed and tested now.

jessetan:
`versioning` (or `VERSIONING`) is different from any other type of option that we've had (except for `READTHEDOCS`) because it is not set by an author but by the build environment.

I propose to not implement this as a theme option, but the way it was in the original PR: as `VERSIONING` (or some similar name), treated the same way as the `READTHEDOCS` build environment setting, and adding a new section in the documentation about how the theme interacts with build environments. It does not need to go into too much detail, since it is not intended for authors, just mentioning that these vars may be set by the build environment and that they will change some functionality of the theme. In future, perhaps ReadTheDocs can even set `VERSIONING` if it has the same effect for version information.

There do seem to be different opinions here so I think @1138-4EB might want to wait with making changes until the reviewers can agree on a direction.

Blendify:
I have just rethought this patch and would really like to see if we can get it merged into `0.2.5`  mainly I really need this functionality for another project. I will be updating this over the next week hopefully.

Blendify:
Re theme option or build context

I think this should be a theme option. The reason for this is that build options should be reserved for things that the user wants to change and the html context are things that should change depending on when/where the docs were built from.

Blendify:
Re-opened as #574

PR #562, opened 2018-01-30
Create CODE_OF_CONDUCT.md

Blendify:
This is more of a general PR that I think we should add to the rest of RTD.
It gives us a good set of standards for how to act and responsibilities for issues we might deal with.

Note that I used the email `contact@readthedocs.org` maybe there is a better email to use here? Or create a special email. (I don't even know if that email is real).

### DIFF
--- a/.github/CODE_OF_CONDUCT.md
+++ b/.github/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, gender identity and expression, level of experience, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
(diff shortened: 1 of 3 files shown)


agjohnson:
We already have a code of conduct that should cover all Read the Docs related projects:
http://docs.readthedocs.io/en/latest/code-of-conduct.html

So, I don't think there is much reason to duplicate here or to have a different code of conduct. Editing `CONTRIBUTING.rst` to reference our general code of conduct makes sense though.

Blendify:
I am still going to create the guideline file because GitHub references this file.

davidfischer:
In order for the code of conduct to be automatically be picked up by github (see screenshot below), I believe it needs to be at the root of the repository and not in the `.github` directory. Here are the [relevant docs](https://help.github.com/articles/about-community-profiles-for-public-repositories/) as far as I'm aware.

screen shot 2018-01-30 at 9 13 43 am

Blendify:
No this works see
https://help.github.com/articles/setting-guidelines-for-repository-contributors/

humitos:
Re-Approve of the new changes :)

PR #566, opened 2018-01-31
Grunt: refactor

Blendify:
- Cleanup font copying
- Create new tasks: install, docs


davidfischer:
The `Gruntfile.js` currently has parse errors so I'm marking this "work in progress" for now.

Blendify:
Ah should work, I was rushed when I wrote this one.

Gruntfile.js
```diff
@@ -181,5 +169,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-browserify');
 
   grunt.registerTask('default', ['exec:bower_update','clean','copy:fonts','sass:dev','browserify:dev','exec:build_sphinx','connect','open','watch']);
-  grunt.registerTask('build', ['exec:bower_update','clean','copy:fonts','sass:build','browserify:build','uglify','exec:build_sphinx']);
+  grunt.registerTask('build', ['exec:bower_update','clean','copy:fonts','sass:build','browserify:build','uglify']);
```
davidfischer: This removed the actual building of the docs. Was that intentional?

Gruntfile.js
```diff
@@ -25,33 +25,15 @@ module.exports = function(grunt) {
...
-          {
-              expand: true,
-              flatten: true,
-              src: ['bower_components/inconsolata-googlefont/Inconsolata-Bold.ttf',
-                    'bower_components/inconsolata-googlefont/Inconsolata-Regular.ttf'],
+                    'bower_components/lato-googlefont/Lato-BoldItalic.ttf',
+                    'bower_components/robotoslab-googlefont/RobotoSlab-Bold.ttf',
+                    'bower_components/robotoslab-googlefont/RobotoSlab-Regular.ttf',
+                    'bower_components/inconsolata-googlefont/Inconsolata-Bold.ttf',
+                    'bower_components/inconsolata-googlefont/Inconsolata-Regular.ttf']
```
davidfischer: There's a syntax error here. A trailing comma is required.

Gruntfile.js
Blendify: Yes, the way I see this is that `grunt build` should just build the theme related files.

davidfischer: I'm good with it. I just want to be sure.

davidfischer:
:ship:

Blendify:
@ericholscher do you agree with the changes done to the grunt tasks?

Gruntfile.js
```diff
@@ -181,5 +169,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-browserify');
 
   grunt.registerTask('default', ['exec:bower_update','clean','copy:fonts','sass:dev','browserify:dev','exec:build_sphinx','connect','open','watch']);
-  grunt.registerTask('build', ['exec:bower_update','clean','copy:fonts','sass:build','browserify:build','uglify','exec:build_sphinx']);
+  grunt.registerTask('build',   ['clean','copy:fonts','sass:build','browserify:build','uglify']);
+  grunt.registerTask('install', ['exec:bower_update','clean','copy:fonts','sass:build','browserify:build','uglify']);
```
jessetan: What's the reason for the new `install` step? I think we should have a single command that produces a built theme with CSS and JS, completely from scratch. It looks like you're doing that with `install`, but what would we need `build` for then?

Gruntfile.js
```diff
@@ -181,5 +169,7 @@ module.exports = function(grunt) {
...
+  grunt.registerTask('docs',    ['exec:build_sphinx','connect','open']);
```
jessetan: This should also call all the stuff that copies and builds assets (`copy`, `sass`, `browserify`). The docs need these assets for proper display and they might not be there if the were `clean`ed.

Blendify:
I decided to revert the changes that we were not sure about. Im not exactly sure what I was going for but I like how it is now.

jessetan:
Do you want to keep the `docs` task as it is now? It will not have styling if run after a `clean` since it does not build the assets.

PR #568, opened 2018-02-03
Run Grunt

Blendify:


agjohnson:
Perhaps there is some way to provide some information on what is changing here, such as combining grunt build and release note update PRs? This doesn't provide me with context as to what exactly is changing in these files.

I'm sure this PR is fine though 👍

jessetan:
Looks like this is just a PR to update the built JS and CSS files. Maybe hold off on this one until 0.2.5 will be released, I get the feeling we're quite close to that (but no pressure to whichever human needs to go through the actual release process)

Blendify:
@agjohnson If we review the PRs then there should not be anything new to review in these PR's We defiantly need to come up with a better plan for the future. @jessetan I will probably run once more before release, however, this makes testing changes easier.

PR #569, opened 2018-02-03
Docs: fix broken link

Blendify:


docs/contributing.rst
```diff
@@ -12,8 +12,8 @@ of a front-end developer (vs. that of a python developer). That means installing
 
 .. seealso::
 
-   If you are unsure of approptiate actions to take while interacting with our
-   communinty please read our :doc:`Code of Conduct <rtd:/code-of-conduct>`.
+   If you are unsure of appropriate actions to take while interacting with our
+   community please read our `Code of Conduct <http://docs.readthedocs.io/en/latest/code-of-conduct.html>`__.
```
agjohnson: I prefer the intersphinx mapping. We should be dogfooding sphinx/rtd features as much as possible :)

agjohnson: For instance, I believe the intersphinx link you're looking for here is `<rtd:code-of-conduct>`, which also isn't a `:doc:` reference.

Blendify: I thought it had to be an absolute path selector but I guess this is assumed for intersphinx links?

jessetan: Intersphinx uses a role that takes a label. The actual (absolute) url is found by looking up this label in the `objects.inv` file retrieved by Intersphinx

PR #572, opened 2018-02-12
Replace the name 'Dave Snider' with RTD

Blendify:


ericholscher:
I don't think we have a standard way of doing it, but probably should. I think "Read the Docs, Inc & contributors" is probably the most legally correct.

ericholscher:
I do feel bad removing Dave's name from the explicit listing, since it was his creation originally. Is there a problem we're trying to solve with this?

Blendify:
Not specifically just that we are not giving hardly any attribution to current authors. I would be fine with add Dave to "Read the Docs, Inc & contributors"

davidfischer:
Are we agreed on "Dave Snider, Read the Docs, Inc. & contributors"?

ericholscher:
"Dave Snider, Read the Docs, Inc. & contributors" sgtm.

PR #574, opened 2018-02-13
Buildthedocs add versioning

### LINKED ISSUE #543: Conflicting interpretation of READTHEDOCS

Blendify:
Adds the ability to use the versioning system even when not on RTD.

Original patch by @1138-4EB 

Closes #543


docs/configuring.rst
```diff
@@ -49,6 +50,8 @@ Base options
 * ``vcs_pageview_mode`` String. Changes how to view files when using `display_github`, `display_gitlab`, etc.
   When using Github or Gitlab this can be: `blob` (default), `edit`, or `raw`,
   on Bitbucket, this can be either: `view` (default) or `edit`.
+  * ``versioning`` Bool. Add 'versions' menu to the bottom of the sidebar. Defaults to ``False``.
```
jessetan: Could you add some documentation (either here or in a a separate section) which information is expected to be made available by the build environment? Perhaps link to some RTD code on GitHub that sets this.
This is an odd setting where just enabling it will not make it work, it depends on other information being available.

1138-4EB: Please, note that this is also indirectly related to `display_github`, because `github_version` should match `current_version`, as long as the version is named after the corresponding branch. Otherwise, the edit/view button will always point to the master branch, no matter which version is actually being browsed.

agjohnson:
So, I'm -1 on these changes from a product perspective. I don't want to further generalize our theme and make maintaining this more work -- it's already a lot of work. If buildthedocs needs a special case to the theme, this project should extend it's own theme off of our theme. Adding more template blocks is something we can manage, but being responsible for the use case of an entirely different project is not on our roadmap.

For instance, we want to just remove this block and not be responsible for maintaining a list of versions in our theme -- this is the correct thing to do from Read the Docs perspective. Having a clash here is just going to be technical debt to us.

So sorry to close work here, but we discussed this use case and don't see the need for it. If we can add more template block extension points in our theme, that is probably fine. We just can't be responsible for downstream breakage.

F

3nids:
I came upon this issue and would just share that I would have love to benefit from this feature (I was looking at this feature for QGIS Python API documentation to handle the different API versions).

3nids:
if this interests anyone, this can be achieved by templating.
see https://stackoverflow.com/questions/49331914/enable-versions-in-sidebar-in-sphinx-read-the-docs-theme/63054018/#63054018

PR #586, opened 2018-03-03
use id to find the future current sidebar item

### LINKED ISSUE #583: :ref: target not highlighted in navigation bar
here is an issue that should've been reported here, as discussed in:
https://github.com/rtfd/readthedocs.org/issues/3679

**my original description:**

here is an example of a link to a labeled section:
https://docs.readthedocs.io/en/latest/getting_started.html#in-rst

the navigation tree is not highlight correctly,
but the following link highlights 'in-markdown' section as expected:
https://docs.readthedocs.io/en/latest/getting_started.html#in-markdown

these two urls are the results of navigation, by clicking the two links in 'Write Your Docs' section
https://docs.readthedocs.io/en/latest/getting_started.html#write-your-docs

notice that 'in-markdown' label and title are approximately the same, and in-rst and 'in restructuredtext' aren't very close.

maybe a sphinx-doc issue but rtd is the only theme I know that highlights navigation tree, so the mistake might be more obvious here.

zprd:
Fixes #583

- find the referenced 'id element' first, then the closest section
- if no association found in the sidebar, fallback to #
- reword comments

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -117,19 +117,19 @@ function ThemeNav () {
         var anchor = encodeURI(window.location.hash) || '#';
 
         try {
-            var link = $('.wy-menu-vertical')
-                .find('[href="' + anchor + '"]');
-            // If we didn't find a link, it may be because we clicked on
-            // something that is not in the sidebar (eg: when using
-            // sphinxcontrib.httpdomain it generates headerlinks but those
-            // aren't picked up and placed in the toctree). So let's find
-            // the closest header in the document and try with that one.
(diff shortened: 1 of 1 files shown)


zprd:
The problem here is that we used href to find the target section. This gives false results when used with :ref: for example, we found the the calling section instead of the targeted one.

The intent here is to rely on the element (span, dt ...) containing the id we are referring to, which should be positioned inside the target section.

zprd:
ping? 👍

Issue #354, opened 2017-01-18
Lists with class open do not work


mitya57:
reStructuredText has a distinction between “compact” and “open” lists. When a list has an `open` class, all its items are wrapped in `<p>` tags so that they have more spacing. For example, this block of reStructuredText:
```rest
.. rst-class:: open

* Item one
* Item two
* Item three
```

Produces the following output:
```html
<ul class="open">
<li><p class="first">Item one</p>
</li>
<li><p class="first">Item two</p>
</li>
<li><p class="first">Item three</p>
</li>
</ul>
```

Compared to compact lists:
```html
<ul class="compact simple">
<li>Item one</li>
<li>Item two</li>
<li>Item three</li>
</ul>
```

However with the RTD theme all lists render as compact, because the CSS sets `margin-bottom: 0` for `p:last-child`:

![Screenshot with the RTD theme](https://cloud.githubusercontent.com/assets/1381584/22066537/c4c2d072-dda6-11e6-9a08-fc40c5a8aeb7.png)

For comparison, this is how it looks with the default theme (Alabaster):

![Screenshot with Alabaster theme](https://cloud.githubusercontent.com/assets/1381584/22066566/e77f677e-dda6-11e6-8819-9fcaccc0e22e.png)

Support for “open” lists would be nice to have as compact lists look ugly when the list items contain multiple paragraphs or images.

Blendify:
I am confused, is there documentation that this is supposed to work?
I tried looking in the Sphinx docs, restructured text docs, alabaster, and the w3 school docs and don't see any. Can you provide information about this?

mitya57:
It is briefly mentioned in Docutils documentation: [one](http://docutils.sourceforge.net/docs/user/config.html#compact-lists) and [two](http://docutils.sourceforge.net/docs/user/latex.html#lists). Note that in Sphinx the `class` directive [becomes `rst-class`](http://www.sphinx-doc.org/en/stable/rest.html#id3).

Our problem was that we often use multiple paragraphs in list items, and the spacing between items is less than the spacing between paragraphs in one item, which makes it difficult to visually distinguish the list items. Here is a screenshot:

> ![rtd-theme-lists](https://user-images.githubusercontent.com/1381584/34174146-08207422-e509-11e7-96e0-d96c4078c5f5.png)

We would like the spacing between lists be at least same or even larger. Using the `open` class would potentially fix this, but unfortunately that does not help.

Blendify:
[...]

PR #591, opened 2018-03-16
Add spacing between items in "open" lists

### LINKED ISSUE #590: Missing spacing between items in "open" lists
### Problem

Docutils [distinguishes between "open" and "compact" lists](http://docutils.sourceforge.net/sandbox/html4strict/data/simple-lists.html). Open lists are meant to have some vertical space between the list items. Docutils automatically generates an open list if any item has multiple paragraphs.[1]

The RTD theme suppresses[2] the vertical spacing between open list items. This looks wrong, as it visually groups the last paragraph in an item with the first paragraph in the next item.

### Reproducible Project

Here's an example:

```rst
With a "compact" (simple) list:

* Each item is a single paragraph
* So docutils can optimize the list
* To use less vertical space

The alternative is an "open" list:

* At least one item has multiple paragraphs.

* Docutils *doesn't* collapse vertical
  space in open lists. This preserves
  visual grouping of related paragraphs.

  It makes the text easier to read.

* The behavior also applies to other blocks
  in list items, like code examples.
```

You can also see the behavior in the wild, e.g. the numbered lists on [this page](https://anymail.readthedocs.io/en/v2.0/installation/#configuring-tracking-and-inbound-webhooks).

#### Error Logs/Results

The example above renders as:

rtd-compact-and-open-lists-before

Notice how the last paragraph of the second bullet seems to be visually grouped with the bullet below it.

#### Expected Results

There should be vertical space between items in an open list:

rtd-open-lists-after

### Environment Info

- Python Version: 3.6.4 (or 2.7.14)
- Sphinx Version: 1.7.1
- RTD Theme Version: 0.2.4

### Notes

[1]: Technically, docutils html4 writer actually generates all lists as open, but then [by default](http://docutils.sourceforge.net/docs/user/config.html#compact-lists) compacts the simple ones. [Sphinx follows](http://www.sphinx-doc.org/en/1.7/config.html#confval-html_compact_lists) the docutils default. You can override this on individual lists by adding the rst class "compact" or "open" (see #354).

[2]: Actually, the RTD theme tries to [*add* margin](https://github.com/rtfd/sphinx_rtd_theme/blob/0.2.5b2/sass/_theme_rst.sass#L85-L86) between `<p>` within `<li>`, but wyrm's .wy-plain-list-disc and .wy-plain-list-decimal [set that back to zero](https://github.com/snide/wyrm/blob/1.0.9/sass/wyrm_core/_type.sass#L207-L208) on `... li p:last-child`. (With docutils html4 writer defaults, any `<p>` in an `<li>` means you're dealing with an open list; the p's are stripped for compact lists.)

medmunds:
Override wyrm's `... li p:last-child { margin-bottom: 0 }` from .wy-plain-list-disc and .wy-plain-list-decimal.

Fixes #590. (And probably #354.)


Blendify:
Thanks for the PR. Will have a look ASAP.

PR #600, opened 2018-04-02
Set collapse_navigation to true

Blendify:
This matches 0.2.4 behavior. Although it might not be desired behavior, the behavior of setting this to true can result in file size to explode and build times also.

This was caused by bf9faa98f0cb64aa3536528231ad9a3f48e819ac which effectly switched the logic of `collapse_navigation`

### DIFF
--- a/sphinx_rtd_theme/theme.conf
+++ b/sphinx_rtd_theme/theme.conf
@@ -6,7 +6,7 @@ pygments_style = default
 [options]
 canonical_url =
 analytics_id =
-collapse_navigation = False
+collapse_navigation = True
 sticky_navigation = True
 navigation_depth = 4
 includehidden = True


agjohnson:
A little more context into the problem here, for anyone coming across this in the future:

When the collapse navigation option was moved into `theme.conf`, it got a default of `False`, however this is treated as a string in templates, so the default was still effectively `True`. Changes going into the `0.3.0` release default this to a boolean, so the default behavior once again changes, to a correctly reflected default of `False`. This PR reverts back to the `0.2.4` default of `True`, as this option balloons resources for builds and makes doc pages slow on large docsets.

We can probably re-enabled no collapsing with either:

* Inspection of doc size and automatically set this option
* Moving navigation to external JSON/HTML blob that is included, thus making the menu more dynamic and removing the need for duplicated menu on every html page

PR #606, opened 2018-04-10
Write theme version at the top of theme.js and thems.css

jessetan:
This makes it easier to determine the exact version of the theme a site is using. Useful for debugging purposes


davidfischer:
Alabaster actually writes it to the HTML. I don't think we need to advertise the theme version but putting it in a comment may not be a bad idea.

jessetan:
Added to CSS as well.

PR #607, opened 2018-04-10
Write minified file to correct location

jessetan:
The minification from #531 only produced a minified js file for the demo doc build, not when building for distribution. This PR also minifies for the latter situation.


davidfischer:
I don't think I understand why this one line makes a difference. Could you explain it?

jessetan:
Looks like a made a mistake, I can no longer reproduce the issue I saw. Closing.

PR #609, opened 2018-04-11
Docs: correct navigation_depth discription

### LINKED ISSUE #539: sidebar table of contents limited nesting?
Hi -- in ```sphinx_rtd_theme/theme.conf``` you have ```navigation_depth = 4``` but in the readme it says
"navigation_depth Int. Indicate the max depth of the tree; by default, all levels are included." the example given in the readme is actually the default setting as well This is a bit confusing and had me trying to reconfigure the sidebar template before I realised what the problem was, it would be good to make it consistent one way of the other.  Otherwise impressed by this theme!

Blendify:
Fixes #539


Blendify:
I think that is a good idea too.

PR #613, opened 2018-04-17
Add a requirements.txt file

davidfischer:
This simply puts the requirements for building the theme into a `requirements.txt` file. This makes it more obvious what is required.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -23,7 +23,7 @@ Set up your environment
 
    .. code:: bash
 
-       pip install sphinx sphinxcontrib-httpdomain
+       pip install -r requirements.txt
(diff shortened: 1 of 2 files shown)


davidfischer:
Actually I realize this isn't necessary. There's already a `requirements.txt` file under `docs`. However, running `grunt` also requires those requirements.

agjohnson:
Also, for packages, i prefer avoiding `requirements.txt` file. `setup.py/setup.cfg` is a better place for these dependencies

davidfischer:
@agjohnson normally I agree. The issue is that building for grunt requires installing packages that won't be installed if you just create a fresh venv.

Issue #610, opened 2018-04-14
Version 0.3.0 breaks sidebar


goerz:
See https://github.com/rtfd/readthedocs.org/issues/3925

jessetan:
See my comment in https://github.com/rtfd/readthedocs.org/issues/3925#issuecomment-381408874, I think this is not caused by the theme but by the way it is packaged on RTD.
Leaving this issue open until we have confirmation on this though.

Blendify:
I can confirm on http://shenfun.readthedocs.io/en/latest/

@jessetan I think this might be from https://github.com/rtfd/sphinx_rtd_theme/commit/5db94ebf5f6f06d36fb961a7cee0b53d426214b6?

jessetan:
@Blendify yes. The code in layout.html should match the code in theme.js. That is not the case in rtfd/readthedocs.org#3925.

PR #614, opened 2018-04-18
Always bundle the theme CSS/JS

davidfischer:
This removes the `{% if READTHEDOCS %}` conditionals around the theme JS/CSS. The goal here is to uncouple the code injected by readthedocs.org into this theme from the version of the CSS/JS on the theme itself.

This change will have a required change in the readthedocs.org repo which I will link when that PR is created. There is also an optional (but highly recommended) change coming in the [readthedocs-sphinx-ext](https://github.com/rtfd/readthedocs-sphinx-ext) repo which I will also link to.


davidfischer:
Here's the corresponding .org PR: https://github.com/rtfd/readthedocs.org/pull/3968

Without merging that PR, the theme JS will be included twice. This doesn't seem to break anything.

There's also a PR on the readthedocs-sphinx-ext repo: https://github.com/rtfd/readthedocs-sphinx-ext/pull/39

Without that PR, the theme CSS will be included twice. This doesn't break anything. One thing to note is that the CSS from media.readthedocs.org is included later in the DOM so it will take precedence over the theme bundled CSS. We should just merge the above PR to fix that.

davidfischer:
Coupled with the other PRs, this fixes #610.

davidfischer:
Will do. I will also update the changelog to mention that 0.3.0 has some issues.

ericholscher:
Maybe we can remove the other 2 `{% if READTHEDOCS %}` sections, and have that be 1.0 :)

davidfischer:
> Maybe we can remove the other 2 {% if READTHEDOCS %} sections, and have that be 1.0 :)

Let's make that a goal for 1.0. However, I think targeting this for 0.4.0 would be good.

PR #615, opened 2018-04-20
Extract hash ourselves to work around difference between Fx and other browsers

### LINKED ISSUE #495: Navigation tree collapse if page link's anchor is unicode text.
Sphinx with recommonmark does not use unicode text header as page link's anchor, just ignore them. If the header is written by unicode text only, anchor is empty like "http://example.com/foo.html#". So I use [unicode_ids extension](https://pypi.python.org/pypi/unicode_ids) to generate unicode anchor text from header.

In that case, navigation to 3rd or below level get fall back to top level content and toc tree collapsed, because sphinx_rtd_theme try to find URI escaped anchor and fail.

Change theme.js from
`var anchor = encodeURI(window.location.hash);`
to
`var anchor = window.location.hash;`
worked for me. Is this acceptable change?

jessetan:
Should fix #495.

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -114,16 +114,16 @@ function ThemeNav () {
 
     nav.reset = function () {
         // Get anchor from URL and open up nested nav
-        var anchor = encodeURI(window.location.hash) || '#';
+        var anchor = encodeURI(window.location.href.split('#')[1]) || '';
 
         try {
             var vmenu = $('.wy-menu-vertical');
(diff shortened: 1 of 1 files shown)


jessetan:
@davidfischer I can also no longer reproduce Firefox' old, buggy behavior.
I've asked the original reporter in #495 if this is still an issue.

# DO NOT MERGE YET

jessetan:
This PR did not solve the original issue, closing.

PR #616, opened 2018-04-20
Add padding to parsed literal blocks

jessetan:
Before:

screen shot 2018-04-20 at 16 52 16

After:

screen shot 2018-04-20 at 16 51 06


sass/_theme_rst.sass
```diff
@@ -53,13 +53,20 @@
     line-height: $base-line-height
     margin-bottom: $base-line-height
 
+  pre.literal-block
+    white-space: pre
+    margin: 0
+    padding: $base-line-height / 2 $base-line-height / 2
+    font-family: $code-font-family
+    display: block
```
davidfischer: A `<pre>` is block level element already.

PR #623, opened 2018-04-30
Add backwards compatiblity for pre 0.3.0 layout.html files

### LINKED ISSUE #610: Version 0.3.0 breaks sidebar

agjohnson:
Until we have broken up the theme javascript/css and the files served by Read
the Docs, backwards compatiblility is required in all changes to javascript.
This reverts some of the changes in 0.3.0 around our javascript.

Refs https://github.com/rtfd/readthedocs.org/issues/3925
Fixes #610
Refs #614 (this will properly fix the issue)


sphinx_rtd_theme/layout.html
```diff
@@ -215,11 +215,7 @@
 
   <script type="text/javascript">
       jQuery(function () {
-          {% if theme_sticky_navigation|tobool %}
-          SphinxRtdTheme.Navigation.enableSticky();
-          {% else %}
-          SphinxRtdTheme.Navigation.enable();
-          {% endif %}
+          SphinxRtdTheme.Navigation.enable({{ theme_sticky_navigation|tobool }});
```
davidfischer: This isn't quite right. I guess you could pass through `|lower` or something...

screen shot 2018-04-30 at 1 38 31 pm

agjohnson: Ah yes, good catch, updated and rebased with 68040a1.

I guess this worked because I didn't test a project overriding `sticky_navigation` in their conf.py. That is, i got the default `true` or `undefined`.

agjohnson:
Yup, holding off on that until this release is tagged. I can start that process next

agjohnson:
Oh, and for some more context here: we are delaying #614 and friends to do an audit of this usage and to ensure we aren't making additional backwards incompatible changes with those PRs. This fix is just a temporary solution that will be deployed likely today or tomorrow.

I'm going to tag a release now, but it won't be fully released to pypi until we release RTD .org/.com

Issue #525, opened 2017-12-22
Decide on use of Modernizr


jessetan:
In #518 the use of Modernizr for the theme was questioned. If we can remove this, it would make the build a bit simpler and the output a bit smaller.

- If the feature detection is not used by the theme, is it still needed to include Modernizr?
- If we do need Modernizr, do we need the kitchen sink build currently included?

remyabel:
According to https://github.com/rtfd/sphinx_rtd_theme/pull/248, Modernizr was only included for old IE support. I've personally removed it from my project and haven't noticed any issues.

Blendify:
I say remove it. But would let @agjohnson again have the final word.

davidfischer:
Just to chime in with a bit of data, IE versions before 11.0 are less than 0.2% of traffic (Nov 1 through today) on readthedocs.io according to GA.

This might be a good time to formalize our browser support policy.

Blendify:
Considering this only affects versions below ie9. It will affect a very low
number but I think yes we should come to a conclusion and document what
version we are willing to support.

jessetan:
@davidfischer Is the 0.2% traffic for the documentation RTD itself or for all the documentation hosted by RTD?

We can keep IE<9 support for the HTML5 elems by just using the htmlshiv in the head and removing the rest of Modernizr. We do not use any of the feature detection stuff in Modernizr.

A formal browser support policy would help. Old IE support is already iffy because the theme uses jQuery which does not work on IE<9. I don't know how much more broken the theme would be if we would also remove the htmlshiv, but we can try to see what happens.

davidfischer:
I missed this but the 0.2% is for all docs sites hosted by Read the Docs. Essentially nobody visiting RTD is using IE<11 now. IE11 is still close to 2%.

Blendify:
I think it is safe to remove this now.

davidfischer:
I we are just including it for IE<11, then it is safe to remove. Do we know precisely which detects we use if any?

davidfischer:
Consider me +1 then. I'm fine dropping support for IE<11 as virtually no users use those old versions.

I do believe that dropping support should involve incrementing the version to 0.4.0 though. We also should be explicit in the docs about which browsers are supported.

jessetan:
> We also should be explicit in the docs about which browsers are supported.

[...]

PR #624, opened 2018-05-01
Remove Modernizr

jessetan:
# DO NOT MERGE YET

This PR removes Modernizr, as discussed in #525.
Behavior of non-evergreen browsers (mainly IE6 - IE11) needs to be verified, as this also removes the html5shiv that is incorporated in Modernizr. Possible effects may be incorrect rendering, buttons not working, navigation not functioning, JavaScript errors in console

This PR does not remove the rAF polyfill for old IE.

- [x] Chrome (latest) verified
- [x] Firefox (latest) verified
- [x] Edge (latest) verified
- [x] IE11 verified
- [ ] IE10 verified
- [ ] IE9 verified
- [ ] IE8 verified
- [ ] IE7 verified
- [ ] IE6 verified


mitya57:
Microsoft has a great service for generating website screenshots with different browsers, including all IE versions since IE 8:

https://developer.microsoft.com/en-us/microsoft-edge/tools/screenshots/

If you deploy your changes somewhere, it will be quite easy to see which browsers get broken.

jessetan:
I know of that page (I even referenced it in the comment above) but it seems to produce some different results from actual tests on browsers. Since I don't know exactly how they are creating the screenshots, I prefer manual testing for such an important change.
Screenshots also don't show JS errors, scrolling behavior, click behavior etc.

mitya57:
Oh, how could I not notice that link :)

jessetan:
@rupl thanks for commenting. I'm not sure how applicable your comment is, since sphinx_rtd_theme is not a Drupal theme, but a Sphinx theme.

rupl:
looooool sorry 😳

Blendify:
I think we should go forward with this.

jessetan:
Closing in favor of #756

PR #626, opened 2018-05-01
Remove overflow

davidfischer:
Fixes this issue:

39494888-f3296432-4d4c-11e8-8d75-34e055a2845b


jessetan:
I see no harm in having this scrollable in case of many versions (#575), but the extra white space is ugly. I'll see if I can revisit this sometime in the future.

PR #638, opened 2018-05-14
Accessibility improvements to the left nav

davidfischer:
This makes a few color changes to the left navigation with an eye toward improved accessibility. Specifically, I looked at the colors on our existing theme and they [don't have enough contrast](http://jxnblk.com/colorable/demos/text/?background=%23343131&foreground=%236f6f6f). The headings are the worst offenders but some of the links aren't fantastic.

screen shot 2018-05-14 at 3 28 40 pm

Ping @mahmoud

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,14 @@
 Changelog
 *********
 
+Master
+======
+
+Fixes
+-----
+
+* Color accessibility improvements on the left navigation
+
 v0.3.1
 ======
 
--- a/sass/_theme_layout.sass
+++ b/sass/_theme_layout.sass
@@ -24,7 +24,9 @@
     padding: 0 $base-font-size
 
 .wy-menu-vertical
+  color: $menu-medium
   width: $nav-desktop-width
+
   header, p.caption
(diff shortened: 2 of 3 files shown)


davidfischer:
By itself though, going to jet black (#000) will not solve the problem. The accessibility rating would only be [acceptable for large headings](http://jxnblk.com/colorable/demos/text/?background=%23000&foreground=%236f6f6f).

The other problem is that the left navigation would blend in with the version menu at the bottom ("Read the Docs v:latest")

I suspect we can slightly darken the captions ("Demo documents") though while still keeping a good rating. I'll see what I can do.

mahmoud:
Excellent! Is there a milestone / timeline for getting this released? I assume I'll just have to re-pin to the new release in my requirements.txt.

Thanks!

davidfischer:
If you've pinned to a precise version, yes, you'll need to re-pin.

If you haven't, or you pinned to something like `sphinx-rtd-theme<0.4` (which will be the default after https://github.com/rtfd/readthedocs.org/pull/4096), then it should just happen automatically. I aim to get both of these changes out next week.

Issue #524, opened 2017-12-22
Decide on code font


jessetan:
Fo non-code text, the theme provides the actual font file. It also include Inconsolata, which could be used for code, but isn't.
The current font stack for code is `Consolas, "Andale Mono WT", "Andale Mono", "Lucida Console", "Lucida Sans Typewriter", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", "Liberation Mono", "Nimbus Mono L", Monaco, "Courier New", Courier, monospace`.
This causes code to look differently on various platforms. It can also look less than ideal or not native on some platforms ([Andale has a comma that looks like a period](https://github.com/rtfd/sphinx_rtd_theme/issues/121#issuecomment-353134143), iOS does not have any of the specified fonts except Courier New, Monaco may not have a bold variant #460)

It would be nice to standardise on one single code font (whether Inconsolata or something else) or at to choose a font stack that uses system fonts for each OS to have a more native look. In the latter case, Inconsolata can be removed from the repo.

davidfischer:
I actually wouldn't mind it if we went to a native font stack as opposed to one where we have to download a web font. I would actually propose we do this for both monospace and regular fonts. Currently fonts require about 1.5MB of downloads to load up the docs.

At the same time, I think we do need to be sensitive about what works especially since we do display monospaced fonts sometimes in bold because they are passed through Pygments.

Bootstrap [switched to all native font stacks](https://getbootstrap.com/docs/4.1/content/reboot/#native-font-stack) for 4.0. Their monospaced native stack is `SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace`.

GitHub also [uses a native font stack](http://markdotto.com/2018/02/07/github-system-fonts/). Their monospaced one is `"SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier, monospace;`.

While a couple popular sites doing it isn't enough to make me want to do it, I do think that coupled with the performance gains are pretty good reasons.

* SFMono-Regular - This is pretty much iOS only.
* Menlo - This is an Apple font available on MacOS.
* Consolas - This is a Windows font.
* Liberation Mono - This is a free font comparable to Courier. It is designed for Red Hat and I believe it is included in a default Ubuntu install as well as some other distros.
* Courier - This is an old font that is widely available with a somewhat permissive license.
* Courier New - This is a widely available proprietary version of the above font.

[...]

PR #639, opened 2018-05-15
Change Lato font to webfont (version 2.015)

jessetan:
This PR:

- Adds additional formats (woff2, woff, eot) to the existing ttf format
- The woff(2) files are about 25%-50% of the size of the ttf files, which decreases time to download
- Supports modern browsers through woff(2), IE < 9 through eot, older Safari/Android through ttf. Unused font files are not downloaded by the browser.
- Updates Lato to the latest official release from latofonts.com
- Does not check for local installs of Lato. This can be added at the slight risk of making the font look different. Users with the font installed will not need to download the webfont.
- Does not use character subsetting. Much smaller files are possible if only the Latin set is required
- Makes the font look sliiiiighlty more squished vertically at the default size of 16px. This is true for any version of Lato I have tested, except for the ttf that the theme previously used. This is also present on readthedocs.org after rtfd/readthedocs.org#3982

Related: rtfd/readthedocs.org#3982, rtfd/readthedocs.org#3564, #524 and #642


sass/_theme_font_local.sass
```diff
@@ -12,27 +12,32 @@
...
+  src: url('../fonts/Lato/lato-bold.eot?#iefix') format('embedded-opentype'), url('../fonts/Lato/lato-bold.woff2') format('woff2'), url('../fonts/Lato/lato-bold.woff') format('woff'), url('../fonts/Lato/lato-bold.ttf') format('truetype'), url('../fonts/Lato/lato-bold.svg#latobold') format('svg')
   font-weight: 700
-  src: local('Lato Bold'), local('Lato-Bold'), url(../fonts/Lato-Bold.ttf) format('truetype')
+  font-style: normal
+
 
 @font-face
   font-family: 'Lato'
+  src: url('../fonts/Lato/lato-bolditalic.eot')
+  src: url('../fonts/Lato/lato-bolditalic.eot?#iefix') format('embedded-opentype'), url('../fonts/Lato/lato-bolditalic.woff2') format('woff2'), url('../fonts/Lato/lato-bolditalic.woff') format('woff'), url('../fonts/Lato/lato-bolditalic.ttf') format('truetype'), url('../fonts/Lato/lato-bolditalic.svg#latobold_italic') format('svg')
```
davidfischer: I don't believe we use bold and italic in the theme anywhere. Is that wrong? This will result in additional disk usage for no benefit although that isn't a huge deal. Regardless, I'd say we shouldn't include it unless there's a reason.

davidfischer:
Also, I believe these are slightly smaller than the ones we are using on .org. I will make a small PR to update the .org.

sass/_theme_font_local.sass
```diff
@@ -12,27 +12,32 @@
 
 @font-face
   font-family: 'Lato'
-  font-style: normal
+  src: url('../fonts/Lato/lato-regular.eot')
+  src: url('../fonts/Lato/lato-regular.ttf') format('truetype'), url('../fonts/Lato/lato-regular.svg#latoregular') format('svg')
```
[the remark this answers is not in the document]
jessetan: Fixed in new commit

sass/_theme_font_local.sass
jessetan: It's not possible to express bolditalic in rst syntax directly, but there are situations where it can happen, like bold text in a caption (which is italic by default) or italic text in a table header (which is already bold by default):

```rst
.. list-table::
    :header-rows: 1

    * - header
      - **bold header**
      - *italic header*
    * - cell
      - **bold cell**
      - *italic cell*
```

is rendered as:
table with italic header

jessetan:
> Also, I believe these are slightly smaller than the ones we are using on .org. I will make a small PR to update the .org.

You generated the woff(2) on .org from the original ttf right? The font files in this PR came from the original latofonts.com website.

jessetan:
Do we want to prefer a locally installed version of the font if available? This can save the download, although the version the user has installed may differ from what we use.

jessetan:
We could reduce bandwidth further by splitting each font file into a latin and one or more non-latin subsets. Browsers that view documentation that only uses latin characters will download the small font file. Other subsets will be downloaded as needed.
Since this requires some font file wrangling and a lot of testing, I prefer to not do that in this PR.

sass/_theme_font_local.sass
```diff
@@ -33,5 +38,4 @@
   font-family: 'Roboto Slab'
   font-style: normal
   font-weight: 700
-  src: url('../fonts/RobotoSlab/roboto-slab-v7-bold.eot')
-  src: url('../fonts/RobotoSlab/roboto-slab-v7-bold.eot?#iefix') format('embedded-opentype'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.woff2') format('woff2'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.woff') format('woff'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.ttf') format('truetype')
+  src: local('Roboto Slab Bold'), local('RobotoSlab-Bold'), url(../fonts/RobotoSlab-Bold.ttf) format('truetype')
```
davidfischer: Somehow this PR reverted this change. I'll fix it up.

PR #641, opened 2018-05-15
Add gitlab group to Gitlab breadcrumbs

boredstiff:
Gitlab has a concept of groups that are very commonly used in organizations, which in this url is positionally equivalent to the 'gitlab_user' key. However, for clarity's sake, I've added the 'gitlab_group' key which makes more sense.

Example of groups: https://gitlab.com/gitlab-org/ci-cd


alexwidener:
I'm not sure where this 'filter' is supposed to be added, but I can't see anything indicative in the source. 

I have a feeling it may be an issue with adding a key and the building happening on the Travis server trying to use a key it didn't previously have, in which case I'm not sure how I can submit a fix to that. 

Happy to hear any tips.

Blendify:
I don't think this is needed we could rename `gitlab_user` to something more appropriate but the API breakage is not worth it.

PR #649, opened 2018-06-06
Version 0.4.0 updates

davidfischer:
Package up the 0.4.0 version of the theme for release.


sass/_theme_font_local.sass
```diff
@@ -38,4 +38,5 @@
   font-family: 'Roboto Slab'
   font-style: normal
   font-weight: 700
-  src: local('Roboto Slab Bold'), local('RobotoSlab-Bold'), url(../fonts/RobotoSlab-Bold.ttf) format('truetype')
+  src: url('../fonts/RobotoSlab/roboto-slab-v7-bold.eot')
+  src: url('../fonts/RobotoSlab/roboto-slab-v7-bold.eot?#iefix') format('embedded-opentype'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.woff2') format('woff2'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.woff') format('woff'), url('../fonts/RobotoSlab/roboto-slab-v7-bold.ttf') format('truetype')
```
davidfischer: While testing this release, I found that this change from #642 was reverted probably in a faulty merge conflict https://github.com/rtfd/sphinx_rtd_theme/pull/639#pullrequestreview-126491151

jessetan:
@davidfischer this PR removes the font files in the `static` directory which breaks direct installs from this repo (git url in requirements.txt). This is not common use, but was it intentional? We do still have the compiled CSS and JS in `static`

davidfischer:
@jessetan, I don't think I understand the problem here. The compiled CSS does not reference the fonts in `static/` it references them in `static/fonts/FONTFAMILY/`.

jessetan:
@davidfischer sorry if I wasn't clear. I meant the font files in subdirectories of the `static` directory. Before this PR, one could put the git repo url in `requirements.txt` and the CSS, JS and fonts files were used as committed to the repo.
I am in favor of not committing the build artifacts (as described in #379), but now we are half way there. The JS and CSS are committed, but the fonts are not in the proper location to be able to use the git repo from master.

davidfischer:
I see what you're saying and now I understand.

I guess there's a couple ways forward:

- Move forward with #379.
- Move the font files in our repo (this will be somewhat awkward)
- symlinks (I don't like symlinks in repos where it can be avoided)

I favor option 1.

PR #655, opened 2018-06-28
Declare the main entry point of the package

davidfischer:
This is needed after installing with `npm` to use `require`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "sphinx_rtd_theme",
+  "main": "js/theme.js",
   "version": "0.4.0",
   "private": true,
   "devDependencies": {


jessetan:
What usecase is supported by installing using `npm`?
The theme code in `js/theme.js` does not make sense on its own. It needs the additional assets in `sphinx_rtd_theme/static`

davidfischer:
> What usecase is supported by installing using npm?

Readthedocs.org bundles the theme.js in some cases and fetches it through [bower](https://github.com/rtfd/readthedocs.org/blob/master/bower.json#L20). If we want to swap out bower for NPM and just do `require('sphinx_rtd_theme')` we will need this.

davidfischer:
Bower is deprecated and no longer recommended.

jessetan:
OK. Does it make sense to make `sphinx_rtd_theme/static/js/theme.js` the entry point as it is the generated JS or will RTD run it through it's own compilation/minification process?

davidfischer:
Typically you don't specify a compressed file as the default entry point as whatever is including it should do it's own compilation process.

PR #659, opened 2018-07-17
Add the ``extranav`` block after the ``menu`` block in the template.

schettino72:
Empty by default. Just so that users have a point to extend the template.


schettino72:
What is the point in "updating" a pull-request when you remove all the original work?

Blendify:
The feature still works the same just this approach is more versatile. You can use the super block: http://jinja.pocoo.org/docs/2.10/templates/#super-blocks to get the behavior you want.

PR #661, opened 2018-07-23
Add http protocol support

zhaojiedi1992:
can use http 
====================================

html_context = {
    "display_gitlab": True, # Integrate GitLab
    "gitlab_user": "xxxx", # Username
    "gitlab_repo": "xxxxxx", # Repo name
    "gitlab_host": "gitlab.aibeike.com",
    "theme_vcs_pageview_mode": "edit",
    "conf_py_path":  "/source/", # Path in the checkout to the docs root
    "repo_scheme": "http"
}

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -42,21 +42,21 @@
               <!-- User defined GitHub URL -->
               <a href="{{ meta['github_url'] }}" class="fa fa-github"> {{ _('Edit on GitHub') }}</a>
             {% else %}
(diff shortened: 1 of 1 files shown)


Blendify:
What is the purpose of this? Why is this an issue? Why is https an issue here?

zhaojiedi1992:
Our company's gitlab code warehouse uses a protocol that is HTTP instead of HTTPS. I have to modify the HTTPS protocol of the breadcrumbs.html inside my installation package for HTTP, but my colleagues need to modify it, and it's very troublesome. If the repo_repo option is available, we only need to set HTTP in conf.py. It is not necessary for me and my colleagues to modify the breadcrumbs.html file.

PR #665, opened 2018-08-14
Small visual improvement to captions in toctree

Blendify:
before
after


jessetan:
I'm -1 on reducing padding, makes it look cramped. I also just noticed that the captions are not exactly vertically aligned with the titles (captions slightly to the left).

If things need to stand out more, I'd rather increase the vertical padding

agjohnson:
I'm probably -1 left margin removal too, it looks unbalanced to me. We had menu captions colored at one point, which i preferred:

image

Also top margin increased for a little separation and to block sections together more.

Blendify:
I like that better too, much more contrast is much nicer. I will update this patch soon to reflect that proposal.

sass/_theme_variables.sass
```diff
@@ -51,6 +51,7 @@ $nav-link-color:                      $blue
 $nav-link-color-visited:              $purple
 $nav-link-color-hover:                lighten($nav-link-color, 6%) !default
 $nav-link-color-alt:                  hsl(33, 100%, 51%)
+$nav-caption:                         $nav-search-background-color
```
jessetan:
```suggestion
$nav-caption:                         $blue
```
Might prevent headers becoming unreadable if the search background color (the blue box behind search and logo) is ever changed, e.g. to black or gray

sass/_theme_variables.sass
```diff
@@ -51,6 +51,7 @@ $nav-link-color:                      $blue
 $nav-link-color-visited:              $purple
 $nav-link-color-hover:                lighten($nav-link-color, 6%) !default
 $nav-link-color-alt:                  hsl(33, 100%, 51%)
+$nav-caption:                         $blue
```
jessetan: This color is a little overbearing, maybe desaturate it a bit?
```suggestion
$nav-caption:                         desaturate($blue, 15%)
```

PR #666, opened 2018-08-14
Avoid printing dot when copyright holder already ends with dot

FrostyX:
When you set `copyright = "My Company, Inc."` in `conf.py` for your documentation and generate html from it, the copyright is rendered like this:

    © Copyright 2018, My Company, Inc..

Which is bad, we ideally want to have only one dot at the end.

It can't be workarounded by setting `copyright = "My Company, Inc"`, because then man pages get rendered wrong, e.g.

    AUTHOR
           My Company, Inc.

    COPYRIGHT
           2018, My Company, Inc

Therefore I am proposing this pull request. It is quite hard to fix it properly because it needs to be done per-theme and therefore it can bring some inconsistencies. If there is a better way to fix it, please let me know.

### DIFF
--- a/sphinx_rtd_theme/footer.html
+++ b/sphinx_rtd_theme/footer.html
@@ -16,9 +16,9 @@
     <p>
     {%- if show_copyright %}
       {%- if hasdoc('copyright') %}
-        {% trans path=pathto('copyright'), copyright=copyright|e %}&copy; <a href="{{ path }}">Copyright</a> {{ copyright }}.{% endtrans %}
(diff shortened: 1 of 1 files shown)


Blendify:
I would rather we remove the second added period and leave it up to the author how the copyright is punctuated.

FrostyX:
@Blendify thank you for the reply. Do you like this solution better?
(Let me know before merging, I will rebase and squash the commits)

FrostyX:
Alright @Blendify, I've rebased the PR.

Blendify:
Ok, looking for one more quick approval.

jessetan:
I'd like it even better if the user could customize the entire copyright message (including copyright sign), but that is more the responsibility of Sphinx.

LGTM

PR #672, opened 2018-09-20
Fix HTML search not working with Sphinx-1.8.

tk0miya:
Unfortunately, Sphinx-1.8 breaks HTML search on sphinx_rtd_theme.
(see: https://github.com/sphinx-doc/sphinx/issues/5460 )

To fix that, could you merge this please?
I'm sorry for trouble you.

Thanks,

Note:
Since Sphinx-1.7, Sphinx has exported search options as
`documentation_options.js`.  This starts to refer it instead of
definitions of JavaScript variables.
In addition, this also uses `js_tag()` function which added in 1.8
to support additional attributes for scripts (ex. async attribute).

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -188,6 +188,7 @@
 
   {% if not embedded %}
 
+    {# XXX Sphinx 1.8.0 made this an external js-file, quick fix until we refactor the template to inherert more blocks directly from sphinx #}
     {% if sphinx_version >= "1.8.0" %}
       <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}"></script>
       {%- for scriptfile in script_files %}
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -188,20 +188,27 @@
 
   {% if not embedded %}
 
-    <script type="text/javascript">
-        var DOCUMENTATION_OPTIONS = {
-            URL_ROOT:'{{ url_root }}',
-            VERSION:'{{ release|e }}',
-            LANGUAGE:'{{ language }}',
-            COLLAPSE_INDEX:false,
-            FILE_SUFFIX:'{{ '' if no_search_suffix else file_suffix }}',
-            HAS_SOURCE:  {{ has_source|lower }},
-            SOURCELINK_SUFFIX: '{{ sourcelink_suffix }}'
-        };
-    </script>
-    {%- for scriptfile in script_files %}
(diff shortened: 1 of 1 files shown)


tk0miya:
+1 for using `scripts` block.

Blendify:
I am all for using blocks also.

sphinx_rtd_theme/layout.html
```diff
@@ -188,20 +188,27 @@
...
-            COLLAPSE_INDEX:false,
-            FILE_SUFFIX:'{{ '' if no_search_suffix else file_suffix }}',
-            HAS_SOURCE:  {{ has_source|lower }},
-            SOURCELINK_SUFFIX: '{{ sourcelink_suffix }}'
-        };
-    </script>
-    {%- for scriptfile in script_files %}
-      <script type="text/javascript" src="{{ pathto(scriptfile, 1) }}"></script>
-    {%- endfor %}
+    {% if sphinx_version >= "1.8.0" %}
```
mitya57: I think comparing versions as strings is a bad idea. For example, if Sphinx 1.10.0 is released, that version would be considered lower than 1.8.0 with such comparison.

jessetan: Parsing the version number would be better, but for simplicity sake this solution is OK provided that Sphinx will not release a 1.10, but will go to 2.x. (that's what I gather from http://www.sphinx-doc.org/en/master/changes.html)

@tk0miya will Sphinx go from 1.9 to 2.0, or might there be 1.10, 1.11 etc?

jessetan:
An issue about the broken search was just filed #673 and I don't have time to do a full "inherit stuff from the base theme using blocks" effort, so let's proceed with these changes (once reviewed) so we keep search functionality with Sphinx 1.8+

Refactoring to extend the base theme is a longer term goal that should still happen.

sphinx_rtd_theme/layout.html
mitya57: Ok, the next version will probably be really 2.0 so in this particular case it's fine. In general it's still a bad idea :)

Blendify:
@agjohnson @ericholscher Poke, we need to do a release!

agjohnson:
I'm also not super into the conditional logic used here, but agreed it's best for now. I'll open an issue to readdress this in our next release and get a release of 0.4.2 out

sphinx_rtd_theme/layout.html
tk0miya: Sorry for late. We'll move to 2.0 on next release, not 1.10.

>I think comparing versions as strings is a bad idea. 

Agreed. But there are no tuple-formed versions inside template.
It would be better if sphinx provides it in future.

gillesdouaire:
release v0.4.2 including this sphinx 1.8 fix soon?

PR #686, opened 2018-10-09
 CSS: Restrict center-alignment to nested <p> in "math" class 

mgeier:
This is how the built-in Sphinx themes do it:

https://github.com/sphinx-doc/sphinx/blob/eb44c73b81ff4e603d9452018d145ff3c204f71f/sphinx/themes/basic/static/basic.css_t#L642-L644

I hope this is actually correct SASS ...

In case you are wondering, one example where the unrestricted centering of the `math` class has unintended consequences is here: https://nbsphinx.readthedocs.io/en/latest/code-cells.html#Math.

### DIFF
--- a/sass/_theme_mathjax.sass
+++ b/sass/_theme_mathjax.sass
@@ -1,5 +1,5 @@
 span[id*='MathJax-Span']
   color: $mathjax-color
 
-.math
+.math p
   text-align: center


jessetan:
What is the unintended consequence in the link you provided? That the formula is left-aligned rather than centered?

mgeier:
@jessetan Sorry, I should have clarified that, it's not obvious ...

The example uses an extension that uses additional CSS to left-align class `MathJax_Display`, but that's not the point of this PR.

It's about the third output with mixed text and math. This should be left-aligned instead of centered.
And in all other Sphinx templates that I tried it was.

jessetan:
Math is centered everywhere in this theme. Do you feel that left aligning math is always better than centering? Or are you more bothered by the inconsistent look in your output?
Since that is caused by an extension that left aligns MathJax output, would an adjustment to that extension or some custom CSS be enough for your use?

mgeier:
> Do you feel that left aligning math is always better than centering?

No, that's not what I'm saying. I'm just suggesting that the RTD theme should do the same as the default Sphinx themes and all 3rd party themes. Display math will still be centered, as it was intended.

I'm sorry about my example above, it's really not obvious, but you shouldn't worry too much about it. Just forget it.

The point is: The default themes and all 3rd party themes are using this CSS, so there is no reason not to do the same.

There is really no downside, so what are you worried about?

jessetan:
In the current state of the theme, this left-aligns the equation label instead of centering it for display math.
I suggest we prioritize #383, then see what needs to be done for this PR.

mgeier:
@jessetan OK, I see. This is because right now the math CSS is completely broken, and I agree that it is important to fix #383 ASAP.

Once it is fixed, my suggestion here will also work.

I've left a comment over there (https://github.com/rtfd/sphinx_rtd_theme/pull/383#pullrequestreview-182938248), probably everything can be fixed in one go.

benjaoming:
I like the suggestion to simply include this in the work combined effort in #383

PR #689, opened 2018-10-17
Style code block captions

### LINKED ISSUE #292: Captions on literal includes and code blocks are not styled

jessetan:
This PR styles the caption of code blocks and literal blocks in the same way as table captions (above item that is captioned, italic, centered, 85% size):

centered code caption

Note that this is different from the way figure captions are styled at the moment (left-aligned, below figure, 100% size). The closest we can come (because we can't move the caption below the code block) is to just make it italic:

left-aligned code caption

Fixes #292. Originally also styled the link displayed on hover, this was split into #690 so we can keep the more contentious changes confined to this PR.

### DIFF
--- a/sass/_theme_rst.sass
+++ b/sass/_theme_rst.sass
@@ -95,6 +95,13 @@
     font-size: 12px
     line-height: 1.4
 
+  .code-block-caption
+    font-style: italic
+    font-size: 85%
+    line-height: 1
+    padding: 1em 0
+    text-align: center
+
   @media print
     .codeblock, div[class^='highlight'], div[class^='highlight'] pre
       white-space: pre-wrap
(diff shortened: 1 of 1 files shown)


Blendify:
> The closest we can come (because we can't move the caption below the code block) is to just make it italic:

Have you tried using the css property `caption-side` and setting it to `bottom`?

jessetan:
That only works for table captions. Code block and literal block captions are placed before the code block in HTML. For figures this is the orher way around.
So the caption side for code block is always different to that of figures, while table caption side can be influenced using CSS. The HTML writer in Sphinx/docutils would need to be changed to make this consistent.

You had a good proposal last year https://github.com/sphinx-doc/sphinx/issues/3526

PR #696, opened 2018-11-04
Move JavaScript back to <head>

### LINKED ISSUE #328: jQuery not available in raw HTML directive
I would like to use jQuery in raw HTML blocks, for example

```rst
.. raw:: html

    <div id="mydiv"></div>
    <script type="text/javascript">
    var element = $('#mydiv');
    var text = document.createTextNode("Hello JavaScript!");
    element.append(text);
    </script>
```

This works nicely with the Sphinx default theme and all other themes I tried, except in the RTD theme, where it gives this error:

    ReferenceError: $ is not defined

I assume this is because jQuery is loaded in the end of the HTML page, where it was moved to in #78.

Is there a way to move jQuery back up again?

Or is there a work-around available?

My actual use case is allowing JavaScript output in pages converted from Jupyter notebooks, see https://github.com/spatialaudio/nbsphinx/pull/62. JavaScript outputs in Jupyter notebooks assume that jQuery is available, see http://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.Javascript.

mgeier:
It was moved away from `<head>` to the bottom of the page in #78.

This is a compressed version of #545 by @Blendify, which fixes #328.

Closes #545.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -30,6 +30,36 @@
 
   {# CSS #}
 
+  {# JavaScript #}
+  {% if not embedded %}
+
+    {# XXX Sphinx 1.8.0 made this an external js-file, quick fix until we refactor the template to inherert more blocks directly from sphinx #}
+    {% if sphinx_version >= "1.8.0" %}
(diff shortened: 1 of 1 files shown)


Blendify:
Committed in a different patch

PR #703, opened 2018-12-12
Change search background

CsatariGergely:
This pr adds the possibility to change the background
of the search area.

Signed-off-by: Gergely Csatari <gergely.csatari@nokia.com>


Gruntfile.js
```diff
@@ -125,7 +125,7 @@ module.exports = function(grunt) {
     usebanner: {
       dist: {
         options: {
-          position: 'top',
+         position: 'top',
```
jessetan: Missing space, please revert this change

Gruntfile.js
```diff
@@ -139,6 +139,9 @@ module.exports = function(grunt) {
       bower_update: {
         cmd: 'bower update'
       },
+      pip_install: {
```
jessetan: This is not related to the functional changes in the PR, please revert (also the changes to grunt.registerTask below)

In fact, doing a pip install should not be necessary for using the dev version of the theme with the docs.

sphinx_rtd_theme/theme.conf
```diff
@@ -10,8 +10,9 @@ collapse_navigation = True
 sticky_navigation = True
 navigation_depth = 4
 includehidden = True
+logo_background_color =
```
jessetan: I think nav_header_background_color, sidebar_header_background_color or something like it would be better. This is more than just the background for the logo (also contains version and search) and there might not even be a logo

Gruntfile.js
CsatariGergely: okay

Gruntfile.js
CsatariGergely: I agree, that it does not related to the functionality of the PR, so I revert it. 

For me my changes were not effective until a pip install --upgrade . , but maybe it is something with my environment.

sphinx_rtd_theme/theme.conf
CsatariGergely: Okay, I've changed it to nav_header_background_color

davidfischer:
Is this the same as #404?

sphinx_rtd_theme/theme.conf
```diff
@@ -10,8 +10,9 @@ collapse_navigation = True
 sticky_navigation = True
 navigation_depth = 4
 includehidden = True
+nav_header_background_color = 
 titles_only = 
-logo_only =
+logo_only =
```
jessetan: Please remove trailing space

jessetan:
Assuming there won't be many of these style overrides, I like this approach more than the one in #404.

Two things I'm not certain on:
- Do we need to prefix this option with something like `style_` so that we can group future style overrides, or do we think we will not have any more?
- This allows changing of the CSS `background-color` property. Would CSS `background` be a better choice so one can do background images or gradients as well?

Blendify:
> Assuming there won't be many of these style overrides, I like this approach more than the one in #404.
> 
> Two things I'm not certain on:
> 
> * Do we need to prefix this option with something like `style_` so that we can group future style overrides, or do we think we will not have any more?
> * This allows changing of the CSS `background-color` property. Would CSS `background` be a better choice so one can do background images or gradients as well?

1. Yes I think we should prefix this.
2. I think so, it gives more freedom.

CsatariGergely:
Thanks for the suggestions @jessetan and @Blendify I've applied both of them.

jessetan:
Code looks OK, just needs some documentation here: https://github.com/rtfd/sphinx_rtd_theme/blob/master/docs/configuring.rst#html-theme-options

CsatariGergely:
@jessetan documentation is added.

docs/configuring.rst
```diff
@@ -25,6 +25,7 @@ file of this repository, and can be defined in your project's ``conf.py`` via
         'prev_next_buttons_location': 'bottom',
         'style_external_links': False,
         'vcs_pageview_mode': '',
+        'style_nav_header_background': '',
```
jessetan: If you are putting this in an example, better to include an actual value.

jessetan:
LGTM

CsatariGergely:
What else is needed to merge this?

CsatariGergely:
@ericholscher , @agjohnson can you please take a look please?

ericholscher:
This feels oddly specific. Is there a reason we'd only support this one setting for a background color? It feels like having the theme be more overridable in general is a good goal, but this is a half measure. The comments on #404 touch on this, but having a full set of colors that can be overridden feels like the proper solution.

ericholscher:
On second thought, this is a good approach for now, but we should do a larger rethink in #404.

CsatariGergely:
Thanks @ericholscher . We can remove this approach once #404 is merged and use a more generic one.

PR #704, opened 2018-12-15
footer.html: show last_updated below copyright

jgehrcke:
I am using release 0.4.2 and noticed that this cosmetic problem:

screenshot from 2018-12-15 17-49-03

(the copyright notice and the last updated notice are not clearly separated).

This is how things look with the submitted patch:

screenshot from 2018-12-15 17-49-34

What do you think?


sphinx_rtd_theme/footer.html
```diff
@@ -36,7 +36,9 @@
         </span>
       {% endtrans %}
     {%- elif last_updated %}
-      {% trans last_updated=last_updated|e %}Last updated on {{ last_updated }}.{% endtrans %}
+      <div>
```
davidfischer: Just like a few lines above we have a `class="commit"` and `class="build"`, we should probably have a `class="lastupdated"`. This way whether people want it on a new line or extra spacing or whatever can be controlled with CSS.

Overall, I'm fine with it being on a new line but I just want that to be controllable.

jessetan: This should not be a `<div>` because it means the browser will close the `<p>` opened in line 16, leaving the closing `</p>` in line 44 dangling.
Better would be to use either a `<p>` which will place the text on a new line by default or a `<span>` that will place it on the same line by default.

ericholscher:
This PR now adds a span, but that doesn't effect the display. You can now hook into the display with CSS though. I'm merging this for this release, but we can discuss the default newline behavior later. 👍

jgehrcke:
Thank you!

PR #707, opened 2018-12-18
Add vertical scroll on Footer menu

dojutsu-user:
Fixes https://github.com/rtfd/readthedocs.org/issues/4496

Result:

1. When there are too many versions and translations
ezgif-4-045218f8984e

2. When there are limited versions and translations
ezgif-4-8a20fdd2583f


jessetan:
Do you have a sphinx project we can use to test this?

sass/_theme_badge.sass
```diff
@@ -35,7 +35,8 @@
       color: $black
   &.shift-up
     height: auto
-    max-height: 100%
+    max-height: 90%
```
jessetan: What is the reason for this line? Why not let it take 100% height?

dojutsu-user: With `max-height: 100%`, this happens:

ezgif-3-b0f25b01dc0b

jessetan: OK, I see. The height of 90% is fine for this example where the rtd theme is not used, but just the badge_only CSS. Here, the versions popup is not aligned to the bottom of the window.
When the rtd theme is used, the versions popup is aligned to the bottom of the window and a  height of 100% would be better.

Can you try keeping `max-height: 100%` on `&.shift-up` as it is, but setting `max-height: 90%` on `&.rst-badge` in `_theme_badge.sass:60`?

dojutsu-user: That works.
I pushed the changes.

sass/_theme_badge.sass
```diff
@@ -64,6 +65,7 @@
     left: auto
     border: none
     max-width: $nav-desktop-width
+    max-height: 100%
```
jessetan: Did you mean:
```suggestion
    max-height: 90%
```
or does 100% work for you as well?

humitos:
Just a comment.

I think there is only one difference with the current behavior (but not sure if it's browser dependent): in current versions of documentation with `alabaster` theme the vertical does not appear when scroll is not needed and in the GIF uploaded to this pull request, it does appear even when there is nothing to scroll.

captura de pantalla_2018-12-20_14-54-50

davidfischer:
> I think there is only one difference with the current behavior (but not sure if it's browser dependent): in current versions of documentation with `alabaster` theme the vertical does not appear when scroll is not needed and in the GIF uploaded to this pull request, it does appear even when there is nothing to scroll.

If I'm understanding this correctly, you're talking about the scrollbar appearing or not. I believe this is browser dependent. Chrome allowed changing some appearance on scrollbars while Firefox did not. However, I believe that is changing.

PR #710, opened 2018-12-26
Bump to development version

Juanlu001:
If one tries to install sphinx_rtd_theme from git in Read the Docs, it's not possible because pip thinks the version is the same:

```
Collecting https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip (from -r /home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/latest/requirements.txt (line 4))
  Cache entry deserialization failed, entry ignored
  Downloading https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip
  Requirement already satisfied (use --upgrade to upgrade): sphinx-rtd-theme==0.4.2 from https://github.com/rtfd/sphinx_rtd_theme/archive/a42c4d74fec660835393a4c9b1e7cb9e654ccc90.zip in /home/docs/checkouts/readthedocs.org/user_builds/poliastro/envs/latest/lib/python3.5/site-packages (from -r /home/docs/checkouts/readthedocs.org/user_builds/poliastro/checkouts/latest/requirements.txt (line 4))
```

(from https://readthedocs.org/projects/poliastro/builds/8324489/)

I propose bumping the version to `.dev0` to signal that it's more recent than the latest stable. It is common practice, and there are even some tools like [versioneer](https://github.com/warner/python-versioneer/) that do it automatically.


Juanlu001:
Just in case, the `.dev0` part comes from https://www.python.org/dev/peps/pep-0440/#developmental-releases

ericholscher:
I'm about to do another release, but how would this work in practice? I do a release with a set version, and then the next commit is bumping it to a dev release?

ericholscher:
Updated to latest version. Will merge this now and see how it goes. We can add it to our default release docs (https://sphinx-rtd-theme.readthedocs.io/en/stable/contributing.html#releasing-the-theme) if it goes well this time.

Juanlu001:
Thanks @ericholscher and @humitos!

PR #729, opened 2019-02-25
Fix #726 Remove inline javascript #

### LINKED ISSUE #726: Remove inline javascript
### Problem

As discussed [previously](https://github.com/rtfd/sphinx_rtd_theme/issues/414), when enabling content security policy, inline JS is considered unsafe.

### Reproducible Project

Some lines that are still offending:

https://github.com/rtfd/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L209-L213

The following has a [PR](https://github.com/sphinx-doc/sphinx/pull/6091) already:

https://github.com/rtfd/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/search.html#L15

#### Error Logs/Results

You get the usual "inline script is blocked by content security policy, etc" which can break functionality.

#### Expected Results

For the first script, it should be moved to a `js_t` file. For the latter, upstream changes from Sphinx need to be inherited. 

### Environment Info

- Python Version: 3.7.2
- Sphinx Version: 1.8.4
- RTD Theme Version: 0.4.3

Blendify:
Remove inline javascript


agjohnson:
Neat idea! I'd like to use template JS for this if we can, it's a bit cleaner.

This fix was applied to the built JS however, we'd need to apply this to the source JS and make sure the templating survives webpack compilation and minification. I'm not sure this is at all possible, but I haven't ever tried.

Holzhaus:
@Blendify Any plans to pick this up again?

PR #730, opened 2019-02-26
Add skip to main for screenreaders

remyabel:
For screenreaders it is useful to add a link that allows them to skip past (i.e, navigation) to the main content. You already have a div that says `role="main"`, but semantically it is more correct for this to be the `<main>` tag.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -102,7 +102,7 @@
 </head>
 
 <body class="wy-body-for-nav">
-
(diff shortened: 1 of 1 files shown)


Blendify:
The main tag is new in HTML5 which sphinx just switched over by default I think in the upcoming Sphinx 2 version. `sr-only sr-only-focusable` these seem to be from bootstrap which we do not use.

remyabel:
Oops you're right. I forgot about that.

PR #737, opened 2019-03-04
Fix: #719 Fix literals going over page width

Blendify:
THis should fix the literal example on https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/demo.html#inline-markup

Fixes: #719


Blendify:
Closing in favor of #1050

PR #742, opened 2019-03-30
Use `font-display: swap` to ensure faster Web font display

Calinou:
This makes the system font fallback display instantly while the font is loading, which results in a much better experience on slower connections (no more flash of invisible text during initial page loads).

See https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display for more information.


sass/_theme_font_local.sass
```diff
@@ -1,42 +1,49 @@
+$font-display: swap
```
Blendify: this should be moved to the global vars

Calinou:
Could this pull request be reviewed again? Thanks in advance :slightly_smiling_face:

jessetan:
LGTM

Issue #741, opened 2019-03-29
Footnotes not rendering like the demo in Sphinx 2.0


mr-ubik:
### Problem

Footnotes do not render like they do in the demo.

### Reproducible Project

- I copied and pasted the RST from the demo itself.
- I am not running any custom CSS.
- I have enabled the following extension
	```python
	extensions = [
    	"sphinx.ext.autodoc",
    	"sphinx.ext.doctest",
    	"sphinx.ext.intersphinx",
    	"sphinx.ext.todo",
    	"sphinx.ext.coverage",
    	"sphinx.ext.mathjax",
    	"sphinx.ext.ifconfig",
    	"sphinx.ext.viewcode",
    	"sphinx.ext.githubpages",
    	"sphinx.ext.napoleon",
    	"sphinx.ext.autosummary",
    	"sphinx.ext.inheritance_diagram",
    	"sphinx.ext.graphviz",
	]
	```
- I have the following `sphinx_rtd_theme` configuration:
	```python
	html_theme_options = {
    	"display_version": True,
    	"sticky_navigation": True,
    	"collapse_navigation": False,
    	"navigation_depth": 4,
	}
	```
##### Note

This issue can be reproduced easily by using Cookiecutter with the "New Python Package" template, preparing a `virtualenv` with `sphinx` and `sphinx_rtd_theme`, changing the default to RTD and making the docs after copying the demo RST to one of the default .rst files that come with the template.

#### Error Logs/Results

![Erroneous Style](https://user-images.githubusercontent.com/16547060/55236761-f727cf80-5230-11e9-9853-4dd15a2354df.png)

#### Expected Results

Display footnotes as they do in the demo.

![aragraph Level Markup — Read the Docs Sphinx Theme 0.4.3 documentation](https://user-images.githubusercontent.com/16547060/55237374-56d2aa80-5232-11e9-818d-cf49ccec2dc1.png)

### Environment Info

- Python Version: 3.7.1
- Sphinx Version: 2.0.0
- RTD Theme Version: 0.4.3

grissu:
I confirm, this happened to me after switching from Sphinx 1.8.3 to 2.0.0.

idnsunset:
Same issue here, currently using Sphinx 2.0.1, footnotes are rendered as description list in html.

iruletheworld:
I can confirm that this still exists in Sphinx 2.2

MatinF:
Also encountering this issue

PR #755, opened 2019-04-28
Scope autodoc styling to .class and .data for Sphinx 2.0

### LINKED ISSUE #752: Definition lists look different with sphinx 2.0
Example rst:

```
Foo:
    Bar
Bla:
    Blub
```

Before (1.8.3):

old

Now (2.0.1):

new

### LINKED ISSUE #754: Sphinx 2.0 produces different rendering for footnotes
Screenshot from 2019-04-26 13-57-48

_Originally posted by @bikeshedder in https://github.com/rtfd/sphinx_rtd_theme/issues/752#issuecomment-487031756_

jessetan:
Sphinx 2.0+ uses less table elements and more definition lists for layout. This causes styling for `autodoc`, `class` etc. to interfere with other layout.
This PR scopes the styling to definition lists in `.class` or `.data`, which is all we use in the demo docs. There may be more examples in the wild, but at least this fixes glossaries.

The diff looks larger than it is, only changes to the SASS is an added comment and two classes. Rest of the changes is due to added indentation.

Fixes #752. Affects #741, but does not fix it


ericholscher:
Would be good to have a review here @davidfischer or @agjohnson 👍

Issue #414, opened 2017-04-18
Content Security Policy support


rudolphfroger:
It would be good to allow users of this theme to implement a Content Security Policy which disallows inline javascript, since that should prevent injected javascript from being executed. Removing inline CSS styles would also be nice. I've searched through the repository of sphinx_rtd_theme and I think the number of inline scripts and styles is already low so my first guess is that this change shouldn't be too much work.

I can help you with testing the changes against a Content Security Policy.

Also see https://github.com/rtfd/readthedocs.org/issues/2793

Blendify:
As stated in #500 I would want to see sphinx adopt this first so we can just inherit the HTML code.
So this is not a bug for us but for sphinx.

remyabel:
The fix has now been merged in Sphinx: https://github.com/sphinx-doc/sphinx/issues/3620

Blendify:
Ok, once #545 is merged we can also adopt this.

davidfischer:
I'm a big fan of [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) but I do want to just note some challenges and design decisions that will need to be addressed in order to fully adopt CSP.

To give just a little background, CSP basically lets a server set an HTTP header (or less commonly it can be specified by `<meta>`) dictate where scripts and CSS can be loaded.  This is typically used to completely disable inline CSS/JS or to ensure that CSS/JS files are only loaded from the same origin or a whitelisted set of origins. One note is that this can mean that external JS files cannot set inline styles as well. The big advantage of CSP is it makes exploiting a XSS vulnerability much harder.

* The first and most clear step is to make sure we have no inline CSS and JS. This is essentially all Sphinx did. We have relatively few inline CSS/JS so this shouldn't be too hard.
* Should the theme add the `<meta>` tag? I'm -1 on this as we do have users who inline JS/CSS. Should it be an option? I'm less opposed to that.
* Should readthedocs.io send the CSP header? I'm -1 on that too. If we just turned that on, a lot of things would probably break like past documentation and existing inlines people have.

remyabel:
[...]

PR #756, opened 2019-04-28
Remove Modernizr, keep html5shiv

### LINKED ISSUE #724: Screen flashes on load
### Problem

When I go to any page using the RTD theme, my browser noticeably flashes white for a few dozen milliseconds before the page shows up. When I comment out `modernizr.min.js` from my HTML pages, the white flashes go away.

### Reproducible Project

I see this in every project that uses the RTD theme. Which means I can repro it here too: https://sphinx-rtd-theme.readthedocs.io/en/stable

Steps to repro:

1) Go to https://sphinx-rtd-theme.readthedocs.io/en/stable
(This already repros the bug, but it's not so easy to notice on first load. So let's go to step 2)

2) Click on "Configuration" on the left sidebar.
Expected: the Configuration page seamlessly loads
Actual: the screen flashes white _and then_ the Configuration page loads

3) You can repro this 100% of the time by clicking on **top level** links on the sidebar.

#### Error Logs/Results

There are no errors on the console.

### Environment Info

- Python Version: n/a
- Sphinx Version: 1.8.4
- RTD Theme Version: 0.4.3
- Browser: Chrome Version 72.0.3626.96 (Official Build) (64-bit)

jessetan:
Modernizr has been part of the theme for a long time, although it is not used. Since it's presence has downsides (reflow on load, increased theme size), we want to remove it.
To keep backward compatibility with older IEs, I've kept html5shiv in this PR so there is less chance of breakage.

Fixes #724, #525. Related #414. Supersedes #624


OskarPersson:
Are there any updates on this?

jessetan:
This PR was broken after our move to Webpack. I've created #848 which does the same as this PR, but updated for the new build process.
Closing this one.

PR #760, opened 2019-05-10
sticky header

GammaGames:
Relevant issues:
#758
godotengine/godot-docs/issues/1637

Before:
Screen record from 2019-05-09 23 49 57

After:
Screen record from 2019-05-09 23 49 29

Mobile:
Screen record from 2019-05-09 23 54 06

I add a flex class and some [flexbox css](https://caniuse.com/#feat=flexbox) and rearrange the sidebar's layout when the user has `sticky_header` set to `True` in `html_theme_options`.

Defaults to false to help prevent issues with version upgrading.


jessetan:
Thanks for submitting a PR, complete with documentation.
Have you tested your implementation with Internet Explorer? It is known to have spotty support for flexbox and the theme needs to work with a wide variety of browsers, old and new.

An alternative would be to place

GammaGames:
Sorry, on linux so I forgot that IE existed. There are a few solutions if it doesn't play well though:

1. Tell users not to use the setting if they want to support users on IE
2. Add another setting, `sticky_header_height` (and a mobile variant) to let users the height of the div and then use `calc(100% - {{ sticky_header_height}})` for the scroller's height.

I tried to use the second solution first but it was really hacky and flexbox is the proper way to do it. It looks like most of the bugs in IE are related to `min-height` or having bad defaults, so we could force expected defaults to get it to work sa expected.

GammaGames:
Edge 44.17763.1.0 looks good:
edge

IE 11.437.17763.0 is a little more interesting, though it's not the scroller that is the problem:
ie11

jessetan:
The use of flexbox will probably be fine after we have decided on browser support in #786

GammaGames:
I agree, my opinion is that it's kinda weird to force support for such a small market share

GammaGames:
I was only doing this for the godot docs, but they now have a solution (godotengine/godot-docs/pull/3099).

PR #770, opened 2019-07-01
Fix input form rendering on Firefox on Linux with a dark system theme

Calinou:
This prevents Firefox on Linux from rendering forms with dark styling when using a dark system theme, which looks bad on an otherwise light website theme.

## Preview


sass/_theme_layout.sass
```diff
@@ -2,6 +2,15 @@
 html
   font-size: $base-font-size
 
+// Set uniform form font styles
+// This prevents Firefox on Linux from rendering forms with dark styling
+// when using a dark system theme, which looks bad on an otherwise light
+// website theme
+input
```
jessetan: The selector for other search field styling is `.wy-side-nav-search input[type=text]`. Would be better to add the styling there to prevent accidentally restyling other input elements. The theme may not use them, but Sphinx extensions may include them and we don't want to apply our styles there accidentally.

sass/_theme_layout.sass
```diff
@@ -2,6 +2,15 @@
 html
   font-size: $base-font-size
 
+// Set uniform form font styles
```
jessetan: The comment is not needed, setting a light bg and dark text is not something that requires explaining IMHO

sass/_theme_layout.sass
```diff
@@ -2,6 +2,15 @@
...
+// This prevents Firefox on Linux from rendering forms with dark styling
+// when using a dark system theme, which looks bad on an otherwise light
+// website theme
+input
+  -moz-appearance: none
```
jessetan: Is just the bg and text color not enough to change the appearance?

Calinou: I can't see any practical difference with `-moz-apperance` none, presumably because there's some custom styling on the input form already (such as rounded corners). If you want, I can remove it.

sass/_theme_layout.sass
Calinou: I'd like the style to affect this form as well:

image

If I add to the `.wy-side-nav-search input[type=text]` selector, it won't affect that form anymore.

sass/_theme_layout.sass
jessetan: If it has no effect, it's better to remove it

sass/_theme_layout.sass
jessetan: OK, then a better selector would be `input[type=text]`

Calinou:
I updated the pull request, please take a look again :slightly_smiling_face:

Calinou:
**Update:** This should no longer be an issue in recent Firefox versions (at least on Fedora), as they now use GTK3 theming for form elements.

Still, it probably doesn't hurt to merge this.

Issue #757, opened 2019-05-03
Working hot-reload converted over to webpack with all dependencies managed by yarn


SimonBiggs:
No need to take any action on this, but I thought it might be of interest.

I forked this library and I wanted to be able to use webpack and yarn. I have managed to get everything installing and up and running (including a very fast (a second or two) hot-reload).

If you want to see the changes I made see the following:

https://github.com/pymedphys/pymedphys/tree/5168a77b3068ea4b247e11c62b60d92229271cb7/packages/pymedphys_sphinxtheme

SimonBiggs:
I'd be happy to make a pull request into this repo if there was interest.

jessetan:
This looks promising, is there anything in the gulp/browserify workflow that you have left out?

SimonBiggs:
I don't think so. I did have to bump a few of the versions of certain dependencies, I visually tested the impact on chrome and Firefox on Linux and windows and saw no issues. But it is something to be aware of.

I can make a pull request and we can begin iterating on it if you like until you're happy with it?

jessetan:
The current theme toolchain is becoming outdated and shows deprecation warnings when run. I'm in favor of moving to the next hot thing (tm), but would like feedback if this is in line with plans by RTD as a whole. Any opinions @ericholscher @agjohnson @humitos ?

ericholscher:
I definitely agree we should try and standardize this with however we end up doing things in RTD. I don't really know this stuff too well, but @davidfischer or @agjohnson probably know best. We're doing a design project over there, which might also address this.

davidfischer:
We are moving to webpack over the next few months for https://github.com/readthedocs/readthedocs.org (long overdue) so this seems in line with what we are thinking.

SimonBiggs:
Okay I'd be happy to get the ball rolling with a pull request with how I achieved it?

Blendify:
Yes please!

SimonBiggs:
@Blendify done :) see https://github.com/readthedocs/sphinx_rtd_theme/pull/771

agjohnson:
Closing this, the webpack conversion seems to be humming along quite nicely! :tada:

SimonBiggs:
Glad to hear it :)

PR #771, opened 2019-07-02
Swap to using webpack

SimonBiggs:
This is pull request is a follow up to the following issue:

https://github.com/readthedocs/sphinx_rtd_theme/issues/757

The live reload server is run with `yarn start`, the production build is built with `yarn build`.


src/sass/_theme_font_local.sass
```diff
@@ -0,0 +1,6 @@
+$lato-font-path: "~lato-font/fonts"
+$lato-variants: ('normal': 400, 'bold': 700)
```
SimonBiggs: I'm not sure if this here is enough. Someone with more understanding of what fonts are required might need to make an adjustment to this file to bring it into compatibility with previous results.

davidfischer: We definitely use the italics one as well. Bolditalic is probably unused.

jessetan: I think we should include it.
Lato is the font for the content. The combination of Sphinx, extensions, and custom CSS could generate a need for a bolditalic fontface. If this is done using a separate `@font-face` rule it will only be downloaded when needed. If it is not needed, it will just be 1 line in CSS, so quite inexpensive.

package.json
```diff
@@ -3,21 +3,32 @@
...
-  "devDependencies": {
-    "bower": "^1.8.4",
+  "scripts": {
+    "start": "yarn webpack-dev-server --open --config webpack.dev.js",
+    "build": "yarn webpack --config webpack.prod.js"
+  },
+  "dependencies": {
+    "bourbon": "~4.3",
+    "bourbon-neat": "~1.7",
     "browserify": "^13.0.0",
```
jessetan: Do we still need the browserify dependency here?

src/sass/_theme_font_local.sass
SimonBiggs: So what would I need to change here, (this is not my forte).

package.json
SimonBiggs: I shall try without it.

SimonBiggs:
> There are a lot of font files in ` sphinx_rtd_theme/static/fonts`. Do you need them if they are installed as dependencies?

Given those dependencies won't be shipped in pypi I would say yes? Web pack is actually outputting those files from the dependencies when it runs the build.

jessetan:
I can not get the dev server to work unfortunately.
If I run `yarn build` I get a freshly compiled `theme.js` in `sphinx_rtd_theme/static/js`.
If I run `yarn start` I see that `theme.js` is recompiled every time I make a change, but the resulting file is nowhere to be found. I also don't see Sphinx being run if I change the rst files.

SimonBiggs:
> I can not get the dev server to work unfortunately.
> If I run `yarn build` I get a freshly compiled `theme.js` in `sphinx_rtd_theme/static/js`.
> If I run `yarn start` I see that `theme.js` is recompiled every time I make a change, but the resulting file is nowhere to be found. I also don't see Sphinx being run if I change the rst files.

Happy to have this pull request closed. I thought it was close to parity, but obviously it is still missing too much.

agjohnson:
@SimonBiggs I'd love to see this PR merged, let me know how I can help. Otherwise, I'll probably pull this down and poke it more.

I don't have really strong opinions about the stack changes, but webpack replaces a lot of fragile tooling and has shown to be a huge improvement over our gulp build system, and looks like a similar improvement over grunt.

I don't have strong opinions on yarn, but the core team doesn't do much JS. Having less JS dependencies is a strong benefit for us, and the reason we're targeting webpack for RTD. If npm can do everything that yarn is doing here, I'm +1 on skipping yarn.

SimonBiggs:
@agjohnson I'm more than happy to have yarn replaced for npm. There are certainly some benefits, but not enough to force people to step away from what they're used to.

I think the best bet would be for you to jump in and poke it, feel free to push to this branch. I can help out if you have questions about decisions I made.

agjohnson:
@SimonBiggs Sounds good. I opened up a PR with what I've changed to get going: #781

Some notes about my attempt:

* `node-sass` fails without python2 -- more specifically `node-gyp` fails, it is not yet compatible with python3. RTD has switched core to using python3 however. I guess we need to note that python 2.7 is required to run this until node-gyp finally goes python3? This is suboptimal
* Core team isn't super flexible on node versions. Hopefully webpack isn't as prone to being picky about versions, but our current Gulp system needs node 8.x to run for some nondeterministic reason -- so that's what core team is stuck with atm :( I use nodenv/pyenv/etc for multiple version support, but I don't know what the rest of the team uses.

SimonBiggs:
> node-sass fails without python2 -- more specifically node-gyp fails, it is not yet compatible with python3. RTD has switched core to using python3 however. I guess we need to note that python 2.7 is required to run this until node-gyp finally goes python3? This is suboptimal

`node-gyp` vs 4.0.0 already works with Python 3:

https://github.com/nodejs/node-gyp/blob/master/CHANGELOG.md#v400-2019-04-24

Upgrading to version 4.0.0 of `node-gyp` should fix that issue I would imagine.

> Core team isn't super flexible on node versions. Hopefully webpack isn't as prone to being picky about versions, but our current Gulp system needs node 8.x to run for some nondeterministic reason -- so that's what core team is stuck with atm :( I use nodenv/pyenv/etc for multiple version support, but I don't know what the rest of the team uses.

Not sure about that unfortunately.

agjohnson:
> node-gyp v4.0.0 already works with Python 3

Ah! I looked into this a bit more. It's actually bourbon-neat bringing in an old node-sass, tied to an old node-gyp. Updating to the latest bourbon-neat fixes the python 2/3 issue :+1: 

But now nothing builds.

This is wyrm showing it's age though, all the errors are from changes to Neat. I'm not sure the best thing to do here.

> Node, version 12.1.0 worked fine after bumping the version of bourbon-neat

So yeah, ultimately a new bourbon-neat fixes a couple of issues then.

Our options, in no particular order of awfulness:

* Try to get fixes into wyrm. I don't think `snide` has bandwidth for this.
* Fork wyrm, apply fixes. Frankly, wyrm doesn't get much use besides us, we might as well maintain it in some fashion. We might be able to see about reassignment of the repo.
* Vendor wyrm until we figure things out
* Use another framework -- big :-1: from me here, I'd rather put this type of effort it a separate theme

Any thoughts? @jessetan and @Blendify specifically

I vote we do a light fork of wyrm into the readthedocs org, fix the Neat issues, install our fork from github here, and worry about the question of maintainership later.

For this PR, are we comfortable merging this with Python 2 and Node 10 support? We'd hopefully pretty quickly figure out the issues with Wyrm and support Python 3 and Node 10+. We can block otherwise, but I'm fine accepting Python 2 + Node 10 for a short while.

@SimonBiggs if you haven't already, feel free to either merge my branch in here, or give us access to your branch via maintainer permissions:
https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

@jessetan it might be good to have you poke the PR i added some fixes on, and see if it solves the issues you hit (minus the obvious issues with Python/Node of course)

SimonBiggs:
> But now nothing builds.
> 
> This is wyrm showing it's age though, all the errors are from changes to Neat. I'm not sure the best thing to do here.

My yarn lock file has all the exact version in it that successfully build with this configuration...

SimonBiggs:
> @SimonBiggs if you haven't already, feel free to either merge my branch in here, or give us access to your branch via maintainer permissions:
> help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Already have:

image

However I believe you set your pull request to target the branch on this repo:

image

So I don't have merge permissions here.

SimonBiggs:
> * Try to get fixes into wyrm. I don't think `snide` has bandwidth for this.
> * Fork wyrm, apply fixes. Frankly, wyrm doesn't get much use besides us, we might as well maintain it in some fashion. We might be able to see about reassignment of the repo.
> * Vendor wyrm until we figure things out
> * Use another framework -- big 👎 from me here, I'd rather put this type of effort it a separate theme

None of these are needed :) The biggest thing I added here I believe is probably my "yarn lock file". Which was hours of dependency iteration finding the right combo that requires no changes from wyrm etc.

agjohnson:
> My yarn lock file has all the exact version in it that successfully build with this configuration...

I tried yarn initially and hit installation issues with the requirements. Specifically, bourbon-neat does still pin an old version of node-sass, which brings in node-gyp 3.8, and that is what fails with python 3. This doesn't seem to be a yarn vs npm thing, using python 2 I can install and build with both npm and yarn.

> None of these are needed :) The biggest thing I added here I believe is probably my "yarn lock file". Which was hours of dependency iteration finding the right combo that requires no changes from wyrm etc.

This PR does build assets correctly, however just not with python 3 or node 12 (apparently, i didn't try yet). I'm describing fixing these issues later though, which requires a new version of Neat, which will require changes to Wyrm.

SimonBiggs:
> > My yarn lock file has all the exact version in it that successfully build with this configuration...
> 
> I tried yarn initially and hit installation issues with the requirements. Specifically, bourbon-neat does still pin an old version of node-sass, which brings in node-gyp 3.8, and that is what fails with python 3. This doesn't seem to be a yarn vs npm thing, using python 2 I can install and build with both npm and yarn.
> 
> > None of these are needed :) The biggest thing I added here I believe is probably my "yarn lock file". Which was hours of dependency iteration finding the right combo that requires no changes from wyrm etc.
> 
> This PR does build assets correctly, however just not with python 3 or node 12 (apparently, i didn't try yet). I'm describing fixing these issues later though, which requires a new version of Neat, which will require changes to Wyrm.

kk :)

agjohnson:
I took a quick swing at upgrading Neat and it wasn't bad. I have the theme building with the upgrade (not *well*, but it builds). The rest of the Wyrm tool chain is more broken and not documented -- `kss`, `node-sass`, etc need more attention than the Neat upgrade.

https://github.com/agjohnson/wyrm/pull/1

So I think it's doable without a lot of effort, but I've changed my mind on maintainership. I don't want to be responsible for maintaining this. I think a vendor fork is the best path forward, allowing us to continue with this PR.

I have some additions to my PR, I'll merge into this branch after.

SimonBiggs:
Makes sense.

jessetan:
@agjohnson which version of bourbon-neat did you use? I've found that version 1.9 works with Node 12 and python 3. It is the last one in the 1.x branch. From version 2.x the incompatible changes start.

agjohnson:
@jessetan yes! :100: that does it for me, on a fresh install even.

I skipped even testing 1.9, assuming the whole 1.x series has the bad deps.

So that, at least for now, puts us in maintainable place. We probably still need to fork Wyrm eventually, but we can kick that can way down the road.

agjohnson:
Okay. Everything is merged here. The last change I'd like some feedback on is:
https://github.com/readthedocs/sphinx_rtd_theme/pull/771/commits/cb87d8f406177a08558ff4e0a3c23cd2b6272a59#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

It's proper to make all of these dev dependencies, no? We don't require any of these to run sphinx_rtd_theme, just to work on the theme.

I think I'm :+1: on merging after some docs are updated. I'd like to merge translation docs before reworking docs here though.

Code review on this PR is not super helpful, it's probably most helpful to check out the branch and try:

```
npm install
npm run build
npm run start
```

`npm install` should pass, `npm run build` should create css/js assets without changes to the repo, `npm run start` should build these assets, and also build docs on any changes, live reloading on port 1919.

SimonBiggs:
> Okay. Everything is merged here. The last change I'd like some feedback on is:
> [cb87d8f#diff-b9cfc7f2cdf78a7f4b91a753d10865a2](https://github.com/readthedocs/sphinx_rtd_theme/commit/cb87d8f406177a08558ff4e0a3c23cd2b6272a59#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)
> 
> It's proper to make all of these dev dependencies, no? We don't require any of these to run sphinx_rtd_theme, just to work on the theme.

I believe any dependency which is for something that strictly helps with the development process (like linting for example) they belong in dev dependencies. Anything that is required to make "npm build" work, they belong in dependencies.

There is a flag that can be passed to npm install which only installs the dependencies and not dev dependencies. That can be used in the build step of a continuous deployment pipeline to only download what is necessary for the build.

jessetan:
I think all our dependencies are dev dependencies, since they are used during development. The theme is never installed using npm with dependencies.

jessetan:
All npm commands work as expected.

Three remarks:

1. If there is no built documentation in `docs/build` (i.e. during the first run), `npm run start` will show "Cannot GET /" in the browser and does not reload.
2. `npm run start` -> `npm run dev`?
3. I'm missing a `npm run clean` or similar command. Might not be needed but it has been useful in the past.

agjohnson:
Also, npm says:

> "dependencies": Packages required by your application in production.

Which would be no packages. I'll leave these dependencies in dev dependencies for now.

> If there is no built documentation in docs/build (i.e. during the first run), npm run start will show Cannot GET /" in the browser and does not reload.

Good point. I'll see if there is a way to make the dev server avoid this.

> npm run start -> npm run dev?

:+1: I actually made this change, but reverted it. `npm run start` is normally the standard, but I agree `npm run dev` makes more sense

> I'm missing a npm run clean or similar command. Might not be needed but it has been useful in the past.

That should be easy enough to add and seems like it would be useful.

agjohnson:
Tests are all breaking because of a change introduced I believe in #405, where setup.py is importing the sphinx_rtd_theme package to get the __version__. This doesn't work. Instead of blocking more on this PR, I'll address with a separate PR right after.

docs/contributing.rst
```diff
@@ -42,31 +35,31 @@ Set up your environment
 Making changes
 --------------
 
-Changes to the theme can be compiled and tested with the default Grunt task:
+Changes to the theme can be compiled and tested with webpacke:
```
SimonBiggs: Webpacke --> webpack

docs/contributing.rst
```diff
@@ -42,31 +35,31 @@ Set up your environment
...
 
 .. code:: console
 
-    grunt
+    npm run dev
```
SimonBiggs: Ended up going with run dev instead of run start?

setup.py
```diff
@@ -55,7 +53,7 @@ def run(self):
 
 setup(
     name='sphinx_rtd_theme',
-    version=__version__,
+    version='0.4.3.dev0',
```
davidfischer: I see that you're addressing this in another PR. You can also consider reading the version from `package.json`. Unfortunately, that file can't also read the version from the Python theme module (and the theme module can't read from `package.json`). I've seen a couple approaches here from reading from `package.json` to having a separate file with just the version (eg. `sphinx_rtd_theme/version.py`) to parsing the text of `sphinx_rtd_theme/__init__.py` instead of importing it. I don't have a particularly strong opinion although any of them seem preferable to hardcoding it.

docs/contributing.rst
davidfischer: The script in `package.json` is named "start". We should fix it or the docs.

docs/contributing.rst
agjohnson: Oopse

davidfischer:
Overall, this is a **huge** improvement over what was there before, IMO. Thanks a lot for the contribution and persistence @SimonBiggs.

docs/contributing.rst
agjohnson: Aye, this might be non-standard node, but jessetan had the same feedback around `npm run start` . I actually added #797 to maybe move towards putting everything in `setup.py` even :upside_down_face:

agjohnson:
Merging! :rocket: 

Thanks everyone! And thanks again @SimonBiggs for getting this rolling!

SimonBiggs:
> Overall, this is a **huge** improvement over what was there before, IMO. Thanks a lot for the contribution and persistence @SimonBiggs.

My pleasure :) thank you for the improvements on it, and for the maintenance of this theme :)

PR #775, opened 2019-07-11
Horizontal Lists: Use full width

Blendify:
Before:

image

After:

image


Blendify:
As far as I know, hlist's are not used by the auto API generator.as for other sphinx elements, this is specific to only the `.. hlist::` directive.

PR #785, opened 2019-07-19
Fix styling of generated cpp documentation

jessetan:
Before:
Screen Shot 2019-07-19 at 16 08 14

After:
Screen Shot 2019-07-19 at 16 07 02


sass/_theme_rst.sass
```diff
@@ -328,7 +328,7 @@
   // that output similar, but not quite the same nesting so this is really the best we can do.
   dl:not(.docutils)
     // In Sphinx 2.0+ most DLs no longer contain .docutils, needs tighter scoping using next line
-    &.class, &.data
+    &.class, &.data, &.type, &.function
```
ericholscher: Annoying that there isn't just a superclass for all of these.

jessetan: Yes, that's why I missed these classes previously.
Could be something to feedback to Sphinx, they made some changes to which elements are used for generated documentation, but not everything has a useful class. Some are just `<em>` for instance, which makes solving issues like #647 more difficult

agjohnson:
I'm going to add some more domains in a separate PR, i've found the output is help in comparing sphinx versions.

PR #790, opened 2019-07-24
Update dev dependencies

RDIL:
Updates dev dependencies to the latest version.


jessetan:
Thanks for your PR. We are in the process of moving the build toolchain to webpack, so some of these dependencies will no longer exist or should be updated after the move to webpack.

agjohnson:
Blocking on #771

agjohnson:
#771 is merged, feel free to rebase. I think all of these deps are removed now.

RDIL:
Okay, I'll have dependabot trigger another bump.

package.json
```diff
@@ -10,8 +10,8 @@
   "dependencies": {},
   "devDependencies": {
     "bourbon": "~4.3",
-    "bourbon-neat": "~1.9",
-    "css-loader": "^2.1.1",
+    "bourbon-neat": "~4.0",
```
agjohnson: We can't use > 1.9 current due to Wyrm

RDIL:
Done @agjohnson!

package.json
```diff
@@ -10,8 +10,8 @@
...
+    "css-loader": "^3.1.0",
```
agjohnson: Tested this and things seemed to work :+1:

agjohnson:
Thanks! Looks great! I'll have to note the pin here

Issue #782, opened 2019-07-18
Fonts are different between local html and readthedocs.org 


daianagan:
Hi!! I am new to Sphinx and ReadtheDocs, and I really like the way it looks on the webpage, though I need some documents to not be on the web (so I want to only make the html file and have it on my computer)
 
### Problem

When opening my index.html file on my computer the fonts are not the same as the ones I see when I open the project in ReadTheDocs webpage (after pushing to github and importing it to readthedocs.org). On readthedocs its a sans-serif font and locally its serif.

I also tried downloading the ReadsTheDocs html zip to check and the font I see is the same as when I make it locally in my computer. 
I have already installed the fonts and nothing changed either. Maybe there is something I need to change on the CSS files for what I have read but I am not familiar with that

I have MacOS Mojave version 10.14.4
### Reproducible Project

The commands for the local html file:
```
mkdir docs
cd docs
pip install pip install sphinx-rtd-theme
sphinx-quickstart (all default params)
(modified conf.py to have html_theme = "sphinx_rtd_theme")
make html
```
For ReadtheDocs view:
Import manually from https://github.com/daianagan/daiana_rtd_issue 

RST:
```rst

.. Test documentation master file, created by
   sphinx-quickstart on Thu Jul 18 13:15:32 2019.
   You can adapt this file completely to your liking, but it should at least
   contain the root `toctree` directive.

Welcome to Test's documentation!
================================

.. toctree::
   :maxdepth: 2
   :caption: Contents:



Indices and tables
==================

* :ref:`genindex`
* :ref:`modindex`
* :ref:`search`

```

#### Error Logs/Results

<img width="1143" alt="Captura de Pantalla 2019-07-18 a la(s) 14 39 52" src="https://user-images.githubusercontent.com/29231158/61481059-9fdb7580-a96d-11e9-8338-77d2ce60d754.png">

#### Expected Results

<img width="1137" alt="Captura de Pantalla 2019-07-18 a la(s) 14 54 06" src="https://user-images.githubusercontent.com/29231158/61480885-54c16280-a96d-11e9-9a88-43692795afb9.png">

### Environment Info

- Python Version: Python 2.7.14 (also tried with python 3.6)
- Sphinx Version: 1.6.6
- RTD Theme Version: latest (pip installed today)

foo-git:
I'm new to GitHub, but I'm also experiencing this behaviour, causing my account creation. I might have some insight on the bug, but no fix :(

[...]

PR #808, opened 2019-07-26
Fix a number of issues with Webpack

agjohnson:
A few of these squeezed by QA and testing:

* badge_only.css was copied to the static output path, it wasn't
  actually building it seems. Added second entry for this to make it
  easier.
* jQuery was not being treated as an external library and was vendored
  into our JS bundle!
* Having the fonts in a relative path from CSS ('../fonts'), causes CORS
  issues. Moved `/fonts` to `/sphinx_rtd_theme/static/css/fonts` essentially. See #782 
* The dev server wasn't compiling assets consistently for me, tuned it hopefully
* This does output an unused `js/badge_only.js`, which I believe is for using CSS through JS. This is a webpack 4 thing and it can't be removed. I say disregard it for now.

Requires #807, based on that until it's merged, then can be repointed to master
Refs #782


jessetan:
My build cannot find fontawesome:
```
ERROR in ./src/sass/theme.sass
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleNotFoundError: Module not found: Error: Can't resolve '../fonts/fontawesome-webfont.eot' in '/Users/jesse/Documents/development/sphinx_rtd_theme/src/sass'
```

This looks similar to an issue with tilde in `$fa-font-path` and `sass-loader` reported in https://github.com/webpack-contrib/sass-loader/issues/566

If I downgrade `sass-loader` to something <7, like 6.0.7 it works.
This is an issue in master as well.

agjohnson:
Hrm odd. I'll poke this on my environment to see what I find.

jessetan:
Looks like it is not unique to my setup: #820

jessetan:
@agjohnson how is the progress on this? It is one of the things that is holding back a release.

agjohnson:
I haven't had time in Sep to focus on the theme, but have some changes on our roadmap that I'm assigned to. I'll be getting back to some of this work in a week or so. If you find a fix, feel free to jump in -- I still have not tried to reproduce the error yet even, so I'm not sure what I'll hit yet.

jessetan:
Thanks @agjohnson, sounds good.
I have no explanation why we have different results, but the issue with the font awesome path is resolved for me if I use `"sass-loader": "^6.0.7"`

davidfischer:
> badge_only.css was copied to the static output path, it wasn't actually building it seems. Added second entry for this to make it easier.
> jQuery was not being treated as an external library and was vendored into our JS bundle!

These look resolved to me

> I have no explanation why we have different results, but the issue with the font awesome path is resolved for me if I use "sass-loader": "^6.0.7"

I don't have this issue and the configuration worked as-is for me.

> Moved /fonts to /sphinx_rtd_theme/static/css/fonts

The fonts "work" but they are duplicated. More or less the same fonts exist in `sphinx_rtd_theme/static/fonts/` and `sphinx_rtd_theme/static/css/fonts/`. Both of these are copied to the docs output when running `npm run dev` or to the module  bundle when running `python setup.py sdist`.

`sphinx_rtd_theme/static/fonts/` does not appear to be used at all and could probably just be deleted. 

`sphinx_rtd_theme/static/css/fonts/` is created and populated by `npm run build`. Its could probably be removed and the directory itself added to `.gitignore`.

pedro-w:
Just to add to this I get the [same error](https://github.com/readthedocs/sphinx_rtd_theme/pull/808#issuecomment-523494248) as @jessetan when running these commands:
```
git clone https://github.com/readthedocs/sphinx_rtd_theme.git
git checkout agj/more-webpack-fixes
pip3 install -e '.[dev]'
npm install
npm run dev
```
@davidfischer is this what you are doing or is there another way to do it?

jessetan:
@davidfischer perhaps it is different environments. I use a pipenv shell on macOS 10.13.6 with npm 6.12.0, Python 3.7.3. The error appears when I run `npm run dev` right before Sphinx is started.

pedro-w:
I looked a bit more into `sass-loader` versions

* 6.x OK
* 7.0.x OK
* 7.1.0 OK
* 7.2.0 Error as above
* 7.3.x Error as above

Don't know if that helps?

davidfischer:
> 7.1.0 OK

This is the version that I had installed so that may explain things.

I had a `package-lock.json` file which is currently gitignored and was pinned to 7.1.0. We should probably commit that file and pin to a specific version. That seems to be what npm suggests:

    npm notice created a lockfile as package-lock.json. You should commit this file.

davidfischer:
I created a PR that could merge into this one that commits the `package-lock.json` and pins `sass-loader` to 7.1: #827

pedro-w:
I am out of my depth on all this npm stuff, but it would be great if someone could figure out what causes the problem (since `sass-loader` 8 is already out). I had a further poke around and one thing I noticed is that the version of the `css-loader` dependency jumps from 0.28 to 2.0 between these two versions. Maybe @evilebottnawi might know.

evilebottnawi:
@pedro-w what is problem?

pedro-w:
@evilebottnawi thanks for the quick response. If you look at [this comment](https://github.com/readthedocs/sphinx_rtd_theme/pull/808#issuecomment-523494248) the problem is that sphinx_rtd_theme's build system can't resolve the fontawesome fonts. Further, it seems that with  `sass-loader` v7.1.0 and before it works, and v7.2.0 and later does not work. So I was hoping you would know straight away if were there any changes to the resolving code in between those two versions that could cause this? Or, has anything like this come up before? Thanks!

evilebottnawi:
@pedro-w maybe you can create minimum reproducible test repo and open issue in sass-loader?

pedro-w:
I believe if you change the import here
https://github.com/readthedocs/sphinx_rtd_theme/blob/ddf840cb7206d7c45270560f077b12daa147f915/src/sass/theme.sass#L31
to `@import ~font-awesome/scss/font-awesome` then it will solve the issues with `sass-loader` versions (7.1 and earlier _vs_ 7.2 and later)

davidfischer:
I created a small PR that merges into this one based on your suggestion. It worked for me and I'd be happy to be able to take a later version. #830

agjohnson:
Looking through some of the remaining comments, there are already some fixes in master that are not in this branch.

> sphinx_rtd_theme/static/css/fonts/ is created and populated by npm run build. Its could probably be removed and the directory itself added to .gitignore.

`master` has the fix for this. I'll try rebasing this branch, but i don't get the `static/fonts` path generated on master + this branch.

If there is any feedback or if anyone else wants to get review in, I'd like to get this merged soon.

davidfischer:
I think we accidentally merged this into `agj/format-webpack-js` rather than master. Shall I revert it and repoint at master?

agjohnson:
Should be merged manually now

Issue #797, opened 2019-07-25
Put all build commands in setup.py command classes


agjohnson:
In the translation workflow, Babel adds helper scripts to `setup.py`, and #778 added some helpers to call all of the required translation pieces.

#771 uses a standard `npm run` to wrap around the webpack calls, however would it make more sense to also put these in `setup.py`? We could even extend the build command class to execute the webpack build prior to the rest of the packaging build steps.

* `python setup.py build` is consistent and fits with our release packaging. It keeps commands in a singular place
* `npm run build` implies that the build process requires npm and installation through npm
* Middle ground: `python seutp.py build` calls `npm run build`, it's all about choices.

Which are you most likely to use (or want to use)?

jessetan:
I'm fine with using `setup.py`, but npm or yarn is still needed to pull in the dependencies.

agjohnson:
Yup of course, I'm just referring to removing `scripts` from package.json.

A quick test shows `python setup.py build` is really nice, `bdist` and friends chain off of the Webpack build nicely.

I added a POC PR at #809 that shows this. The `npm run dev` command has waffled again to not collide with `python setup.py develop`. Suggestions welcome there.

agjohnson:
Already merged. Commands are `python setup.py build` and `python setup.py watch` for now.

PR #809, opened 2019-07-26
Add webpack commands into setup.py

agjohnson:
This is a POC that shows building webpack through standard `setup.py`
commands. Any call to `setup.py build` or `bdist` or `sdist` will
trigger a Webpack build of the assets first. A non-zero exit code will
halt the process.

Also, moved the `npm run dev` command, which here is `python setup.py
watch`, though there is perhaps something better here. There is already
`python setup.py develop`, which has a separate function, so I don't
want to collide there.

Example output here:
https://gist.github.com/agjohnson/cdaab364fe598daa7f3bef750cfb84dd

Refs #797

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ class WebpackBuildCommand(setuptools.command.build_py.build_py):
     def run(self):
         if not 'CI' in os.environ:
             subprocess.run(['webpack', '--config', 'webpack.prod.js'], check=True)
-        super(WebpackBuildCommand, self).run()
+        WebpackBuildCommand.run(self)
 
 
 class WebpackDevelopCommand(distutils.cmd.Command):
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ class WebpackBuildCommand(setuptools.command.build_py.build_py):
     def run(self):
         if not 'CI' in os.environ:
             subprocess.run(['webpack', '--config', 'webpack.prod.js'], check=True)
-        super().run()
+        super(WebpackBuildCommand, self).run()
 
 
 class WebpackDevelopCommand(distutils.cmd.Command):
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -5,12 +5,43 @@
...
 from setuptools import setup
 
 
+class WebpackBuildCommand(setuptools.command.build_py.build_py):
+
+    """Prefix Python build with Webpack asset build"""
+
+    def run(self):
+        if not 'CI' in os.environ:
+            subprocess.run(['webpack', '--config', 'webpack.prod.js'], check=True)
```
jessetan: This does not work for me as webpack is not in my PATH. Perhaps you installed it as a global package?
It works with `npm run build` because npm adds `node_modules/.bin` to the PATH (https://docs.npmjs.com/cli/run-script)

```suggestion
            subprocess.run(['node_modules/.bin/webpack', '--config', 'webpack.prod.js'], check=True)
```

setup.py
```diff
@@ -5,12 +5,43 @@
...
+
+    def initialize_options(self):
+        pass
+
+    def finalize_options(self):
+        pass
+
+    def run(self):
+        subprocess.run(
+            ["webpack-dev-server", "--open", "--config", "webpack.dev.js"],
```
jessetan:
```suggestion
            ["node_modules/.bin/webpack-dev-server", "--open", "--config", "webpack.dev.js"],
```

setup.py
agjohnson: This is probably because I have `node_modules/.bin` in my `PATH` actually. Using the local node_modules prefix seems most correct. I've always used the PATH pattern, but perhaps it's not standard.

FRidh:
Fetching stuff from the web during build-time is bad practice, please don't do that.
https://github.com/readthedocs/sphinx_rtd_theme/issues/1014

PR #813, opened 2019-08-26
Print a warning when building docs with a pre-release version of the theme

jessetan:
As it says in the description, this will print a warning when building a document and the theme version is a pre-release according to `parse_version` from pkg_resources.
The intention is to make it more attractive to do more frequent pre-releases as described in #379 by giving authors a more obvious indication that they are running a version that they should be only using temporarily.

I've checked the functionality of this PR, but any input on my attempt at translation setup is welcomed.


sphinx_rtd_theme/__init__.py
```diff
@@ -23,6 +34,14 @@ def get_html_theme_path():
 def setup(app):
     app.add_html_theme('sphinx_rtd_theme', path.abspath(path.dirname(__file__)))
 
+    # Warn if a prerelease version of the theme is used
+
+    parsed_version = pkg_resources.parse_version(__version__)
+    if parsed_version.is_prerelease:
+        logger.warning(_('You are using a pre-release version (%s) of sphinx-rtd-theme.'), __version__)
+        logger.warning(_('Use a stable release if you encounter problems.'))
+        logger.warning(_('See <https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html>.'))
```
Blendify: I think the URL is unnecessary I think a user who may have installed a dev version already knows how to install the theme

jessetan: I agree it is a bit much, but I want to make this as actionable as possible.

PR #822, opened 2019-10-03
Use bumpversion for version number management

jessetan:
This is an alternative for #812. It uses a CLI tool to bump the version number so we can manage it in a single place.
The benefit over #812 is that it actually writes the version number to the files, which means that we do not rely on `pkg_resources.get_distribution` (which breaks if the package is not pip-installed) and we do not read it from setup.cfg (which may cause trouble with reathedocs.org code)

I hope that this PR contributes to the speed at which we can release dev builds, so we can get better and earlier feedback.

Release instructions have been updated, but the process is:

1. bump the version using `bump2version [major|minor|patch|dev]`. Using `major|minor|patch` will always have a `dev0` suffix, using `dev` will bump the number in the suffix (`dev0` -> `dev1` -> `dev2`)
2. If you want to release a dev version, proceed with the other steps
3. If you want to release a release version, run `bump2version --allow-dirty release` to remove the `devX` suffix. The `--allow-dirty` flag is needed to prevent complaints about non-committed changes.


agjohnson:
Ah, interesting idea. I didn't know about this tool. We have scripts that perform some of the same functions for our other repositories already. It might make sense to standardize on a common tool instead of custom scripts across our repositories.

.bumpversion.cfg
```diff
@@ -0,0 +1,19 @@
...
+[bumpversion:part:release]
+optional_value = gamma
+values = 
+	dev
+	gamma
+
+[bumpversion:file:setup.py]
+
+[bumpversion:file:sphinx_rtd_theme/__init__.py]
+
```
agjohnson: Perhaps `package.json` as well then? I know we don't package anything there at the moment, but maybe a good place to keep the version number updated either way

docs/contributing.rst
```diff
@@ -89,10 +89,12 @@ Releasing the theme
 
 To release a new version of the theme, core team will take the following steps:
 
-#. Bump the version in ``sphinx_rtd_theme/__init__.py``, ``setup.py`` and
-   ``package.json``.  We follow `semver <http://semver.org/>`_ and `PEP440`_
+#. Bump the version by running ``bump2version [major|minor|patch|dev]``.
```
agjohnson: Minor nitpick: I'm a little confused by how to call this. The way I parse this is if I want to bump the minor release number, I'd run: `bump2version minor` and we'd have `0.5.0dev0`.

It's not supposed to be run as `bump2version 0.5.0dev0` is it?

docs/contributing.rst
```diff
@@ -89,10 +89,12 @@ Releasing the theme
...
+   We follow `semver <http://semver.org/>`_ and `PEP440`_
    (with regards to alpha release and development versions). The version
    increment should reflect these releases and any potentially breaking changes.
+#. If the new version is not a dev version, run ``bump2version --allow-dirty release`` to
```
agjohnson: Is `release` a string, or the release version number? I assume it's a string `release`.

.bumpversion.cfg
agjohnson: Oh, also the version number in the docs could be a static string updated with this tool.

docs/contributing.rst
jessetan: Correct.
We are at `0.4.3.dev0` now, if you run `bump2version dev`, the version will be `0.4.3.dev1`.
If you run `bump2version minor`, the version will be `0.5.0.dev0`.

If you want to force a particular version, you can run `bump2version --new-version 0.9.2 patch` (the word `patch` has no effect here).

I've added some more text to the instructions

.bumpversion.cfg
jessetan: Oops, forgot `package.json`, now added.

Also added the `conf.py` of the documentation, there is no version number in the documentation body text.

docs/contributing.rst
jessetan: Yes it is the exact string ``release``

Letme:
How about we use `setuptools_scm` where we generated a fixed file `__version__.py` at packaging time, so that you can than just read that when asking for version?

Requires no manual interaction - you just tag and it will automatically release (with some help of `.travis-ci.yml`)

jessetan:
Using `setuptools_scm` could make releases easier. Do you know if it supports releasing dev versions to pypi?

Letme:
I added a simple pull request to demonstrate the usage. Yes, it supports releasing dev versions to pypi - thats what travis will handle for us.

agjohnson:
We have more than the python package version number to change, so I lean towards hard coding the version number with the tool here. In readthedocs.org repository, we read this from setup.cfg, but I think this PR is probably a better solution. I'm looking to get away from solutions that need to execute python.

agjohnson:
Also, I'm considering our other repositories and a more generally adopted pattern. Having a general tool like bump2version will get us more flexibility to fit the pattern in our other repositories

Letme:
The write_to option takes also write_template as option for non txt or py files. Anyway, just decide and lets go.

agjohnson:
I made two small changes here:

* Changed the versioning guidelines for development releases to match our other librarys. We don't use the `.dev0` postfix, we use `alpha1` and `rc1` for other packages. I was have trouble getting bump2version to support both. I think `alpha1` works in the same way `.dev0` would however.
* Moved the config to setup.cfg with the rest of the tool configurations. I see some benefits to both, and we can always move back if we hit issues.

PR #826, opened 2019-10-08
Small fixes to setup.py

davidfischer:
Ensure `setup.py build` and `setup.py test` work on a clean checkout.


setup.py
```diff
@@ -19,6 +19,7 @@ class WebpackBuildCommand(setuptools.command.build_py.build_py):
 
     def run(self):
         if not 'CI' in os.environ:
```
davidfischer: If we actually run `npm install` as it does below, perhaps we could actually run this in CI which seems somewhat useful.

PR #827, opened 2019-10-09
Commit package lock file

davidfischer:
This PR is set to merge into `agj/more-webpack-fixes` which is for https://github.com/readthedocs/sphinx_rtd_theme/pull/808.

It commits the `package-lock.json` file with specific pinned versions of node modules.


jessetan:
LGTM, this fixes my FontAwesome import issue.

PR #833, opened 2019-10-16
Add automatic deployment to pypi when tag is pushed

Letme:
`setuptools_scm` works that it provides a version even when you build it
locally and it generates the version into a file at the time of
packaging, which means there is no run-time-dependenty on the
`setuptools_scm` package later on (and also no dependency on git, or
internet access). This makes a simplified release flow, because it
enables you to just add a tag and CI pushes the end package to the PyPi
for end users to use.

Option is also to provide development packages, etc. (just the matter of
how far we want to take it). There is no required tag name, although we
should follow some of the standard formats of Major.minor.patch or
whatever it is preferred.

Before merging we should fill in the secure password (hashed with original repo) and username of the maintainer for pypi. This is added as allow edits from maintainers so that they can add all the details before merging.

closes #822 and #812 if we decide to go this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,24 @@ install:
   - pip install sphinx
 script:
   - tox
-  
+
 branches:
   only:
   - master
+
+jobs:
+  include:
+    - stage: deploy
+      python: 3.6
+      env:
+        - TOXENV=py36-sphinx18
+      deploy:
+        # production pypi
+        - provider: pypi
+          distributions: sdist bdist_wheel
+          user: deply_username
+          password:
+            secure:
+          on:
+            branch: master
+            tags: true
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,6 @@ import setuptools.command.build_py
 from io import open
 from setuptools import setup
 
-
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -13,7 +13,24 @@ install:
...
+        - TOXENV=py36-sphinx18
+      deploy:
+        # production pypi
+        - provider: pypi
+          distributions: sdist bdist_wheel
+          user: deply_username
+          password:
+            secure:
+          on:
+            branch: master
```
agjohnson: What is the purpose of `branch: master` here? We don't want to release to PyPI on master merges, only on new tags.

Letme: Its just an example - we need to fill in password and username as well for this to work.

Letme:
OK, I will prepare travis only change for releasing on tag. I am against effort of keeping bumpversion (commited) and tags in sync, but that's just lazy me.

.travis.yml
agjohnson: I can help with the pypi credentials before merge.

I had to look this one up. `branch: master` here defines _which branch to run the command on_. It does not specify that pushes to master trigger a new release.

PR #835, opened 2019-10-18
Fix always showing dead scrollbar in flyout

agjohnson:
Stop always showing a scrollbar on the flyout menu, though preserve
flyout menu scrolling if the content doesn't fit in the viewport.


agjohnson:
I still need to test badge only on this. I'll set WIP until i do.

Blendify (changes requested):
This appears to mess up the padding on the badge only, the "Read the docs" text and the version on the right do not line up

PR #836, opened 2019-10-18
style: merge some sass theme sources

1138-4EB:
On the one hand, the comment about `bower` in `theme.sass` is outdated. On the other hand, there are four files which are 5-50 lines each (~125 all together). These are merged into `theme.sass` and `_theme_layout.sass`.

* theme_breadcrumbs into theme_layout
* theme_mathjax into theme
* font_awesome_compatibility into theme
* font_local into theme


agjohnson:
Having separation between these files is fine I think, I don't think there is strong reason to remove this separation right now. At very least, I'd rather that `theme.sass` does not have anything but import statements, rules should be in separate files from `theme.sass`.

PR #838, opened 2019-10-19
Address Sphinx 2 compatibility issues

### LINKED ISSUE #741: Footnotes not rendering like the demo in Sphinx 2.0

### LINKED ISSUE #746: Classifier delimiter is not shown with Sphinx >=2.0.0
### Problem

https://github.com/sphinx-doc/sphinx/pull/5976

Sphinx >=2.0.0 defaults to the HTML5 writer.

### Reproducible Project

https://gitlab.com/polyzen/wiki

#### Error Results

With 2.0.1:
https://5cae6f482f618700081c1c0e--polywiki.netlify.com/self-hosted_applications/

#### Expected Results

With 1.8.5:
https://5cae7189c3de3d0009d505a0--polywiki.netlify.com/self-hosted_applications/

### Environment Info

- Python Version: 3.7
- Sphinx Version: 2.0.1
- RTD Theme Version: 0.4.3

### LINKED ISSUE #798: Python domain param display differs in Sphinx 2.1
In Sphinx 1.8.5:
image

In Sphinx 2.1:
image

Affects param, type, return, etc:
image

### LINKED ISSUE #799: List item spacing different in Sphinx 2.1
In Sphinx 1.8, things look good:
image

List item spacing in 2.1 is seemingly haphazard:
image

### LINKED ISSUE #800: Definition list display different in Sphinx 2.1
In Sphinx 1.8:
image

In Sphinx 2.1, the domain styling leaks through:
image

Also affects field lists in the same way:
image

### LINKED ISSUE #801: Option lists display different in Sphinx 2.1
In Sphinx 1.8, a table is used:
image

In Sphinx 2.1, a definition list is used and it looks way better:
image

Switching to the HTML5 writer on all Sphinx versions would allow us to drop the ugly table layout.

### LINKED ISSUE #802: Tables have extra padding with Sphinx 2.1
Sphinx 1.8:
image

Sphinx 2.1:
image

### LINKED ISSUE #803: Sidebars have extra spacing on the bottom margin in Sphinx 2.1
Sphinx 1.8:
image

Sphinx 2.1:
image

### LINKED ISSUE #831: Possible path error?
Not sure if this is intentional or not but in this line:
https://github.com/readthedocs/sphinx_rtd_theme/blob/ddf840cb7206d7c45270560f077b12daa147f915/docs/conf.py#L51
`html_theme_path` points to a directory outside the repo. So on my system when I build I get warnings like
```
WARNING: file 'imlv.zip' on theme path is not a valid zipfile or contains no theme
```
where `imlv.zip` just happens to be a zip file with the same parent directory as `sphinx_rtd_theme`.

### LINKED ISSUE #855: Field list display different in Sphinx 2
Sphinx 1:
Screenshot 2019-11-06 at 15 30 45

Sphinx 2:
Screenshot 2019-11-06 at 15 30 40

This is *not* fixed with #838

agjohnson:
The main change here is that we load information into the template so
that CSS can make decisions about what element hierarchy it can expect,
instead of covering both html4 and html5 writer with a single set of
rules. I also fixed the tox config so that it generates docs for
testing.

This is also in part going to revert the changes in #785. Instead of
maintaining a list of `dl` classes to apply the domain styles to, this
PR instead maintains a list of definition list classes to *not* apply
the domain definition list styling too. There are too many possibilities
to cover with an include list.

Fixes #800
Fixes #801 
Fixes #803
Fixes #831
Fixes #802  
Fixes #798 
Fixes #799 
Fixes #741 
Fixes #746 
Fixes #855 

### Note for issues that will be closed by this PR

I will be cutting a development release for the theme as soon as this PR is merged. We could use some testing at that point for the release `0.5rc1`, which will be available on PyPI (hopefully) early next week.


sphinx_rtd_theme/layout.html
```diff
@@ -7,10 +7,11 @@
   {%- set titlesuffix = "" %}
 {%- endif %}
 {%- set lang_attr = 'en' if language == None else (language | replace('_', '-')) %}
+{%- set sphinx_writer = 'writer-html5' if html5_doctype else 'writer-html4' %}
```
jessetan: An alternative would be to use a `sphinx2` class. If we ever have similar styling incompatibilities with e.g. Sphinx 2.6 or 3.0, they will still use the HTML5 writer and we will need a different class to differentiate.

agjohnson: True, I could see the writer output changing in the future and we might need more granularity eventually. In this case, you _can_ still use the html 4 writer with Sphinx 2, so the most truthy check would be the writer class (`html5_doctype` seemed to be the closest option already in the context data passed by sphinx).

When we cross it, I could see us needing logic like that also uses the sphinx version (assuming there is some major backwards incompatible change at some point)

agjohnson:
Added a method to also test the badge only css and I found a maybe sphinx 2 related bug. I added a switcher debug link:

image

tox.ini
```diff
@@ -15,18 +15,17 @@ deps =
...
-    py.test {posargs}
-
-[testenv:docs]
-changedir = {toxinidir}/docs
-deps =
-    -r{toxinidir}/docs/requirements.txt
-commands =
-    sphinx-build -b html -d {envtmpdir}/doctrees .  {envtmpdir}/html
+    py.test {posargs} tests/
+    sphinx-build -b html -d {envtmpdir}/doctrees docs/ {envtmpdir}/html
```
jessetan: Building the docs with all these versions of Sphinx fails the tests. It looks like this theme has not been able to build its own documentation with Sphinx < 1.5 for some time.

I can get some tests working by changing the first line of `setup()` in `sphinx_rtd_theme/__init__.py` to:
```py
if sphinx.version_info >= (1, 6, 0):
        app.add_html_theme('sphinx_rtd_theme', path.abspath(path.dirname(__file__)))
```

After that change, the sphinx13 and sphinx14 tests fail with
```
sphinx_rtd_theme/.tox/py27-sphinx13/lib/python2.7/site-packages/sphinx/locale/__init__.py", line 195, in _
    return translators['sphinx'].ugettext(message)
KeyError: 'sphinx'
```
 I suggest to revert these lines and separately address compatibility with Sphinx 1.3 and 1.4 (or remove them from the test), so this PR can land.

Blendify:
@agjohnson can you please update this patch so the tests pass?

davidelbaze:
up ? ^^

tox.ini
agjohnson: We've discussed removing support for old versions of Sphinx, there likely is not a lot of benefit to supporting these old versions. Even 1.5 is 4 years old at this point.

agjohnson:
I am currently finishing up this PR and I will be breaking out some of the unrelated work into separate PRs.

I am resolving the remaining issue with field lists, definition lists, option lists, and footnotes with CSS `grid`. This is a fairly well supported option and it works more consistently than flex box.

agjohnson:
Rebased and should be ready for final review! I've enabled PR builder, though the assets look cached at the moment. PR builds should use the locally available theme for testing, though we should probably be using `?{{ commit }}` or something similar to avoid caching the static assets through each build.

src/sass/_theme_rst.sass
```diff
@@ -239,22 +242,63 @@
...
+        vertical-align: top
+  // HTML5 writer
+  html.writer-html5 &
+    dl dt span.classifier
+      &:before
+        content: " : "
+    dl.footnote,
+    dl.field-list
+      display: grid
+      grid-template-columns: max-content auto
```
davidfischer, line 269:
```suggestion
      grid-template-columns: max-content auto
      > dt
        padding-left: 1rem
```

There is a small amount of left padding on the field lists in the HTML4 writer table output.

agjohnson: Oh yeah. I left this one because i think it looks better without the left padding. Any thoughts on which to use?

agjohnson: We can come back to this later, I'll just match the output.

agjohnson:
Thanks for review everyone! I'll move on to cutting a release candidate this afternoon.

PR #843, opened 2019-10-22
Admonitions: Improve Color Vierity

Blendify:
Before some Admonitions shared colors this makes each type more unique.

### DIFF
--- a/src/sass/_theme_rst.sass
+++ b/src/sass/_theme_rst.sass
@@ -128,6 +128,16 @@
     @extend .wy-alert.wy-alert-danger
   .warning, .caution, .attention, .admonition-todo
     @extend .wy-alert.wy-alert-warning
+  
+  .seealso .admonition-title
+    background-color: #7a87e6
+	.seealso
+    background-color: #e7ebfa
+	.important .admonition-title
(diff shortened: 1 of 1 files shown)


agjohnson:
Can you post some screenshots? It would probably be best to have these colors defined as variables somewhere.

eine:
> It would probably be best to have these colors defined as variables somewhere.

IMHO, this is a separate feature, which (if applied) should include not only the two types enhanced here.

src/sass/_theme_rst.sass
```diff
@@ -128,6 +128,16 @@
     @extend .wy-alert.wy-alert-danger
   .warning, .caution, .attention, .admonition-todo
     @extend .wy-alert.wy-alert-warning
+  
+  .seealso .admonition-title
+    background-color: #7a87e6
+	.seealso
```
eine: Tabs and spaces are mixed. Better be consistent.

agjohnson:
> IMHO, this is a separate feature

I'm not talking about a new feature. These should be sass variables similar to the rest of color usage in the sass styles.

Blendify:
If I get time I will update this a soon as I can.

PR #845, opened 2019-10-22
Don't allow select on code block shell prefixes

### LINKED ISSUE #787: `code:: console` is illegible
Using Sphinx 2.x and the following:

```rst
.. code:: console

    /bin/sh -c 'echo hello'

```

The output doesn't meet a11y guidelines around type contrast

image

Blendify:


src/sass/_theme_rst.sass
```diff
@@ -94,6 +94,11 @@
     font-family: $code-font-family
     font-size: 12px
     line-height: 1.4
+  
+  // Do not allow selection of code block prompts
+  div.highlight .gp 
+    user-select: none;
+    pointer-events: none;
```
agjohnson: This won't compile I believe, due to the semicolons. Is `.gp` class the leading `#` and `%`?

jessetan: Good point, I was confused by this. The semicolons have a red background, probably because GitHub sees that it is invalid syntax in SASS (would be valid for CSS).
I mistook the red background for a git diff with characters that were removed.

Please remove the semicolons.

PR #847, opened 2019-10-24
Defer loading JS

jacobtomlinson:
We use this theme as the base for our custom Sphinx theme in the Dask project. We've noticed some render flicker when we load the page and form looking at the profiler this seems to be due to JavaScript resources being fetched and executed before the style sheets.

This PR adds the defer attribute to the script resources which will fetch them asynchronously and execute them once the rest of the page has finished loading. Alternatively the script tags could be moved to below other DOM resources.

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -30,11 +30,11 @@
 
   {# JAVASCRIPTS #}
   {%- block scripts %}
-  <script type="text/javascript" src="{{ pathto('_static/js/modernizr.min.js', 1) }}"></script>
+  <script type="text/javascript" src="{{ pathto('_static/js/modernizr.min.js', 1) }}" defer></script>
   {%- if not embedded %}
   {# XXX Sphinx 1.8.0 made this an external js-file, quick fix until we refactor the template to inherert more blocks directly from sphinx #}
     {% if sphinx_version >= "1.8.0" %}
-      <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}"></script>
+      <script type="text/javascript" id="documentation_options" data-url_root="{{ pathto('', 1) }}" src="{{ pathto('_static/documentation_options.js', 1) }}" defer></script>
       {%- for scriptfile in script_files %}
         {{ js_tag(scriptfile) }}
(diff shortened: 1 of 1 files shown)


jessetan:
Is your issue solved by #756?

jacobtomlinson:
That would help with the symptoms, but doesn't resolve the underlying issue.

Removing modernizr will make this issue less noticeable, but it will still exist on poor network connections and if child themes decide to include other large JavaScript resources through the `script_files` template variable.

Loading and executing JavaScript in the head, especially before CSS, is going to give users a worse experience generally.

agjohnson:
I'll mark as needing some replication. I know this is an issue already however it might help to test this locally, with a RTD instance, where we might be introducing JS that could be incompatible with this change.

jacobtomlinson:
Yeah the only issue I can think of would be introducing other JS which depends on a script that has been deferred. It should also be deferred to ensure the correct loading order.

Let me know if there is anything I can do to help.

jessetan:
Unfortunately I don't think we can add `defer` to everything, some JS is not suitable for deferred loading.

The theme code can be deferredm but that bundles jQuery and we have had issues before when we moved jQuery out of the `<head>`, see #328. We had discussions on performance in #545.

Instead I think we should match what Sphinx does; when custom JS is added, it supports setting attributes (like `defer`) that end up in the HTML.
Docs: https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file
Code: https://github.com/sphinx-doc/sphinx/blob/76e9f57c2e4457dec11d258435b7a6c6ec5de005/sphinx/themes/basic/layout.html#L92

Together with #756, that could solve most of the scripts in head. If not, then we may need a "put theme code + jQuery in head or not" config switch.

jacobtomlinson:
Yeah this was the issue I was talking about, if inline JavaScript depends on a library like jQuery then it can't be deferred like this.

Given the discussions in #545, which seem to roughly be "putting the JS in the head will be slow but more stable", it would be good to allow finer control with a config switch. That was a user can sacrifice the ability to use jQuery inline to get good performance.

rdb:
The flicker issue seems mostly solved in my experiments in Firefox by just moving the CSS link tags above the JavaScript link tags.

jacobtomlinson:
Replaced by #853

PR #848, opened 2019-10-25
Remove Modernizr, keep html5shiv (Webpack edition)

### LINKED ISSUE #724: Screen flashes on load

jessetan:
Modernizr has been part of the theme for a long time, although it is not used. Since its presence has downsides (reflow on load, increased theme size), we want to remove it.
To keep backward compatibility with older IEs, I've kept html5shiv in this PR so there is less chance of breakage.

This is a rework of #756 after our move to Webpack and friends. Rebasing did not make much sense.
Fixes #724, #525. Related #414. Supersedes #624 and #756


sphinx_rtd_theme/layout.html
```diff
@@ -9,8 +9,7 @@
 {%- set lang_attr = 'en' if language == None else (language | replace('_', '-')) %}
 
 <!DOCTYPE html>
-<!--[if IE 8]><html class="no-js lt-ie9" lang="{{ lang_attr }}" > <![endif]-->
-<!--[if gt IE 8]><!--> <html class="no-js" lang="{{ lang_attr }}" > <!--<![endif]-->
+<html lang="{{ lang_attr }}" >
```
agjohnson: Was modernizer also responsible for the `no-js` class here? I was considering using that in the future for no-JS compatibility on some of the CSS, but that functionality would be easy enough to replicate if so.

jessetan: I think the `no-js` was there by default and then removed by Modernizr.
We can do something similar if we want to support users without JS.

jessetan:
Any particular IE versions you want to see? Microsoft no longer offers free screenshots using Browserstack, that requires a paid account.
They do offer VMs with IE8 and up.

agjohnson:
I tried saucelabs, but they dropped IE8 support. IE9 looked as expected though. I tried a demo account on browserstack, but couldn't test IE8. I even tried ie8 through wine to no avail.

I guess I'm not sure how thorough we need to be here then, testing sounds annoying. IE8 only accounts for 0.1% of RTD traffic, which also isn't 100% for projects using this theme.

jessetan:
Perhaps the best way to get feedback on this is to release a dev version of the theme and to wait for user responses or bug reports.

agjohnson:
Sounds good.

PR #853, opened 2019-11-05
Fix flickering issues by changing order of meta tags

rdb:
I noticed that our own documentation, and many other instances of Sphinx documentation using this theme, all exhibited a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) when switching pages as the CSS was still being loaded.  I opened the HTML files and noticed that the JS files were being loaded first.  I'm not sure why; isn't it more important to load the CSS first?

This PR does that by moving the CSS a bit further up so that it can be loaded before the JavaScript needs to run.  It fixes the flickering issues I have observed in Firefox.

Also see #847.


agjohnson:
I'm going to mark this as needing replication. I'm +1 on the fix if someone can verify bumping the CSS up in the head does in fact help with the fouc. I'm focusing on html5 writer compatibility currently, but will have some time to review when we have an RC ready.

OskarPersson:
Doesn't #848 solve your problem?

rdb:
Not quite, because that's not the only bit of JS that is being loaded (I'm adding additional JS files as well in my `conf.py`).  It general, it seems clear to me that the CSS should be loaded before the JS.

Blendify:
I can confirm the flickering issue

jacobtomlinson:
Any chance of this getting merged?

rryan:
Can also confirm the FOUC issue on Firefox on this project that was built today: https://mixxx.org/manual/latest/

jacobtomlinson:
@agjohnson as multiple people have now confirmed the issue could this please be reviewed again?

MatinF:
I can confirm this issue as well

Issue #739, opened 2019-03-28
Sphinx deprecation warning


PhilippSelenium:
### Problem

sphinx_rtd_theme/search.html:20: RemovedInSphinx30Warning: To modify script_files in the theme is deprecated. Please insert a <script> tag directly in your theme instead

### Reproducible Project

Sphinx: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/builders/html.py#L108
RDT: https://github.com/rtfd/sphinx_rtd_theme/blob/b82993dfdb5559b40a0ab4839015bc6c093b6ddf/sphinx_rtd_theme/search.html#L20

```rst
Place RST code here (if any).
```

#### Expected Results

No deprecation warning

### Environment Info

- Python Version: Python 3.6.7
- Sphinx Version: Sphinx==2.0.0
- RTD Theme Version: sphinx-rtd-theme==0.4.3

mik-laj:
Fix is available: https://github.com/rtfd/sphinx_rtd_theme/pull/728

Blendify:
Fixed

peterjc:
This has been fixed with the merge of #728 (April 2019), but there has not been a release to PyPI since - currently at sphinx-rtd-theme 0.4.3 (Feb 2019) which has the ``RemovedInSphinx30Warning`` message.

This breaks trying to run ``sphinx-build -W ...`` to catch regressions in documentation, so a point release would be most welcome.

alanbchristie:
Wow - yes - any chance someone can release a `0.4.4`? It has been 6 months now.

jack1142:
Weird, it doesn't break with `sphinx-build -W ...` for me 🤔

peterjc:
You're right about  ``sphinx-build -W ...`` not aborting due to  ``RemovedInSphinx30Warning`` (I had a another warning further up the log which I had overlooked), sorry.

So the warning does seem to be cosmetic only, and not in itself enough to justify a point release.

alexsuter:
I would also appreciate to get a new release. Because I get a warning in my build and I would like to set the quality gate of my build to zero warnings.

haschdl:
I just updated to Sphinx 2.2 and I was disappointed to get the same Warning, then happy to see it's been fixed, then back to disappointed to see no Pip release. Anyway to facilitate a new pip release?

vaneseltine:
[...]

PR #856, opened 2019-11-11
Documentation for installation case from repo at specific commit

epogrebnyak:
As discussed in https://github.com/readthedocs/sphinx_rtd_theme/issues/739#issuecomment-552283163


agjohnson:
Core team has decided to not support this installation method. It won't work for much longer as our build assets aren't checked into the repository. There is some discussion at #379

PR #860, opened 2019-12-19
Remove Extra Padding on Highlighted search results

Blendify:
There are some cases where the highlighted CSS class creates overlapping text, for example on [Rendering » Shader Nodes » Texture](https://docs.blender.org/manual/en/dev/render/shader_nodes/textures/index.html?highlight=texture) when highlighting "texture". 

This issue is exacerbated by some custom styling rules that we have set, however, this issue is still visible in the stock theme.


src/sass/_theme_rst.sass
```diff
@@ -223,9 +223,10 @@
   // Sphinx can highlight searched text with ?highlighted=searchterm
   .highlighted
     background: $highlight-color
-    display: inline-block
+    box-shadow: 0 0 0 3px $highlight-color
+    display: inline
     font-weight: bold
-    padding: 0 $base-line-height / 4
+    padding: initial
```
jessetan: Is this needed? I see no difference if I remove the line altogether, but there may be cases I did not check.

src/sass/_theme_rst.sass
```diff
@@ -223,9 +223,10 @@
   // Sphinx can highlight searched text with ?highlighted=searchterm
   .highlighted
     background: $highlight-color
-    display: inline-block
+    box-shadow: 0 0 0 3px $highlight-color
```
jessetan: If you change 3px into 2px, the example you gave will not have the highlights stick together.

PR #862, opened 2020-01-03
sass: merge theme sources

eine:
Ref #836

On the one hand, the comment about `bower` in `theme.sass` is outdated. On the other hand, there are four files which are 5-50 lines each (~125 all together). These are merged.

* theme_breadcrumbs into theme_layout
* theme_mathjax into theme
* font_awesome_compatibility and font_local into fonts


src/sass/theme.sass
```diff
@@ -40,11 +27,17 @@
...
-@import theme_font_local
+@import theme_fonts
+
+/* MATHJAX */
+
+span[id*='MathJax-Span']
+  color: $mathjax-color
+
+.math
+  text-align: center
```
Blendify: I think this should be moved to https://github.com/readthedocs/sphinx_rtd_theme/blob/master/src/sass/_theme_rst.sass

eine:
*ping* @agjohnson

agjohnson:
Going to block this until we have html5 writer changes figured out.

PR #863, opened 2020-01-03
html: move parts to head.html, sidenav.html and breadcrumbs_aside.html

eine:
In order to make it easier to find the content, in this PR some parts of `layout.html` are split to new files `head.html` and `sidenav.html`. At the same time, `versions.html` is merged into `sidenav.html`.


eine:
@Blendify, I tried, but `test_builders.py` seems to choke on it: https://travis-ci.org/readthedocs/sphinx_rtd_theme/jobs/632443541

sphinx_rtd_theme/body.html
```diff
@@ -0,0 +1,64 @@
+{% block extrabody %} {% endblock %}
```
Blendify: Rename to body.html

sphinx_rtd_theme/layout.html
```diff
@@ -105,136 +105,6 @@
 </head>
 
 <body class="wy-body-for-nav">
```
Blendify: I think this should be moved to the body include and inside a `{% block body %}`

Blendify:
Please also update the tests

sphinx_rtd_theme/body.html
eine: I don't understand what do you want to rename.

sphinx_rtd_theme/layout.html
eine: A block with name body exists already.

Blendify:
Will try to look into the tests

eine:
> Will try to look into the tests

I'd be glad to be given any hint. I had a look at it, but tests feel fragile and I was not able to run tox in a container..

Blendify:
I would compile documentation with and without this patch and then diff two
html files to see if there is any difference

agjohnson:
It would be best to raise issues for large changes to discuss the changes first.

The largest issue with moving these pieces around is that a number of themes extend from our theme and changes to how the files are structured can break downstream usage. Moving files around is something it makes sense for us to be careful with.

So I'd probably be -1 on moving files around, I see more potential problems from the change than benefits.

eine:
@agjohnson, the purpose of this PR is precisely to make it easier for downstream themes to extend/enhance some parts of it. Having everything together makes it difficult to keep rebasing on top of master, because potentially all the significant modifications to the layout do produce conflicts. Very specifically, ad-hoc snippets which are exclusive to projects hosted on RTD are present in the middle of the layout. After this PR, all those snippets are located at the bottom, which makes it easier to modify. See `body.html` and `sidenav.html`.

agjohnson:
Overall, I agree that having content split out would be easier, if we had started there. Unfortunately, we can't easily move these blocks around for two reasons:

It would also potentially cause problems for anyone that has a `body.html` or `sidenav.html` as this will affect include resolution order of our own templates -- we'd load their body/sidenav first.

The larger issue I'm concerned about is that anyone with a template `layout.html` which was overriding a block or a nested block that has now been moved to `body.html` will abruptly have their content changed back to our our default content. The block now exists in `body.html` and would need to be overridden with a matching `body.html` extended from our `body.html` by anything downstream.

In terms of the use case you're describing, users don't need to rebase on top of master in order to alter our templates. To customize a template, users only need to include a `_templates/layout.html` which looks like:

```
{%- extends "!layout.html" %}

{% block footer %}
   ...
{% endblock %}
```

The Sphinx docs touch on this in their templating guide. We should perhaps have docs on this as well, or at least point to the Sphinx docs.

eine:
> It would potentially cause problems for anyone that has a `body.html` or `sidenav.html` as this will affect include resolution order of our own templates -- we'd load their body/sidenav first.
> 
> The larger issue I'm concerned about is that anyone with a template `layout.html` which was overriding a block or a nested block that has now been moved to `body.html` will abruptly have their content changed back to our our default content. The block now exists in `body.html` and would need to be overridden with a matching `body.html` extended from our `body.html` by anything downstream.

Yes, this would be a breaking change, and should be released as such.

> In terms of the use case you're describing, users don't need to rebase on top of master in order to alter our templates. To customize a template, users only need to include a `_templates/layout.html` which looks like:
> 
> ```
> {%- extends "!layout.html" %}
> 
> {% block footer %}
>    ...
> {% endblock %}
> ```

I think that this is limited to the granularity defined in the template. Which might be too coarse for multiple tweaks. For example, `sidebartitle` includes the logo, title, version and search box.

agjohnson:
> Yes, this would be a breaking change, and should be released as such.

Backwards incompatible changes are something that we only do when necessary. To me, this PR doesn't describe enough potential benefit to warrant the breaking change. The upside might be to improve extensibility of the theme, however causing downstream breakage is going to be an overall negative experience to our users/customers, no matter how we describe backwards incompatibility.

> I think that this is limited to the granularity defined in the template. Which might be too coarse for multiple tweaks. For example, sidebartitle includes the logo, title, version and search box.

There might be a more specific issue in here. However, it's still normal usage to override nested blocks using the standard Sphinx templating flow, no hard fork required. The only case I see for forking the theme is if you want to generate style/color changes in our base stylesheets.

However, I think what you have highlighted is a few things that the theme is not communicating well:

* How you should expect to extend the theme. The answer here is to use local `_templates` overrides or create a theme based on our theme
* What blocks we have documented and how there are used
* How to override individual blocks

I think documenting these would have provide much more value to users then code reorganization and wouldn't induce backwards incompatible changes. Overall, the theme is lacking a lot of documentation it should have, so any additional documentation would likely be a larger net benefit.

I'm sorry to close this PR, however I don't see this one getting merged due to the potential problems for downstream users/customers.

PR #864, opened 2020-01-03
package.json: remove option '--open' from script 'dev'

eine:
When `yarn dev` is executed inside a docker container, `--open` requires a browser to be installed and an X11 server to be shared. Instead, it is more *idiomatic* to just bind a port and remove the option.

Developers that want to execute the script on the host and have the browser opened automatically, can still do so: `yarn dev --open`.

In fact, the command inside the container needs to be `yarn dev --host 0.0.0.0`. Therefore, an alternative solution would be to keep script `dev` and add another one (say `dev-docker`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "0.4.3",
   "private": true,
   "scripts": {
-    "dev": "webpack-dev-server --open --config webpack.dev.js",
+    "dev": "webpack-dev-server --config webpack.dev.js",
     "build": "webpack --config webpack.prod.js",
     "preinstall": "bin/preinstall.js"
   },


davidfischer:
I'm fine with this solution overall. I don't love that the browser opens to an error screen before the docs are built. I'm open to any better solution.

agjohnson:
What's the use case for doing development without access to X11 or a browser?

eine:
@agjohnson, this request is precisely to allow development with access to X11 or a browser that are not installed in the same environment/context/box/container/VM as node/yarn. As commented above, I run `yarn dev --host 0.0.0.0` inside a docker container and just use the browser on the host.

The current command (with `--open`) requires me to install an additional browser and to share and X11 server with the container. However, that should not be a requirement.

eine:
*ping* @agjohnson

agjohnson:
I'm fine with dropping --open. We won't official support the workflow you are describing, but I don't think the open flag was important either way. I have a PR that makes more changes to the development server in #908, which addresses some of the issues with the webpack build flow, though the open call still seems to have sometimes before test docs are built.

PR #865, opened 2020-01-03
update and move 'Previous' and 'Next' buttons

eine:
In this PR, 'Previous' and 'Next' buttons' style is updated and they are moved to better fit the layout (header and footer). See comparison below:

prevnext

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -68,15 +68,19 @@
     {% endblock %}
   </ul>
 
-  {% if (theme_prev_next_buttons_location == 'top' or theme_prev_next_buttons_location == 'both') and (next or prev) %}
-  <div class="rst-breadcrumbs-buttons" role="navigation" aria-label="breadcrumb navigation">
-      {% if next %}
-        <a href="{{ next.link|e }}" class="btn btn-neutral float-right" title="{{ next.title|striptags|e }}" accesskey="n">{{ _('Next') }} <span class="fa fa-arrow-circle-right"></span></a>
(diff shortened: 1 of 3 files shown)


jdillard:
I agree there is room for improvement, but I think the footer change is too invasive and degrades usability by burying the buttons.

I think showing the titles of the previous and next page makes users more likely to traverse forwards or backwards by giving them a preview of the content:
image

Like shown here: https://docs.netgate.com/pfsense/en/latest/install/installing-pfsense.html

eine:
@jdillard, I agree with both points. On the one hand, I had not considered massive custom footers, such as the example you linked. On the other hand, adding the titles makes a better use of the additional space. I think that taking two full rows might be unnecessary, but that can be easily fixed by changing the size of the text.

Hence, I'd be glad to see this PR taken over or superseeded by another one. @jdillard, are you up to cherry-picking the relevant modifications from netgate's docs?

jdillard:
> I think that taking two full rows might be unnecessary, but that can be easily fixed by changing the size of the text.

On longer titles it takes up more than two rows when the title wraps. This leads to needing to catch the edge case on smaller screens of making sure the previous and next titles are clearly separated. Personally, I don't think cons of the added vertical real estate outweighs the pros added usability.

> Hence, I'd be glad to see this PR taken over or superseeded by another one. @jdillard, are you up to cherry-picking the relevant modifications from netgate's docs?

I started working on it and have a local build going (first contribution). I hope to have something public to share soon.

jdillard:
Ok, I made a new pull request at #875, because that's what I am more familiar with. Feedback is most welcome.

PR #867, opened 2020-01-14
Update JS dependencies

davidfischer:
This should fix the github JS vulns warning. I tested this by running `npm run dev` and `npm run build` to ensure that the JS still runs and compiles CSS/JS.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -630,13 +630,10 @@
       }
     },
     "arr-diff": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-2.0.0.tgz",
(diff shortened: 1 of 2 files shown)


eine:
@davidfischer, it seems that `matchdep` is not used (imported/required) anywhere. Is it an actual dependency?

davidfischer:
I didn't actually check whether it was needed. I'll check if it is and remove it otherwise. Thanks for the tip.

davidfischer:
It is not used. I removed it!

eine:
Good! BTW, It would be handy to use david-dm and/or dependabot to keep dependencies up to date: [devDependencies Status](https://david-dm.org/readthedocs/sphinx_rtd_theme?type=dev)

PR #874, opened 2020-01-30
Mark the extension as safe for parallel reading and writing

p12tic:
This allows projects that use this theme to enable parallel building of documentation.

See http://www.sphinx-doc.org/en/stable/extdev/#extension-metadata for more details.

Tested on the documentation of the BuildBot project which produces 10MB of identical HTML files in both parallel and sequencial sphinx modes.


Blendify:
This is for sphinx extensions and not themes, the theme does not register a sphinx extension so I do not think this is required?

p12tic:
We do register the theme as an extension. From the README.rst:

```
...
extensions = [
    ...
    "sphinx_rtd_theme",
]
...
```

When I build the documentation using `sphinx -j<number_of_jobs>` it breaks pointing out that the theme does not mark itself as safe for parallel processing.

p12tic:
```
the sphinx_rtd_theme extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
```

This is the error coming from `sphinx-build`.

Blendify:
Ah, it seems this changed when translations were added. My bad!

Issue #816, opened 2019-09-10
A button for removing `highlight` content from the page


dojutsu-user:
It would be nice to have a button or something else which removes the highlighted content from the page.

![63571884-243d9b80-c59f-11e9-9fbf-b0f34f6533c8](https://user-images.githubusercontent.com/29149191/64588323-bcb98580-d3bf-11e9-9f95-97dde25f0bc3.gif)
_(link: http://2.python-requests.org/en/master/user/quickstart/?highlight=payload#passing-parameters-in-urls)_

Ref: https://github.com/readthedocs/readthedocs.org/pull/6087

agjohnson:
I missed the underlying change here. I'm -1 on this change and -1 on the underlying removal of the search highlighting. The link in this case is not easy to find for users, it is superfluous for most users, and there is better UX to display that we are showing search results. Additionally, search results should be able to highlight across themes in similar ways, so this is an RTD problem.

We need to revisit the search highlighting in RTD, there are better ways to accomplish whatever problem we were hitting with search. Going to close this as I don't think this is a good answer and would be a theme dependent fix to a platform problem.

mgeier:
See #876 for a possible implementation.

PR #876, opened 2020-02-03
Proof of Concept: Hide search matches

mgeier:
Unfortunately, the highlighting of search results has been removed in https://github.com/readthedocs/readthedocs.org/pull/6087.

I don't like this, and I think the real problem is that the RTD theme doesn't have a simple way to remove those highlights, like other Sphinx themes have, see #816.

This PR shows a *really simple* way to add this feature back to the theme.

Once we have this, we can revert the removal of the highlighting, see https://github.com/readthedocs/readthedocs.org/issues/6305.

I'm not interested in the styling (can somebody else please do that?), I just want to demonstrate a *very simple* way to get this functionality.

The relevant CSS selectors would probably be `#searchbox .highlight-link` for the `<p>` and `#searchbox a` (or `#searchbox p a`?) for the link itself.

I think it makes sense to add the "Hide Search Matches" link close to the search text field, but if desired, it could be placed anywhere be simply adding a `<div id="searchbox"></div>`. That's all there's to it.

Please note that this does *not* remove `?highlight=...` from the URL. Such a feature, if desired, should be implemented in Sphinx itself.

What do y'all think about this?

### DIFF
--- a/sphinx_rtd_theme/searchbox.html
+++ b/sphinx_rtd_theme/searchbox.html
@@ -1,5 +1,5 @@
 {%- if builder != 'singlehtml' %}
-<div role="search">
+<div id="searchbox" role="search">
   <form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get">
     <input type="text" name="q" placeholder="{{ _('Search docs') }}" />
     <input type="hidden" name="check_keywords" value="yes" />


agjohnson:
Any reason to continue with this PR with search highlighting back on RTD?

mgeier:
Well yes. Highlighting is back (which is great!), but there is still no button or whatever to hide the current highlighting in the RTD theme, right?

But this PR is just a proof of concept, feel free to create a nicer styling.

YaZko:
I am not sure how up to date the current PR is, but I would be very interested in the highlighting being fixed indeed. 
More specifically, it currently highlights the searched word in all subsequently visited pages. 
As far as I'm concerned, I would prefer highlighting to be altogether disabled: I don't quite ever foresee a use case where I would want the technical term I looked for to find a session be bright yellow for its 30 or so occurrences in the text I'm gonna read. 
But if people like it, I would at least appreciate a light way to disable it (ideally, hitting escape for instance) other than manually editing the URL to remove the option.

In particular, @cpitclaudel mentioned on Zulip that it could be easily disabled by editing the following CSS:
```
.rst-content .highlighted {
    background: #F1C40F;
    display: inline-block;
    font-weight: bold;
    padding: 0 6px;
}
```

marxin:
I would like to support the pull request.
The feature is really missing as one can't easily (URL needs to be adjusted) hide the highlighted text.

mgeier:
> Please note that this does _not_ remove `?highlight=...` from the URL. Such a feature, if desired, should be implemented in Sphinx itself.

I've just suggested such a change in https://github.com/sphinx-doc/sphinx/pull/9551.

ana-borges:
I would also like to support this PR. The highlighting is extremely annoying for me.

mgeier:
@YaZko 

> But if people like it, I would at least appreciate a light way to disable it (ideally, hitting escape for instance)

In the meantime, this has been implemented in Sphinx: https://github.com/sphinx-doc/sphinx/pull/9337

PR #896, opened 2020-04-06
Templates: Add block for footer content info

Blendify:
Allows users to expand this area to add there own content


sphinx_rtd_theme/footer.html
```diff
@@ -49,8 +50,8 @@
     {% set readthedocs_web = '<a href="https://readthedocs.org">Read the Docs</a>'  %}
       {% trans sphinx_web=sphinx_web, readthedocs_web=readthedocs_web %}Built with {{ sphinx_web }} using a{% endtrans %} <a href="https://github.com/rtfd/sphinx_rtd_theme">{% trans %}theme{% endtrans %}</a> {% trans %}provided by {{ readthedocs_web }}{% endtrans %}.
   {%- endif %}
+  {%- endblock %}
```
agjohnson, line 53: This endblock should be moved. The start begins right after the div start, but the div ends on line 46. Anyone using this block would override only the closing tag of the div

stsewd (approved):

Blendify:
Looking back at this, I think the blocks should encompass the div so users can completely

PR #900, opened 2020-04-15
Release 0.4.4

stephenfin:
Do the necessary work to release 0.4.4 and get things working with Sphinx 3.x. After this is merged, I think someone will just need to to tag the relevant commit (`HEAD~`) and push the release to PyPI.

@ericholscher I _assume_ this falls on you to do? /o\ Let me know if there's anything else I can do to help.

**EDIT** It might look like there's more going on here than is necessary, but I've broken things up with commits, as seen [here](https://github.com/readthedocs/sphinx_rtd_theme/pull/900/commits).


setup.py
```diff
@@ -85,7 +85,7 @@ def run(self):
 
 setup(
     name='sphinx_rtd_theme',
-    version='0.4.3.dev0',
+    version='0.4.5.dev0',
```
jessetan, line 88: This seems like a typo (also in `__init__.py`)

stephenfin: Yeah, I've done this weirdly but I think it's correct. This is set to `0.4.4` (here and in `__init__.py`) in [`HEAD~`](https://github.com/readthedocs/sphinx_rtd_theme/pull/900/commits/a701997e75c8b53868584d0807dc61541db3451c) and changed to `0.4.5.dev0` in [`HEAD`](https://github.com/readthedocs/sphinx_rtd_theme/pull/900/commits/f8cc1843c504b7555d2bcade19af3ecf25f88780)

stephenfin: Let me know if you want me to pull out the post-version bump patch. If not, you just need to make sure you take `HEAD~`, *not* `HEAD` :smile:

jessetan: Ah, now I understand what you did. This should also work, provided there are no additional commits needed to make 0.4.4 work.

Since there may be someone else working on this, I think it might be safer to keep this on `0.4.4`, go through all release shennanigans, then ciommit a version bump. How does that sound?

ericholscher:
@agjohnson is the main point person on the theme.

setup.py
stephenfin: Yup, good call (y) Thinking about it, this wouldn't work if you ended up with a merge commit either. Will drop that shortly

docs/changelog.rst
```diff
@@ -11,7 +11,22 @@ New Features
...
+:Date: Apr 15, 2020
+
+New Features
+-------------
+
+Fixes
+-----
+
 * Fix definition lists looking different with Sphinx 2.0+
+* Add compatibility with Sphinx 3.x
```
polyzen, line 29: Where was Sphinx 3.x compatibility added?

stephenfin: Commit a49a812c8821123091166fae1897d702cdc2d627

stephenfin:
I wonder if I should roll in dropping support for Python 2.7 and 3.4 here? I'm thinking it would be better get this out the door and cut a 1.0.0 release that does that at a later date, but I'm willing to be overruled :smile:

docs/changelog.rst
polyzen: 3.0.0 was released a year later. That was for Sphinx 2.x compatibility:
https://github.com/sphinx-doc/sphinx/issues/6088

polyzen: Sphinx 2.x compatibility is still ongoing:
https://github.com/readthedocs/sphinx_rtd_theme/projects/1

jessetan: "Improve compatibility with Sphinx 2.x" is more appropriate

stephenfin:
Okay, it seems this is clearly not ready so it's probably best I close it. Feel free to cherry-pick any of the commits here that might be useful.

utzig:
Why was this abandoned?

PR #901, opened 2020-04-16
add view github link option

derekflint:
This commit contains a small change to allow you to set a link to a github repo to view.

html_context = {
    "display_github_url": 'https://github.com/<org>/<repo_name>',
}

Navigation would show "View on Github" with link.


septatrix:
I think this would be even better if it were vendor agnostic or added for all three vendors for which we currently have the "Edit on Vendor" option. Maybe call the option `display_repo_url` and make it a tuple of vendor name and url or just add it for all three vendors for now

jessetan:
This is already possible by setting the following in `conf.py`:
```py
html_context = { "display_github":True, "github_url": "MY URL" }
```
Please reopen if you feel this PR adds additional value.

derekflint:
@jessetan the behavior you described above is to show 'Edit on Github' **not 'View on Github'**.  

I also tried your changes to config.py

`html_context = { "display_github":True, "github_url": "https://github.com/augerai/a2ml" }`

clicking the link described in your example takes a user to 'https://github.com///blob/index.rst.

It doesn't appear to use my github_url value.

jessetan:
Sorry, there are additional fields `github_repo`, `github_user`, `github_version`.

The `github_url` can also be set as file-wide metadata in rst: `:github_url: https://github.com/augerai/a2ml`

PR #906, opened 2020-04-26
Upgrade node-sass to avoid security vulnerabilty

mondeja:
See [this](https://www.npmjs.com/advisories/961) for more info.


agjohnson:
Thanks for the update!

PR #907, opened 2020-05-01
Added Estonian translation

ookull:


agjohnson:
Thanks. Could you bring these translations into Transifex instead? We manage all of our translations there, not via submitted gettext.

https://sphinx-rtd-theme.readthedocs.io/en/latest/contributing.html#translations

agjohnson:
I've uploaded the strings to transifex. Review will happen on our project there:
https://www.transifex.com/readthedocs/sphinx-rtd-theme/language/et/

Thanks for the updates

Issue #764, opened 2019-05-23
Add the option for a right sidebar


choldgraf:
This is a suggestion for a new feature...I'd be happy to give a shot at implementing it if folks are interested.

I've found that, with a wide-enough screen, it's helpful to have a sidebar to the right with within-page links. Currently, the RTD theme includes in-page links within the TOC to the left.

In some of my projects, I've got a very small amount of CSS that will pop out the within-page links to the right of the page when the screen is wide-enough. It works something like this:

![2019-05-23_11-59-05](https://user-images.githubusercontent.com/1839645/58279070-44687f00-7d52-11e9-8266-4e66ddd848be.gif)

I'm wondering if this is something that would be useful as an addition to this theme...maybe something that could be activated w/ a particular configuration? My goal thus far has been to make this possible without modifying any of the HTML - it is a CSS-only modification, something like this:

```
@media screen and (min-width: 1270px)  {
    div.wy-side-scroll li.toctree-l1.current > ul {
        position: fixed;
        margin-left: 0px;
        overflow-y: auto;
        padding-left: 5px;
        border-left: 1px solid #767677;
        height: 95%;
        top: 70px;
        left: 1100px;
    }

    div.wy-side-scroll li.toctree-l1.current a {
        border-right: none;
    }

    div.wy-side-scroll li.toctree-l1.current ul {
        max-width: 400px;
    }

    div.wy-side-scroll li.toctree-l1.current ul, li.toctree-l1.current li, li.toctree-l1.current a {
        background-color: transparent !important;
    }

    /* Handle the visibility of right sidebar */
    div.wy-side-scroll li.toctree-l1.current ul {
      display: block !important;
    }

    div.wy-side-scroll li.toctree-l1.current span.toctree-expand {
        display: none !important;
    }

    /* Make the padding correct */
    div.wy-side-scroll li.toctree-l2 li.toctree-l3 > a {
      padding: .4045em 4.045em !important;
    }

    div.wy-side-scroll li.toctree-l4 > a {
      padding: .4045em 5.663em !important;
  }
}
```

There are some hacks in there, but I was impressed we could get such a big shift out of a relatively small amount of CSS.

Let me know if folks would be interested in this and I'm happy to discuss changes that would need to be made etc.

agjohnson:
[...]

PR #919, opened 2020-05-07
Add local TOC as right sidebar for #764

jdillard:
This PR adds a table of contents for the local headings on the page (excluding the page title) as a right sidebar, as mentioned in #764. This implementation is similar to the one in [GitLab's Handbook](https://about.gitlab.com/handbook/) in that is moves the local TOC to the top of the page on smaller screens.

Demo:

local-toc-sphinx-rtd-theme

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -193,6 +193,12 @@
           <div role="main" class="document" itemscope="itemscope" itemtype="http://schema.org/Article">
           {%- block document %}
            <div itemprop="articleBody">
+            {% if toc|length > title|length + 75 %}
+            <nav id="local-table-of-contents" role="navigation" aria-labelledby="local-table-of-contents-title">
+              <h4 id="local-table-of-contents-title">On This Page</h4>
+              {{ toc }}
+            </nav>
+            {% endif %}
             {% block body %}{% endblock %}
            </div>
            {% if self.comments()|trim %}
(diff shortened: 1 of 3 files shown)


agjohnson:
Thanks for giving a try on this! I tried this same thing a while back and it looked promising, but did introduce some UX issues with the existing global toc. We'll want to address these sort of issues and put this all under a configuration setting, but I think calling `{{ toc }}` directly is the way to go with a secondary side nav. I'll also have some feedback on css styles, i think we can reduce that a good deal and make the side nav more subtle.

To start, can you run `python setup.py build` to regenerate the built CSS so we can review the changes? This PR is viewable at https://sphinx-rtd-theme--919.org.readthedocs.build/en/919/

jdillard:
> Thanks for giving a try on this! I tried this same thing a while back and it looked promising, but did introduce some UX issues with the existing global toc. We'll want to address these sort of issues and put this all under a configuration setting, but I think calling {{ toc }} directly is the way to go with a secondary side nav.

No problem! Yea, having the ability to turn the local TOC off, at a minimum, would be nice.

>  I'll also have some feedback on css styles, i think we can reduce that a good deal and make the side nav more subtle.

For sure, CSS isn't exactly my strong suit so there are probably some leftovers from all the trial and error.

> To start, can you run python setup.py build to regenerate the built CSS so we can review the changes?

This has been done now and pushed.

Blendify (changes requested):
We will want this to be a theme option, to start out at least.
When the option is enabled I think `theme_includehidden` should be forcible disabled.
This should solve the UX of using both the global toc and local toc together.

The css side seems okay, there should be a way to collapse the local toc when the page is narrow.

net-sniper:
Great functionality, but I don't know how to configure it?

PR #945, opened 2020-06-09
Add a bottom margin for anything that does not match p, ul, ol

### LINKED ISSUE #943: Bottom margin of code block in bullet list is too small (regression from 0.5.0rc1)
### Problem

Bottom margin of code block in ~~bullet~~ (apparently number list is affected as well, [link](https://jack1142-red-prs.readthedocs.io/en/v3-bump_deps_html5_writer/framework_config.html#tutorial)) list is too small in version 0.5.0rc2 (on both html writers)

### Reproducible Project

Links below.

#### Error Logs/Results

Version 0.5.0rc2 (on both html writers):
image
[Link (html4 writer)](https://jack1142-red-prs.readthedocs.io/en/v3-bump_deps/install_windows.html#installing-red)
[Link (html5 writer)](https://jack1142-red-prs.readthedocs.io/en/v3-bump_deps_html5_writer/install_windows.html#installing-red)

#### Expected Results

How it looked before (0.5.0rc1 and older):
image
[Link](https://docs.discord.red/en/latest/install_windows.html#installing-red) (this one is on 0.4.3 as I don't have 0.5.0rc1 link handy but I did test locally that it looked the same on 0.5.0rc1)

### Environment Info

- Python Version: 3.8.3
- Sphinx Version: 3.0.3
- RTD Theme Version: 0.5.0rc2

agjohnson:
All other elements should maybe have a bottom margin, but paragraphs and
nested lists should not have a bottom margin.

Fixes #943


agjohnson:
From the PR docs @ https://sphinx-rtd-theme--945.org.readthedocs.build/en/945/demo/lists_tables.html#bullet-lists:
image

Bullet list spacing is preserved, code/admonition elements have bottom margin

agjohnson:
I've updated this PR to more closely match docutils intended difference between `ul` and `ul.simple`: https://docutils.sourceforge.io/sandbox/html4strict/data/simple-lists.html

0.5.0rc2 was moving towards treating everything as a `ul.simple` list. This PR now styles both list styles independently, so complex lists have more spacing, but simple lists still look nice and compact.

0.4.3 has bugs that display simple lists with spacing on the first 2 levels. 0.5.0 will differ as it both resolves the bugs in 0.4.3 that cause simple lists to have unnecessary margins, and complex lists will more consistently have the correct spacing on list items.

The one place that I don't agree with docutils, or docutils doesn't specify what the output should be perhaps, is when a simple list is nested inside a complex list. The simple list uses different margins from the parent list and looks inconsistent. I'll leave this as-is for now as I assume this is intended

Example:
image

agjohnson:
Also worth mentioning that docutils suggests the following spacing:

image

This looks considerably worse and I guess I'm probably -1 against this. I don't think most authors know about the difference between a simple and complex list and would definitely hit this on accident.

agjohnson:
And finally, I added extra margin, but only for list items with multiple paragraphs. I think this is a good compromise for now. Example here: https://github.com/readthedocs/sphinx_rtd_theme/issues/943#issuecomment-645040623

PR #968, opened 2020-06-30
pl locale

artur-trzesiok:


agjohnson:
Thanks! Unfortunately, we manage translations at Transifex, not Github. I'll try to import these translations soon, but if I don't get to it I believe you can also join as a contributor and import the file. I'm happy to make you a contributor or reviewer there if you'd like

PR #994, opened 2020-09-04
Ignore prev/next nav buttons using meta fields

jdillard:
I've found sometimes the content that follows or precedes in the toctree isn't directly relevant and the nav buttons need to be ignored or sometimes you want to not show any nav for specific page for a landing page type feel. This may not be the best approach, but it can at least serve as a proof of concept.

To use, add the `:ignorenav:` meta field to the top of the desired page(s). Accepted values are `prev`, `next`, and `both`. `prev` will override/ignore the previous nav button, same applies to `next`, and `both` will ignore both buttons.

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -180,6 +180,11 @@ The following options can be used as :ref:`file-wide metadata
 
     Force the :guilabel:`Edit on GitLab` button to use the configured URL.
 
+.. confval:: ignorenav
+
+    Acceptable values are ``prev``, ``next``, and ``both``. ``prev`` will
+    override/ignore the previous nav buttons, similar applies to ``next``, and
+    ``both`` will ignore both previous and next buttons.
 
 Other configuration
 ===================
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -70,12 +70,15 @@
     {% endblock %}
   </ul>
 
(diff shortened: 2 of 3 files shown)


Blendify:
I think such a feature belongs in sphinx itself and not the theme. Sphinx should add a meta element that let's you specify what next page should be. That should eliminate the need to disable the buttons.

Blendify:
Closing in favor of #995.

jdillard:
I don't get how the two are related? Would specifying a blank value not show anything? I didn't see how you account for not showing a prev/next butting in https://github.com/sphinx-doc/sphinx/issues/9024

Blendify:
The original issue only referenced overriding the link so not sure if this is a feature anyone is asking for. The feature request  in sphinx could of course be updated with this feature though.

jdillard:
I guess I'll update the feature request on sphinx. As stated in the original issue this PR wasn't really relevant, I just edited the comment vs deleting it (perhaps I should have deleted it to sever the tie).

Sometimes you have content that is part of a larger TOC, where the previous and next pages aren't directly relevant and it makes since to not have nav options. I can't imagine this is a unique use case. This happens with the product manuals here, where the next product manual isn't really relevant, but it makes sense to have all the product manuals in a single TOC tree: https://docs.netgate.com/pfsense/en/latest/solutions/netgate-2100/index.html

Blendify:
That's understandable, i just want to minimize the amount of features we want to support here when they can be integrated into sphinx and used by any theme.

Issue #973, opened 2020-07-07
At high zoom, hidden page elements still capture keyboard focus


jonels-msft:
### Problem

When browser zoom is high enough that the table of contents disappears, the menu items are still active and trap the tab key. This leads to the effect of tab selection disappearing offscreen.

### Reproducible Project

https://pg-auto-failover.readthedocs.io/en/latest/

#### Error Logs/Results

https://www.youtube.com/watch?v=noJQCQq5P80

#### Expected Results

When the menu is hidden on zoom, its elements should lose their tab indices. Tabbing should select items visible on screen, and never leave the visible page items for invisible menu items.

Additionally, the hamburger icon (☰) should be selectable via keyboard to show/hide the menu.

### Environment Info

- Python Version: 3.7.3
- Sphinx Version: 1.8.5
- RTD Theme Version: 0.4.3

PR #1009, opened 2020-10-08
When menu is hidden, their elements can not be focusable

mondeja:
As was requested in #973, if the menu is not displayed in the screen, their focusable elements (links in toctree, logo link and search bar) should not be focusable using TAB key.


jonels-msft:
@Blendify or someone else, can you please review this PR?

Issue #970, opened 2020-07-06
Search input boxes lack labels


jonels-msft:
### Problem

For accessibility, form inputs should have corresponding labels. The search box above the table of contents, and the search box inside the version popup menu aren't associated with labels.

### Reproducible Project

https://pg-auto-failover.readthedocs.io

#### Error Logs/Results

![label-1](https://user-images.githubusercontent.com/48225118/86639358-5fba3c00-bf9e-11ea-987a-26f629f2f3c2.JPG)

![label-2](https://user-images.githubusercontent.com/48225118/86639368-634dc300-bf9e-11ea-8e09-bebbeaca6a6f.JPG)

#### Expected Results

Fix any of the following:
* aria-label attribute does not exist or is empty
* aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
* Form element does not have an implicit (wrapped) `<label>`
* Form element does not have an explicit `<label>`
* Element has no title attribute or the title attribute is empty

See [WCAG 3.3.2](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)

### Environment Info

- Python Version: 3.7.3
- Sphinx Version: 1.8.5
- RTD Theme Version: 0.4.3

PR #1016, opened 2020-10-25
Label tag to conform with WAVE accessibility

eglerean:
Adding a <label> tag to conform with https://wave.webaim.org/ WAVE accessibility tool. For more details on this webaim requirement, see https://webaim.org/standards/wcag/checklist#sc2.4.6

This pull requests is a possible solution to issue #970


stsewd:
I'm just attaching a screenshot of how this looks like.

Screenshot_2020-10-26 Read the Docs Sphinx Theme — Read the Docs Sphinx Theme 0 5 0 documentation

Holzhaus:
Is this to improve screenreader support? If so, could that label just be hidden using CSS?

stsewd:
I think is better if we copy the upstream changes https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/themes/basic/searchbox.html looks like they have aria labels for search already. Let me know if that works

stsewd:
Created https://github.com/readthedocs/sphinx_rtd_theme/pull/1024

PR #1022, opened 2020-12-01
HTML Search: Use sphinx's built in search template file

Blendify:
Instead of using our own template uses sphinx's basic search.html instead.
This change is almost identical except for the new search page includes 
a search box. As a result, the search box is hidden on the search page.

image


stsewd (commented):
Looks good. Just one thing, we are missing the `display_vcs_links`  var that is used here

https://github.com/readthedocs/sphinx_rtd_theme/blob/38752d455ff0206678e7d9ac591c4237f8af6226/sphinx_rtd_theme/breadcrumbs.html#L27

I think we can just check if `pagename` is different from search here
https://github.com/readthedocs/sphinx_rtd_theme/blob/38752d455ff0206678e7d9ac591c4237f8af6226/sphinx_rtd_theme/breadcrumbs.html#L41

PR #1055, opened 2021-02-23
Migrate to bootstrap (WIP)

Blendify:
This patch makes the initial changes needed to move from wyrm to bootstrap.

Some things that need to be done:

[] use bootstrap layout system
[] fix fonts
[] fix some styling 
[] fix breadcrumbs aside

### DIFF
--- a/src/sass/_theme_breadcrumbs.sass
+++ b/src/sass/_theme_breadcrumbs.sass
@@ -16,16 +16,9 @@
     background: none
     &.literal
       color: $text-color
-.wy-breadcrumbs-extra
-  margin-bottom: 0
-  color: $text-light
-  font-size: 80%
-  display: inline-block
 
 
 +media($mobile)
-  .wy-breadcrumbs-extra
-    display: none
   .wy-breadcrumbs li.wy-breadcrumbs-aside
     display: none
 
--- a/src/sass/_theme_breadcrumbs.sass
+++ b/src/sass/_theme_breadcrumbs.sass
@@ -2,7 +2,7 @@
    padding-left: 0px
    background-color: inherit
 
-+media($mobile)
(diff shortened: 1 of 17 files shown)


stsewd (commented):
I think @agjohnson was discussing migrating to SUI

Blendify:
I am not familiar with SUI can you provide a link? Note a didn't spend
too much time on it, this is just a proof of concept. I wanted to make a
bootstrap theme because it would be easy for 3rd parties to extend.
For example, https://github.com/executablebooks/sphinx-panels is a great
extension built on bootstrap for adding bootstrap derivatives to sphinx.

stsewd:
@Blendify this is it https://semantic-ui.com/

Blendify:
It would be nice to hear the pros of SUI. Bootstrap is a lot more popular but it seems that semantic-ui might be more customizable

polyzen:
Semantic UI hasn't seen commits for 2 years: https://github.com/Semantic-Org/Semantic-UI/branches

polyzen:
https://bulma.io/alternative-to-bootstrap/

stsewd:
@polyzen I reply it to that in https://github.com/readthedocs/sphinx_rtd_theme/issues/544#issuecomment-784549548, let's keep that conversation there, so this PR isn't overwhelmed with that.

src/sass/_theme_breadcrumbs.sass
```diff
@@ -1,34 +1,14 @@
-.wy-breadcrumbs
-  +clearfix
+.breadcrumb
+   padding-left: 0px
+   background-color: inherit
```
Blendify, line 3: Set `$breadcrumb-bg` here instread.

PR #1059, opened 2021-02-26
Cleanup: Move Breadcrumbs to layout and clean up unused rules

Blendify:
I moved breadcrumbs to the rest of the layout as it seems to belong 
better here.
I also went through and removed some unused rules.

These changes can be split up into two patchs if needed.


src/sass/_theme_breadcrumbs.sass
```diff
@@ -1,34 +0,0 @@
...
-  color: $text-light
-  font-size: 80%
-  display: inline-block
-
-
-+media($mobile)
-  .wy-breadcrumbs-extra
-    display: none
-  .wy-breadcrumbs li.wy-breadcrumbs-aside
-    display: none
```
Blendify, line 30: This code was also removed, the reset was just moved.

src/sass/_theme_breadcrumbs.sass
```diff
@@ -1,34 +0,0 @@
...
-    padding: 5px
-    border: none
-    background: none
-    &.literal
-      color: $text-color
-.wy-breadcrumbs-extra
-  margin-bottom: 0
-  color: $text-light
-  font-size: 80%
-  display: inline-block
```
Blendify, line 23: This code was also removed, the rest of this file was only moved.

benjaoming:
This would have to be updated since we have new SASS for breadcrumbs since https://github.com/readthedocs/sphinx_rtd_theme/pull/1073. I did kind of like this separate file, since https://github.com/readthedocs/sphinx_rtd_theme/pull/1073 was one of the few PRs that didn't have a merge conflict to resolve in `_theme_layout.sass`

PR #1060, opened 2021-02-26
Templates: Cleanup Whitespace

Blendify:
While not perfect this cuts it down a lot and makes the html output much cleaner

This saved over 10mb in a project with 1500 HTML files  translated into a dozen languages


stsewd (approved):
> This saved over 10mb in a project with 1500 HTML files translated into a dozen languages

:exploding_head:

PR #1066, opened 2021-02-28
NPM: Update to version 7

Blendify:
NPM made some changes to package.json and package-lock. From what I can 
tell these changes are backward compatible, see 
https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json

I also moved dependencies that affect the build package to regular 
dependencies instead of devDependencies. As far as I know this is 
correct usage and allows developers to easily see what affets the built 
backage and what affects the devOps process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,22 +7,24 @@
     "build": "webpack --config webpack.prod.js",
     "preinstall": "bin/preinstall.js"
   },
-  "dependencies": {},
-  "devDependencies": {
+  "dependencies": {
     "bourbon": "~4.3",
     "bourbon-neat": "~1.9",
+    "font-awesome": "^4.7.0",
+    "html5shiv": "^3.7.3",
+    "jquery": "^3.4.1",
+    "lato-font": "^3.0.0",
+    "roboto-fontface": "^0.10.0",
+    "wyrm": "^1.0.9"
+  },
+  "devDependencies": {
     "copy-webpack-plugin": "^5.1.1",
     "css-loader": "^3.1.0",
     "file-loader": "^3.0.1",
-    "font-awesome": "^4.7.0",
-    "html5shiv": "^3.7.3",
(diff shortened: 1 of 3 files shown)


stsewd (approved):
great! we may hold to merge this as well for a bigger release

Blendify:
This is not a user affecting change so it shouldn't matter if it's a big or small release. I do want to break this into two patches though...

package.json
```diff
@@ -7,30 +7,31 @@
...
-    "css-loader": "^3.1.0",
-    "file-loader": "^3.0.1",
     "font-awesome": "^4.7.0",
     "html5shiv": "^3.7.3",
-    "imports-loader": "^0.8.0",
     "jquery": "^3.4.1",
     "lato-font": "^3.0.0",
+    "roboto-fontface": "^0.10.0",
+    "wyrm": "^1.0.9"
+  },
```
agjohnson, line 19: Raised in the other PR, I don't think this is necessarily correct. In the case that we wanted to make this available as an NPM/etc package, we shouldn't require users to install these dependencies when they wouldn't be used for anything outside the build process.

agjohnson (commented):
This looks entangled with removing Wyrm, so would have to be on a later release. The other PR that touches this file probably is more close to merging, but this one has to be pushed pretty far out.

Blendify:
This is not related to removing wyrm, the splitting of dev requirements are also not necessary, I meant to update this patch.

package.json
```diff
@@ -17,7 +17,6 @@
     "url": "https://github.com/readthedocs/sphinx_rtd_theme/issues"
   },
   "homepage": "https://github.com/readthedocs/sphinx_rtd_theme",
-  "dependencies": {},
```
Blendify, line 20: npm7 doesnt like this being empty.

PR #1068, opened 2021-03-04
Nav: Fix right padding on level2+

Blendify:
The sets the right padding the same as the first level.
The current padding is too large and varies between toc levels.
Now, the right padding is the same between all levels and only the left padding changes

Before:
image

After:
image


Blendify:
Another example: 
image

stsewd (approved):

PR #1070, opened 2021-03-04
Use dart-sass instead of node-sass

Blendify:
Node sass is now deprecated, dart-sass is now the main implementation. 
There is no change to the built assets.

For more information see 
https://sass-lang.com/blog/libsass-is-deprecated


Blendify:
We should wait until we drop bourbon for this, I will make this change in my V2 branch.

PR #1071, opened 2021-03-11
Tests: Remove test for sphinx version no longer supported

Blendify:
We no longer run tests against sphinx 1.4 so this can be removed.


stsewd (approved):
:+1:

PR #1073, opened 2021-03-11
Update Breadcrumb Styles

Blendify:
- Fixes extra padding to the left/right
- Used slash instead of >>
- Make literalls match other link style
- Adds demo docs for testing

Before:

Screenshot from 2021-03-11 14-48-45

After:

image


src/sass/_theme_breadcrumbs.sass
```diff
@@ -1,21 +1,19 @@
...
-    padding: 5px
-    border: none
-    background: none
-    &.literal
-      color: $text-color
+    all: inherit
+    color: inherit
+
+.breadcrumb-item::before
+  content: "/ "
```
nienn, line 15: The look of the new breadcrumb separators resulted a bit too intrusive.
I would maybe shrink them and make them lighter.

```suggestion
  content: "/"
  color: #bbb
  font-size: 13px
  padding: 0 6px 0 3px
```
_The new padding is required to re-center them, after changing the font size._

nienn:
Changes look good!
I don't particular like the new separator style — made a suggestion on that — but this might be more of a personal preference.

Now that we can test a long breadcrumbs path directly on the docs I'm also noticing that the floating right `.wy-breadcrumbs-aside` does not sit well on small screens.
**Before:**
b-before


**After:**
b-after

benjaoming:
I really like the PR and @nienn's improvement! I did a quick verification and ended up with exactly the same results :+1: 

Before @nienn's suggestion:

image

After @nienn's suggestion:

image

docs/demo/level1/level2/index.rst
```diff
@@ -0,0 +1,9 @@
+
+******************************************
+:mod:`Breadcrumb Level 2 <test_py_module>`
```
benjaoming, line 3: Very clever test this!

benjaoming (approved):
This is a really positive improvement and adds a very nice demo with 6 nested sections :+1: 

I don't see any risks of regressions here.

src/sass/_theme_breadcrumbs.sass
olivaja: Hola cómo estás

Issue #938, opened 2020-06-03
Deprecate HTML4 writer


agjohnson:
The 0.5.0 release is dealing with a lot of incompatibility added by Sphinx switching the default html writer to HTML5. While we can continue supporting both the html4 and html5 writer in our stylesheets, we will be duplicating development efforts and QA on any stylesheet changes that touch list styles or other elements that were significantly altered with the switch to the HTML5 writer.

The issue that I'm seeing with supporting the html4 writer is that we'll have spread testing and QA of the theme pretty thin, as we already have to support 3 versions of Sphinx and multiple versions of Python. Adding on top of this testing of both html4 and html5 writer for at least multiple versions of Sphinx feels like a lot of testing for little benefit.

So: what are our feelings on the HTML4 writer? Is it important to support it? This is almost a conversation about which versions of Sphinx we want to support, as I think most users just use the default version of the writer.

I think what I'd propose is:

* Aim to support Sphinx 2.0 and 3.0 with a 1.0 release. Do we not support Sphinx 1.8 at all?
* Start throwing a soft deprecation warning about the html4 writer. Users using Sphinx<2.0 will likely see this warning. The deprecation warning will alert the user to switch to a newer release of Sphinx or the HTML5 writer
* In a following release, 2.0 perhaps, we throw a harder deprecation warning or error if using the HTML4 writer. In this release, I guess we remove the HTML4 writer specific CSS rules.

Daltz333:
👎 to supporting Sphinx 1.8.
👍 for deprecating, I doubt it will effect the vast majority of users. However I believe this should be a harder warning especially if a stable release cycle isn't found (2.0 being in 3 years???)

agjohnson:
version 1.0 and 2.0 would be dependent on backwards incompatible changes to our API, not based on time.

Daltz333:
Ah, I understand. I had doubt based on the development momentum of this project as of recently (seems like there is a lot of "core" internal changes that had to be addressed). I'm fine with deprecating with semver.

PR #1075, opened 2021-03-12
Require Sphinx2

### LINKED ISSUE #679: Fix conditional logic around search include for Sphinx 1.8
The logic introduced in #672 could stand to be a bit more defensive -- a tuple comparison or something similar would be great.

Blendify:
We now depend on the html5 builder and we want to deprecate the html4 writer see: #938. A printed warning message will come in a different PR. Before this can happen, we have to ensure that the html5 writer is available which was added in sphinx2.

This patch also drops python2 and python versions less than 3.5. Sphinx2 requires these so it is safe to remove support.

Fixes #679


stsewd (commented):
We may need to hold on this, there is a discussion about our tool support policy to be done, see https://github.com/readthedocs/readthedocs.org/pull/7859

Blendify:
That should not affect this because readthedocs.org uses the latest version of each, this only affects if a user pins an old version of sphinx against a new version of them

agjohnson:
We do need to hold off on this until we come up with a plan for deprecation and backwards incompatible changes across the next few versions. Going to set this one to block and put it on a future milestone.

In particular, this could break a good number of documentation builds, anywhere where users are using an old Sphinx and an unpinned theme. This will require some more thought and perhaps application logic.

Blendify:
> We do need to hold off on this until we come up with a plan for deprecation and backwards incompatible changes across the next few versions. Going to set this one to block and put it on a future milestone.
> 
> In particular, this could break a good number of documentation builds, anywhere where users are using an old Sphinx and an unpinned theme. This will require some more thought and perhaps application logic.

The theme will now install sphinx2 if sphinx2 is not found, so I do not think that this will break any builds.
I am still undecided as to what version this should go in, this should go in before we remove support for html4 (remove html4 styling) so if version 1.0.0 should remove html4 support then it makes sense to include this change in the next point release. On the other hand, jumping to html5 by default is still a major change a can warrant to be included in 1.0.0.

agjohnson:
This is effectively dropping Python 2.7 support. While not technically wrong, this potentially affects a large number of users. I don't have the number of builds using python2.7 still on RTD, but it was still a large number fairly recently. Anyone not pinning the theme release would end up with broken builds, which we want to avoid.

This PR might require more work, perhaps there is a packaging mechanism we can use to point python2 users to a supported version of the theme and avoid this scenario.

Otherwise, we probably don't want this change until we also take on html4 writer depreciation and bootstrap support in a release. If we merge this early, that will be the final release available to python2 users and any changes in between now and a bootstrap release would be unusable. It would be a better experience for these users to delay this change as long as possible, especially if we're not gaining anything in particular with the upgrade.

We should also release a deprecation warning now on dropping Sphinx <2 support.

PR #1076, opened 2021-03-12
Deprecate html4_writer

### LINKED ISSUE #938: Deprecate HTML4 writer

Blendify:
This does not remove the fixes that we made for the html4_writer, these could be removed in a future release.

Depends on #1075
Fixes #938

### DIFF
--- a/sphinx_rtd_theme/__init__.py
+++ b/sphinx_rtd_theme/__init__.py
@@ -33,6 +33,8 @@ def config_initiated(app, config):
 # See http://www.sphinx-doc.org/en/stable/theming.html#distribute-your-theme-as-a-python-package
 def setup(app):
     app.require_sphinx('2.0')
+    if app.config.html4_writer:
+        logger.warning("'html4_writer' is not compatible with sphinx_rtd_theme")
     # Register the theme that can be referenced without adding a theme path
     app.add_html_theme('sphinx_rtd_theme', path.abspath(path.dirname(__file__)))
 


Blendify:
image

agjohnson:
Also going to block this PR. It might make more sense to put this in a release before removing Sphinx 2 support. Currently, we have not described to users that it's required to upgrade to the HTML5 writer, so jumping right to breaking builds will be a bad experience for users.

Blendify:
Note, this does not break builds using the html4 writer this might not require sphinx2, I am a bit unsure if sphinx will work if we evaluate a config value which is not present.

Blendify:
Closing in favor of 114fb85b441db180af2f53dc0f74f16bf7c72e16

PR #1080, opened 2021-03-12
Fix #1078: Citation not styled properly

### LINKED ISSUE #1078: Citation not styled properly
### Problem

Citation should be styled as same as footnote, but it is not. This is exactly the same issue as described in #1030 and I've confirmed it hasn't been fixed.

In https://sphinx-rtd-theme--1039.org.readthedocs.build/en/1039/demo/demo.html#references, the citation definitions under the `Citation` section are actually footnotes not real citations , that's why you can't see the wrong styling.

Let's see how docutils defines citation:

> Citations are identical to footnotes except that they use only non-numeric labels such as [note] or [GVR2001].

Citation definition produces a `dl` element with `class="citation"` while footnote definition produces a `dl` with `class="footnote"`.

#### Expected Results

Citation should be styled as same as footnote.

### Environment Info

- Python Version: 3.9.1
- Sphinx Version: 3.5.1
- RTD Theme Version: 0.5.1

Blendify:
Turns out our demo docs werent being rendered by sphinx as citations so this may have slipped though the sphinx 2 fixes.

I updated the demo docs and included the need style fixes.


agjohnson:
The citation display looks incorrect in this PR 
image

This should match the surrounding styles, and should have matched the html4 output

Blendify:
That should be how it looks before this PR. with this or I get:

image

agjohnson:
It looks like you didn't build the assets on this PR, so the PR build output looks broken. Can you update this PR with the built assets?

Blendify:
Sure, but we should figure out what cased the assets to no longer be built.

agjohnson:
So, these assets always had to be manually built when opening/changing a PR, which is actually one of the reasons we don't want to check these assets in to the repository. It makes doing PR review tricky, and contributors need to know to build assets on their PRs after each change.

By removing the assets from the repository, and only building them when needed -- on RTD/CI and at release time -- we can keep these files out of PR and review entirely.

benjaoming:
It would seem that in order to get this PR in, we need to merge latest `master` and then rebuild assets. I'll be getting a dev env up and hopefully I can push that easily after that :crossed_fingers:

benjaoming:
Verified that I got the intended output that @Blendify mentioned in https://github.com/readthedocs/sphinx_rtd_theme/pull/1080#issuecomment-801495939

If tests build, I'm going to :ship:

benjaoming (approved):
LGTM! The new changes render well.

image

benjaoming:
(sorry needed another merge for conflicts on theme.css to go away)

@agjohnson should we switch on "Squash and Merge" as a default?

agjohnson:
@benjaoming indeed :+1: Squash and merge should be our default on repositories now

agjohnson (approved):

benjaoming:
Going to track this in #1307 

Btw we now have Squash and Merge as the default :+1:

PR #1082, opened 2021-03-12
Tests: Run against nav toctree and not builtin toctree

Blendify:
Before, the test were checking the `.. toctree::` directive and not the navigation toctree in our template. This is now fixed.

A white space change was made so we dont have to include the white space in the search string.

The test project was update to not collapse the navigation and to use infinite toc depth to match the old test case.


benjaoming (approved):
Great change! LGTM :+1: If tests pass, let's merge :)

benjaoming:
Since the tests don't pass, we would need some more work on this.

PR #1083, opened 2021-03-12
Docs: Always use local theme

Blendify:
This changes non PR builds on rtd to use the local theme and not the old 0.4.3 theme. We want our docs to show what the theme looks like based on the current release, not an old version.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,21 +4,9 @@ import sys
 import os
 import re
 
-# If we are building locally, or the build on Read the Docs looks like a PR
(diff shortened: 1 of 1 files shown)


stsewd (approved):

agjohnson:
It's not clear why this was conditional logic exists. I wanted to verify this before merge.

agjohnson:
I can't dig up any history on this one. The conditional logic was added and removed a several times. I think the conditional was mostly to preserve `master` from breaking docs. We're using `stable` as the default version in the docs, and could consider disabling `latest` entirely, as we have specific versions building anyways.

PR #1084, opened 2021-03-12
Fix definition lists inside api docs

### LINKED ISSUE #1052: definition lists in
### Problem

Definition lists are rendered differently when defined within domain documentation.

May be loosely related to https://github.com/readthedocs/sphinx_rtd_theme/issues/979.

### Reproducible Project

```rst
Foo
   foo
*Bar*
   bar
``Baz``
   baz

.. py:class:: pub

   Foo
      foo
   *Bar*
      bar
   ``Baz``
      baz

   .. py:method:: qux

      qux
```

Screenshot 2021-02-12 at 12 57 15

Strangely the definition lists render differently to lower-level objects documented within the domain, are the two styles mixing? The rendering is especially awkward for literals.

I took a look at [this page](https://sphinx-rtd-theme--981.org.readthedocs.build/en/981/demo/api.html#test_py_module.test.Foo.another_function), however, there aren't any definition lists in that reference example. Perhaps it's an unusual thing to do.

#### Expected Results

I think that definition lists should render the same in both contexts.

### Environment Info

- Python Version: 3.7
- Sphinx Version: 3.4.3, 5.0.1
- RTD Theme Version: 0.5.1, 1.0.0

Blendify:
This is similar to #1080 (and is needed to fix citations in the demo here)

Fixes #1052


agjohnson:
@Blendify can you generate assets here too? The PR build would be the most helpful

ariso353:
@agjohnson and @stsewd any update on the review?

benjaoming:
It would seem that in order to get this PR in, we need to merge latest master and then rebuild assets. I'll be getting a dev env up and hopefully I can push that easily after that :crossed_fingers:

benjaoming:
This seems to render well, great work @Blendify :tada: 

image

benjaoming (approved):
LGTM :+1:

benjaoming:
Ping @agjohnson - I would say this looks good!

Here's another screenshot - the new visual regression test code added by @Blendify 

image

PR #1093, opened 2021-03-17
Add python version requirements to setup.py

Blendify:
This is need in the future so we can gracefully remove python 2.7 support.
See #1075.

The required python versions are based on sphinx 1.6 which is our minimal required version.

https://packaging.python.org/guides/dropping-older-python-versions/


setup.py
```diff
@@ -116,8 +116,9 @@ def run(self):
             'sphinx_rtd_theme = sphinx_rtd_theme',
         ]
     },
+    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
```
agjohnson, line 119: If we need `python_requires` of `>=2.7` here to begin the process of deprecating 2.7, then I think we should isolate that from deprecating other python versions. Currently, it doesn't seem there is a reason for us to drop support for any of the lesser supported python3 minor versions.

agjohnson (commented):
We probably want to specify the 2.7 requirement, if that is an important step in pypi giving out 1.x as the last python2.7 release. The other deprecations seem separate however. Is there any reason we're removing those here?

Blendify:
The python3 versions that were dropped are because sphinx has already dropped support for them when sphinx 1.6 is removed. We require sphinx 1.6 because it is the first version with experimental HTML5 and needed for some other bug fixes in the theme.

setup.py
Blendify: These python3 versions have been dropped since  sphinx 1.5: https://www.sphinx-doc.org/en/master/changes.html#release-1-5-released-dec-5-2016

setup.py
```diff
@@ -106,7 +106,7 @@ def run(self):
         'theme.conf',
         '*.html',
         'static/css/*.css',
-        'static/css/fonts/*.*'
+        'static/css/fonts/*.*',
```
astrojuanlu, line 109: Woops! Was this affecting user installations?

Blendify: It does not seem like it

astrojuanlu (approved):
I'm not an expert in the theme but this pull request doesn't do anything revolutionary. Python 3.0 to 3.2 are universally recognized as Bad™, and 3.3 is just too old at this point. As @Blendify has clarified, these versions haven't been supported since Sphinx 1.5 anyway. There are no outstanding discussions left. In summary: I'm merging it. Thanks @Blendify for your persistence!

PR #1094, opened 2021-03-18
Remove font-smoothing

Blendify:
This is a nonstandard CSS feature and is not supported on many platforms. And has been removed from the CSS standard specification.
It is advised to not use it according to MDN along with many other guides on the internet.

https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth


Blendify:
Some argue that this is still worth it on darker backgrounds but not sure if it is worth it. And could cause more problems then it fixes. If we want to keep this we will need to create a custom mixin if we want to remove bourbon.

nienn:
This looks good! `font-smoothing` should be removed.

As for the dark backgrounds, this argument from [responsivedesign.is/articles/font-smoothing](https://responsivedesign.is/articles/font-smoothing/) makes a good point:

> If you’re design is placing text onto a dark background then you have a good reason to look towards using the WebKit and Firefox prefixed font-smoothing options to make the text look lighter, but you should be warned that these only work on Mac and OSX and leaves the billions of other users with a sub-standard view.
> 
> **What should I do?**
> I recommend not using it, and instead if you’re looking to get a lighter version of the font then use the font-weight to control this so that it appears nicely across both Mac and Windows, and works across all browsers without the need of a browser prefix.

Further reading: [stop-fixing-font-smoothing](https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/) from usabilitypost.com.

benjaoming (approved):
LGTM :+1:

PR #1101, opened 2021-03-25
layout: allow arbitrary CSS attributes

### LINKED ISSUE #1100: arbitrary add_css_file attributes are not supported
### Problem

It is sometimes necessary to add a CSS file with attributes such as `media` (for print, dark mode, etc.). As of today, this is not possible when using `sphinx_rtd_theme` because attributes seem to be ignored (excepting `title`): https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L27

### Reproducible Project

Create a custom CSS file and add it to your project like this:

```python
app.add_css_file("css/my_customization.css", media="print")
```

In the example above, `media` attribute is ignored on the rendered pages.

#### Error Logs/Results

N/A

#### Expected Results

All attributes given to CSS files should be taken into account.

### Environment Info

- Python Version: 3.8.5
- Sphinx Version: 3.5.3
- RTD Theme Version: 0.5.1

gmarull:
Allow arbitrary CSS attributes when using app.add_css_file() on Sphinx.
It can be useful e.g. to insert `media` or any other attribute. Note
that rel and type are set by default.

Fixes #1100


Blendify:
Does feature for other sphinx themes? This is what the basic theme does, does this syntax work?

```
    {%- for css in css_files %}
      {%- if css|attr("filename") %}
    {{ css_tag(css) }}
      {%- else %}
    <link rel="stylesheet" href="{{ pathto(css, 1)|e }}" type="text/css" />
      {%- endif %}
    {%- endfor %}
```

gmarull:
@Blendify thanks for pointing that out! I was not aware of `css_tag`.

Blendify (approved):

Blendify:
The change is good, however it requires sphinx 1.8 but currently we support versions down to 1.6 so this will need to be an if else block see https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L59 or wait until #1075.

If you use the if else block I will add this to the upcoming 1.0 release.

Blendify (changes requested):
Meant to request changes

sphinx_rtd_theme/layout.html
```diff
@@ -26,7 +26,11 @@
     {%- if css|attr("rel") %}
       <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
-      <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+      {%- if sphinx_version >= "1.8.0" %}
+        {{ css_tag(css) }}
+      {%- else %}
+        <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+      {%- endif %}
```
Blendify, line 33: this can be simplified:

```
  <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
  <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
  {%- for css in css_files %}
     {%- if css|attr("filename") %}
        {{ css_tag(css) }}
     {%- else %}
         <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
     {%- endif %}
  {%- endfor %}
```

Blendify (changes requested):

sphinx_rtd_theme/layout.html
```diff
@@ -26,7 +26,11 @@
     {%- if css|attr("rel") %}
       <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
```
Blendify, line 28: The logic here is a bit messed up I'm starting to realize it is a bit more complicated than I originally thought.

```
 <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
  <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
  {%- for css in css_files %}
    {%- if sphinx_version >= "1.8.0" %}
      {%- if css|attr("filename") %}
        {{ css_tag(css) }}
      {%- else %}
         <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
      {%- endif %}
    {%- else %}
      {%- if css|attr("rel") %}
        <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
      {%- else %}
        <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
      {%- endif %}
    {%- endif %}
  {%- endfor %}
```

Blendify (changes requested):

sphinx_rtd_theme/layout.html
gmarull: I see `filename` attribute already present on 1.8.0, so why is the additional check needed?

Blendify: Not sure but this is what sphinx does: https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/themes/basic/layout.html#L97

The case for before 1.8 is to match the template before this commit: https://github.com/sphinx-doc/sphinx/commit/c5d6942b88a5902a843ba41c74cb01c402413e48

gmarull: ok, should be ready now

gmarull: @Blendify ping

Blendify (approved):
Looks good but we are going to hold off until 1.1 release.

Blendify:
It seems to merge is blocked however, maybe because you force pushed a commit

benjaoming:
We still support Sphinx 1.6 in 1.1... but will be dropped in 2.0. See: https://sphinx-rtd-theme.readthedocs.io/en/stable/development.html#release-2-0-0

gmarull:
any chance this can be merged? @Blendify

agjohnson:
Note, this was merged in #1519

PR #1105, opened 2021-03-30
Tests: Fix deprecation warning with new sphinx versions

Blendify:
These builders were split in sphinx and the old import path will be removed in the upcoming Sphinx 4 release.


agjohnson:
This can also be done in a backwards compatible way, either with a try block or inspecting the Sphinx version.

astrojuanlu (approved):
LGTM!

PR #1106, opened 2021-03-30
Cleanup: Remove workaround/compatibility for old 0.3 version

Blendify:
This should now be safe to remove. I tested locally and sticky nav work both on and off.

I am not sure about the rtd side if this cases issues.

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -18,14 +18,6 @@ function ThemeNav () {
     nav.enable = function (withStickyNav) {
         var self = this;
 
-        // TODO this can likely be removed once the theme javascript is broken
-        // out from the RTD assets. This just ensures old projects that are
-        // calling `enable()` get the sticky menu on by default. All other cals
-        // to `enable` should include an argument for enabling the sticky menu.
-        if (typeof(withStickyNav) == 'undefined') {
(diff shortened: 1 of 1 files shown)


agjohnson (commented):
I believe this is okay, but I also feel like I'm overlooking something.

Background on this is that we used to host a common theme.css and theme.js for all projects to share. That file needed to be backwards compatible between HTML output from multiple theme versions because of that. Projects should have individual assets now, but it is also functionality that we have not yet fully deprecated that I know of.

agjohnson:
Changes seem okay technically speaking, but it might be worth double checking on RTD usage first. In theory, old projects might use the common assets still, but we haven't updated the assets in the rtd.org repo in a while. Any changes here wouldn't be reflected in the common assets unless we decided to copy the built assets into rtd.org vendored assets again.

@davidfischer you helped split up common hosted assets back into per-build assets if I recall. Does this change seem relatively safe to you?

I did bump this out to 1.1 or later. I'd like to work towards a 1.0rc1 this week or next, and would like to limit the number of changes we're putting into that release.

benjaoming:
Marking this for 2.0 unless someone does a quick intervention :)

PR #1107, opened 2021-03-30
Remove all instances of READTHEDOCS

### LINKED ISSUE #650: Remove all instances of READTHEDOCS & release 1.0
We should make this theme like any other Sphinx theme when running on RTD. This means removing all the `{% if READTHEDOCS %}` blocks. There are only a couple left, but we should remove them all.

After we do this and have the theme fully abstracted, I think we should release it as a 1.0. I feel silly having it under 1.0, and want to move to real semvar on our projects going forward.

Blendify:
The removes all the if checks for read the docs.

On read the docs builds this content is now injected. However, we want to support non read the docs builds and provide a way to generate the version menu.

This diff includes changes from #578 and #983

Fixes #650

This will also need some changes on RTD side


agjohnson:
This is another design decision to get back to, probably.

Based on our most recent conversations, our badge flyout version selector menu is actually going to move outside the theme. Maintaining that as part of our sphinx theme is a fairly confusing pattern.

The version of the version selector that is built into this theme may or may not be a feature of Read the Docs as well, in which case we don't want to support that for outside projects -- at least not on our theme.

We instead would want to expose version and language selection as integrated features for the theme, not as part of the menu.

eric-wieser:
In a project of my own, I'm trying to use the `READTHEDOCS` var to detect whether the `current_version` is available as is done here:

https://github.com/readthedocs/sphinx_rtd_theme/blob/3532ffc89bbf89ef7d2ac62d65b3683f588a9673/sphinx_rtd_theme/layout.html#L156-L158

Two questions then:

* If `READTHEDOCS` is removed, how can I detect whether RTD metadata (like PR vs non-PR builds) is vailable
* This already seems to have stopped working - where is the `READTHEDOCS` variable populated?

Blendify:
@eric-wieser this task is to remove all the instances where the `READTHEDOCS` var is used in the theme, the var is populated from https://github.com/readthedocs/readthedocs.org/blob/master/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl there are no intentions of removing the var itself, just the instances of it in the theme.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,50 @@
-{% if READTHEDOCS %}
+{%- block versions_menu %}
+{%- if theme_versions_menu|tobool %}
 {# Add rst-badge after rst-versions for small badge style. #}
   <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="{{ _('Versions') }}">
     <span class="rst-current-version" data-toggle="rst-current-version">
-      <span class="fa fa-book"> Read the Docs</span>
```
benjaoming, line 5: @agjohnson @Blendify 

Trying to understand this PR - won't this effectively remove Read the Docs branding from the theme even when published on Read the Docs?

* if this is the case, then I think it's hard to merge the PR for 1.1 without an alternative.
* if this is NOT the case, then I don't see any issues here.

agjohnson: It will be reinjected via our API, but then there will be side effects like something similar to a FOUC, where the Read the Docs branding reappears after the API call finishes.

I still feel like this whole issue requires a bit more thought. Our primary concern is users hosting docs on RTD, so this change is mostly outside our immediate concerns. If any user needs to remove the RTD branding, they can override `versions.html` now.

I'd say this is a good PR to come back to, after more discussion about goals here and goals for the flyout.

PR #1109, opened 2021-04-02
hlist: Make columns collapse when viewport is narrow

### LINKED ISSUE #1108: Use flexible layout for hlist directives
### Problem

The current hlist can look bad on mobile or when using many columns.

### Reproducible Project

https://docs.blender.org/api/current/bpy.types.AlphaUnderSequence.html

#### Error Logs/Results

image

As you can see the list items extend beyond viewport when the viewport is narrow.

#### Expected Results

The columns should break into new rows.

### Environment Info

- Python Version: 3.9.x
- Sphinx Version: 3.5.1
- RTD Theme Version: 0.5.1

Blendify:
When the items of the hlist cannot fit into the viewport a collumn is split into a new row until the content fits.

This dramically improves mobile experience, for example https://docs.blender.org/api/current/bpy.types.AlphaUnderSequence.html will introduce horizonal scrolling on the whole viewport which causes visual glitchs with other content.

Fixes #1108


Blendify:
Before:
image


After:
image

src/sass/_theme_rst.sass
```diff
@@ -253,9 +253,13 @@
...
     font-size: 90%
 
-  // Horizontal lists have multiple columns; use full width
   .hlist
     width: 100%
+    tr
+      display: flex
+      flex-flow: row wrap
+    td
+      margin-right: auto
```
agjohnson, line 262: Currently, this theme does not use flex box layout because we have not officially dropped support for IE11, or IE10 for that matter. There is limited support for flex box in IE10+11, but you will need to add more rules.

This PR does break hlists on IE11:

image

This is what IE11 looks like on latest release, for comparison:

image

agjohnson (changes requested):
This fix seems good on other browsers, but IE11 does have problems with the flex box layout. If we want to merge this before dropping IE11 support, flex box can be used to some degree in IE10+11. So far, IE10+11 support has been the reason we haven't used flex layout more.

Blendify:
I would be in favor of dropping ie 11, the bug you show is avoidable on the user side if they set `figwidth` on the figure directive. The bug you show is also not present on all ie11 using browser stack, on windows 7 and ie 11 it displayed correctly for me. I bet the majority of users that use ie 11 are on windows 7. 

There are also already issues with ie11 with the theme (see the logo) so maybe we go ahead and formally drop compatibility now. Also, the bug you show is similar to #1050 maybe it is acceptable to introduce a minor regression 

Does adding `display: -ms-flexbox` fix the bug? According to the docs I have read `flex-flow` is supported on ie11 without a prefix. If the solution is that simple, I can just add that and we can hold off on dropping ie11 a little longer.

agjohnson:
> Does adding display: -ms-flexbox fix the bug?

I did not test this. In theory it might though, IE11 requires a prefix on flex box rules for what support it did have.

> There are also already issues with ie11 with the theme (see the logo)

This is a minor display issue for now, it probably doesn't warrant dropping ie11 support immediately by itself.

> The bug you show is also not present on all ie11 using browser stack, on windows 7 and ie 11 it displayed correctly for me.

I tested on saucelabs ie11 + win7, against the PR build output.

agjohnson:
If the `-ms` prefixed rules do not solve this bug for IE11, I'd say we should hold off until a release that drops ie11

agjohnson:
Oops, I take that back. It also looks like you didn't rebuild assets yet, only the SASS source was updated

Blendify:
> Most recent update seems to have introduced vertical spacing between elements

I think that is a docutils 0.17 issue, rebased the patch on master and updated assets so hopefully, work as expected.

benjaoming (approved):
According to our own docs, we only _partly_ support IE11 and will remove support in 2.0.

See: https://sphinx-rtd-theme.readthedocs.io/en/stable/development.html#supported-browsers

From this, I would gather that this change is entirely fine for 1.1, since it will not introduce a "major bug".

agjohnson:
It would probably make more sense to revisit in 2.0, when we do drop IE11 support.

PR #1113, opened 2021-04-05
Setup: Require docutils<0.17'

### LINKED ISSUE #1112: broken alignment for captions in toctrees in ver. 0.5.1
### Problem

I have been using the older version `0.4.3` and had a nice toctrees like below:
Screenshot 2021-04-05 at 00 13 22

However, the search functionality was bugged so I updated to `0.5.1` and now I get:
Screenshot 2021-04-05 at 00 13 33

And it seems to me that the captions for both toctree elements is not aligned anymore. Neither horizontally nor vertically (werid space between the caption and the first section, no space between last section of the previous toctree and a new one).

### Reproducible Project

```rst
.. toctree::
   :hidden:
   :maxdepth: 2
   :caption: General Information

   custom/installation
   custom/background
   custom/example-results
   custom/citing-contributing
   custom/changelog
   custom/license

.. toctree::
   :hidden:
   :maxdepth: 2
   :caption: API

   custom/users-api
   custom/internal-api
```

#### Expected Results

I would like to have a page layout just like for `0.4.3`.

### Environment Info

- Python Version: `3.7.4`
- Sphinx Version: `3.5.3`
- RTD Theme Version: `0.5.1`

Blendify:
Docutils 0.17 changes the html output for some markup, until support is added to properly style this new markup we should pin to use an older version.

It may not be a bad idea to keep a pinned version in the theme to prevent this issue in the future.

Fixes #1112

If possible this should be committed into its own 0.5.2 release ASAP!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -117,7 +117,8 @@ setup(
         ]
     },
     install_requires=[
-        'sphinx>=1.6'
+        'sphinx>=1.6',
+        'docutils<0.17', # https://github.com/sphinx-doc/sphinx/issues/9001
     ],
     tests_require=[
         'pytest',


JWCook:
I just discovered this as well (in https://github.com/reclosedev/requests-cache/pull/218). Specifically, I found that the semantic tags output by the new HTML5 writer (https://docutils.sourceforge.io/RELEASE-NOTES.html#id46) break list formatting in sphinx-rtd-theme. For example, `<div class="section">` elements are now `.<section>`

ericholscher:
@Blendify Thanks for jumping on this. I'll try and get it released today. My current plan is to just branch from 0.5.1 and just apply this patch for a 0.5.2. There's some backwards incompatible changes in `master` that I don't want to ship to everyone just for a bugfix. Does that make sense?

ericholscher:
@Blendify Sounds good. I'll get it shipped here soon. I think I can just cherrypick it onto a branch off 0.5.1. 👍

ericholscher:
The release PR is now here: https://github.com/readthedocs/sphinx_rtd_theme/pull/1114

Blendify:
This got merged in a seperate commit see ca2719b32fcae38923d4830963431e54ded69aae

PR #1116, opened 2021-04-05
Use static setuptools metadata

astrojuanlu:
This migrates most of the metadata keys to the new static/declarative `setup.cfg`.

Also, I tried fixing the bumpversion configuration, but when I tried `bumpversion micro` it introduced a few bugs in the `package-lock.json` and whitespace changes in `setup.cfg`, so I am not sure if we are still using it.


agjohnson:
Yeah, I still use bumpversion. I'd like to use it on other projects as well too. I'm going to target a future release for this work, after we get a 1.0 release out with some of the current work pending.

astrojuanlu:
Reported the bumpversion issues here https://github.com/readthedocs/sphinx_rtd_theme/issues/1117

astrojuanlu:
Rebased on top of `master` 👍🏽

benjaoming:
:smile:  @agjohnson - just stumbled upon gold here, the bumpversion issues that got introduced in package-lock.json are already fixed here by @astrojuanlu

benjaoming:
Updated to latest master. I've included the Jinja fix from #1316.

setup.py
```diff
@@ -86,73 +86,11 @@ def run(self):
 
 
 setup(
-    name='sphinx_rtd_theme',
     version='1.0.1alpha1',
```
benjaoming, line 89: Shouldn't we also have the version in setup.cfg?

benjaoming (approved):
This is really nice work :100:  looking forward to getting rid of setup.py all-together.

I'll do a version bump in another PR, so I might move the version to setup.cfg there.

PR #1122, opened 2021-04-09
Update html Layout fixes issues and improves html5 usage

Blendify:
This patch updates the css layout to use new html5 semantic tags. This patch also reworks the sidebar to reduce the needed css/html nesting and makes things simpler. As a result this improves the responsiveness of the sidebar. The sidebar is now always fixed to 300px (this was a bug report will update to add the task).

This patch helps with html5 adoption and makes moving to bootstrap easier. This patch also helps with docutils 0.17 which makes better use of semantic tags.


agjohnson:
This feels like a big backwards incompatible change that is going to conflict with switching to Bootstrap. I'd probably vote we only make backward incompatible changes to this DOM structure once. These are common elements for users and customers to restyle.

The side navigation width issue on tablet viewport size might be better to address after a bootstrap port, if we decide do anything with it once we're using bootstrap.

For some additional context here, tablet viewport usage on our docs is < 0.5% of our monthly pageviews. It might make sense to just have the side menu experience match mobile to give us one less difference to maintain. Agreed 85% or 100% width will be odd on tablet viewport sizes though.

Issue #473, opened 2017-10-05
Fixed top menu bar in mobile


umbe1987:
Hi everybody,
I really like this theme. Would it be possible to always visualize the top menu bar on mobile even after scrolling down the page? I mean, the bar with the title and the three rows which you can tap to visualize the various chapters/sections.
Actually, if one needs it on mobile, and he's down the page very far from the top, he has to scroll all the way to the top to find it.
Is there an option to have it fixed?
Thanks in advance!

umbe1987:
Any thoughts on this?

I am working to improve this, but it's really hard due to my poor knowledge of the code/of programming in general.
I just found out from the developer window that the class controlling the aspect of the nav bar on mobile is called `wy-nav-top`. This has CSS property `position:relative`. Of course, changing this to `position:fixed` makes it always visible on top, but it shows up completely "ugly" (it's small and to the left, see image below).
I would be happy to contribute with a pull request, but I would need some help.

Thanks again!

![immagine](https://user-images.githubusercontent.com/15016826/31766993-cd960d68-b4c9-11e7-8b85-f211948960a9.png)

jessetan:
Just did a quick test of this; would be doable by adding `position:fixed; width:100%; top:0` to `.wy-nav-top` for mobile/tablet breakpoint and setting `margin-top: $gutter / 4 + 50px` on `.wy-nav-content` for same breakpoint.

Opinions on adding this option? @Blendify @davidfischer @agjohnson 
I don't like things eating up my real estate on mobile but if it is a common request, we can do a theme option.

umbe1987:
Thank you so much, I tested your solution and it's exactly what I needed.
I don't think it takes much of the display, it's just a very small portion of it and most importantly, it's a must have if you want to navigate to another chapter easily. I would love to see it implemented, but meanwhile I can live with this implementation.
Again, thanks for solving this, really appreciate it!

umbe1987:
Any news on this? I am actually using it and it's fine. The only real problem is that when I expand the main navigation panel (`wy-nav-side shift`), the page scrolls to the top, and, no matter if I was reading (saying) the middle of he page, as soon as I close it I am not where I left before openin the panel, but at the top of the page.

umbe1987:
[...]

PR #1125, opened 2021-04-11
Header: Hide on scroll

Blendify:
This patch solve the issue where users have to scroll to the top of the page to open the sidebar.

This patch will work best with #1122 but can be independent.

Fixes: #473


src/theme.js
```diff
@@ -2,6 +2,17 @@ var jQuery = (typeof(window) != 'undefined') ? window.jQuery : require('jquery')
 
 // Sphinx theme nav state
 function ThemeNav () {
+    /* When the user scrolls down, hide the navbar. When the user scrolls up, show the navbar */
+    var prevScrollpos = window.pageYOffset;
+    window.onscroll = function() {
+      var currentScrollPos = window.pageYOffset;
+      if ((prevScrollpos > currentScrollPos) || (currentScrollPos < 40)) {
```
Blendify, line 9: Add a 5px delta to make it less sensitive to small accidental movements.

Issue #1128, opened 2021-04-13
Support the new Google analytics gtag.js 


ThomasTNO:
Similar to: https://github.com/readthedocs/readthedocs.org/pull/7691

Can we add support for `gtag.js` here?

Blendify:
Feature added in ef9a0584ef6d43f898a2b787a541e31a79f560e8

PR #1129, opened 2021-04-13
Added support for gtag

ThomasTNO:
Solves: https://github.com/readthedocs/sphinx_rtd_theme/issues/1128

Implements the new `gtag.js` Google Analytics.

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -13,7 +13,7 @@ For example:
 .. code:: python
 
     html_theme_options = {
-        'analytics_id': 'UA-XXXXXXX-1',  #  Provided by Google in your dashboard
+        'analytics_id': 'G-XXXXXXXXXX',  #  Provided by Google in your dashboard
         'analytics_anonymize_ip': False,
         'logo_only': False,
         'display_version': True,
@@ -97,7 +97,7 @@ Miscellaneous options
 .. confval:: analytics_id
 
     If specified, Google Analytics' javascript is included in your pages.
-    Set the value to the ID provided to you by google (like ``UA-XXXXXXX``).
+    Set the value to the ID provided to you by google (like ``UA-XXXXXXX`` or ``G-XXXXXXXXXX``).
 
     :type: string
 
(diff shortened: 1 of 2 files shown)


Blendify:
Is gtag backwards compatible with the old ga tracking code?

sphinx_rtd_theme/layout.html
```diff
@@ -220,16 +220,23 @@
...
-
-    ga('create', '{{ theme_analytics_id }}', 'auto');
-    {%- if theme_analytics_anonymize_ip|tobool %}
-    ga('set', 'anonymizeIp', true);
-    {%- endif %}
-    ga('send', 'pageview');
+      (function () {
+        // New Google Site Tag (gtag.js) tagging/analytics framework
+        // https://developers.google.com/gtagjs
+        var script = document.createElement("script");
```
Blendify, line 226: Why are you using JS to create the tag? Why not just create the tag yourself in html. This is a lot more complicated than what is recommended on https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs

ThomasTNO:
> 
> 
> Is gtag backwards compatible with the old ga tracking code?

Yes, it is. https://developers.google.com/analytics/devguides/collection/upgrade/analyticsjs

sphinx_rtd_theme/layout.html
ThomasTNO: Fully agree, resulted from copy-paste laziness. Issue has been addressed

docs/configuring.rst
```diff
@@ -97,7 +97,7 @@ Miscellaneous options
 .. confval:: analytics_id
 
     If specified, Google Analytics' javascript is included in your pages.
```
Blendify, line 99: We should add docs here about the particular implementation we use.

Blendify (commented):
Changes seems fine, one small documentation comment.

@agjohnson is there anything that would cause concern here on read the docs?

docs/configuring.rst
ThomasTNO: Added a link to `gtag.js` in the docs

Blendify (approved):

ThomasTNO:
Will this be taken into account for the next release?

PR #1130, opened 2021-04-14
Docs: Build with Sphinx --pre release

humitos:
If a Requirement specifier includes a pre-release or development
    version (e.g. >=0.0.dev0) then pip will allow pre-release and development
    versions for that requirement. This does not include the != flag.

From https://pip.pypa.io/en/latest/cli/pip_install/#pre-release-versions

> Do not merge this PR, it's just a test to build the theme with Sphinx --pre release

Edit: I did this as an example for https://github.com/sphinx-doc/sphinx/issues/9088#issuecomment-819375712


Blendify:
I guess this should be used with https://github.com/readthedocs/sphinx_rtd_theme/pull/1123?

humitos:
@Blendify I added this as an example of https://github.com/sphinx-doc/sphinx/issues/9088#issuecomment-819375712

I don't think we want to add this to `master` for building our docs. However, it may be a good idea to add it to the Sphinx versions that are used to test our theme so we can find issues on our theme before the release is done.

docs/requirements.txt
```diff
@@ -1,2 +1,2 @@
-sphinx>=3.0
+sphinx>=4.0.dev0
```
ericholscher, line 1: We should pin docutils==0.17 as well to test

humitos: Seems you already done it.

humitos:
@ericholscher One failure that I found is that this page breaks: https://sphinx-rtd-theme--1130.org.readthedocs.build/en/1130/demo/demo.html. The table of content does not render properly.

See the original: https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/demo.html

It seems the CSS selector is too specific based on `.section` class that it was replaced by a `<section>` html tag in docutils==0.17

Blendify:
Can this be closed now?

humitos:
Yes. Let's close it. It was a test for the pre-release of Sphinx.

However, we should probably create an automated way of building our theme with the latest pre-release of Sphinx to detect breaking changes earlier.

PR #1132, opened 2021-04-18
Align top and side navigation background colors on mobile

eerimoq:
Hello!

This PR aligns top and side navigation background colors on mobile when configured with `style_nav_header_background` in `conf.py`. Currently only the background of the side navigation header is affected by the configuration, but it would be great if the top navigation bar in mobile phones was affected as well.

Configuration:
```python
html_theme_options = {
    'style_nav_header_background': '#7289da'
}
```

Before this PR:
image
image

After this PR:
image
image

Thanks!


Blendify (approved):

Blendify:
That is up to either @ericholscher or @agjohnson we have been trying to push out a release soon. Unfortunately, things have been slow from their side.

PR #1135, opened 2021-04-21
Farsi translation

hadi-f90:
I've heard that this theme supports RTL languages. So, here's the Persian Translation of the theme messages as the first RTL language translation of this theme.


agjohnson:
Hi! Thanks for the contributions. We manage translations in Transifex: https://sphinx-rtd-theme.readthedocs.io/en/stable/contributing.html#translations

I can try to upload your edits, but it's best if edited and review happens on Transifex otherwise the tooling gets confused.

With regards to RTL, we don't support RTL correctly yet -- while some elements will support RTL, navigation and some other block level content always assume LTR. We'd be happy to have the translations either way however!

agjohnson:
I was able to bring this into Transifex without any issue and reviewed the strings there. I just did a reverse translate using Google translate to review the string though. We'll bring in translations from Transifex for the next release

PR #1137, opened 2021-04-24
Fix field list with long terms

Blendify:
Currently, fieldlist extend the full content and may even overflow. Now with this patch, terms are limited to 20% of the field list.

This patch also refactors the margin to be handled by css grid gap property.

Possible future changes:

These are things that I have noticed that can be fixed in future patches.

- Remove left padding -- seems unnecessary to me
- Remove bottom margin on last child globally
- Collapse grid on mobile


Blendify:
Could this have a review, here is a bug report we got about this issue: https://developer.blender.org/T100721

benjaoming (approved):
This seems like the correct fix to me -- certainly, we could want it to wrap on mobile instead, but nonetheless this is an improvement of current behavior.

benjaoming:
I think that @agjohnson is trying to wrap up 1.1 for an imminent release. So targeting this for 2.0. I know that it's just a small fix, but I think @agjohnson appreciates keeping the work load minimal for the release to come out.

PR #1140, opened 2021-04-30
Remove discouraged type="text/javascript" from scripts

andersk:
`type="text/javascript"` is unnecessary and even discouraged by the [specification](https://html.spec.whatwg.org/multipage/scripting.html#attr-script-type): “Authors should omit the type attribute instead of redundantly setting it.”

It causes the validator to raise [warnings](https://validator.w3.org/nu/?doc=data%3Atext%2Fhtml%2C%3C!DOCTYPE+html%3E%3Chtml+lang%3Den%3E%3Cmeta+charset%3DUTF-8%3E%3Ctitle%3ETest%3C%2Ftitle%3E%3Cscript+type%3D%22text%2Fjavascript%22+src%3D%22test.js%22%3E%3C%2Fscript%3E), too.


Blendify (approved):

Blendify:
Originally this was needed for html4 support but html4 was never really well supported by the theme.

PR #1149, opened 2021-05-20
Disable inline linenos user-select

ndunnewind:
In Sphinx 4, linenos are shown inline by default. These inline linenos are currently selectable. This PR fixes this by setting `user-select` to `none` for `span.linenos`, as in https://github.com/sphinx-doc/sphinx/commit/0ce7c1f57732dda24708fbd3f8f5c67f653f3850.


ndunnewind:
I just noticed this will already be fixed by #1123.

PR #1170, opened 2021-06-25
Adds highlight color to linked anchors and changes the global highlight color to a lighter yellow

nienn:
### Adds highlight color to linked anchors
Highlights any targeted `dt` and any `h1` `h2` `h3` `h4` or `h5` that is a direct child of a targeted section by adding CSS styles to those elements.

**Targeted anchor headers**
HighlightedAnchorHeader

**Targeted anchor `dt`s**
HighlightedAnchorDt1
HighlightedAnchorDt2

<br><br>
### Changes the global highlight color to a lighter yellow
It also styles the `.highlighted` text to `color: $text-color` _(_ #404040 _)_.

**Before**
HighlightedText-before

**After**
HighlightedText-after


Blendify:
The lighter yellow is a nice change however, I find highlighting the whole section heading distracting.

Perhaps a better solution would just be to pulse/animate a highlight once. This would solve the UX issue while not being distracting

nienn:
I agree that would be a good solution for the linked anchors 👍 

For search results (class `.highlighted`) I think it's a better user experience to leave the results highlighted.

campellcl:
This would especially be nice for the Sphinx [`glossary` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary) as mentioned in #877!

PR #1171, opened 2021-06-25
Fix support for logos as urls

tronical:
Sphinx since 4.0 appears to support html_logo to also contain urls.  As
a consequence, the other themes in sphinx appear to be just using the
new logo_url variable, and in this theme that fixes support for it, too.


sphinx_rtd_theme/layout.html
```diff
@@ -131,7 +131,7 @@
             {#- Not strictly valid HTML, but it's the only way to display/scale
                 it properly, without weird scripting or heaps of work
             #}
-            <img src="{{ pathto('_static/' + logo, 1) }}" class="logo" alt="{{ _('Logo') }}"/>
+            <img src="{{ logo_url }}" class="logo" alt="{{ _('Logo') }}"/>
```
Blendify, line 134: This will fail users of older sphinx versions use an updated version of the theme. You will have to do a version comparison and keep both methods available. I am going to put a hold on this PR until we can figure out how to do proper tuple comparisons; see https://github.com/readthedocs/sphinx_rtd_theme/pull/1123

Blendify (changes requested):

Blendify (approved):
Looks good pending https://github.com/readthedocs/sphinx_rtd_theme/pull/1123

tronical:
Thanks for making those changes!

PR #1175, opened 2021-07-14
Add a CI test for unbuilt assets

agjohnson:
Many of our PRs have out of date assets, making it hard to test the
latest changes. This will at least throw an error on the PR when build
assets aren't updated.


agjohnson:
Looks like assets are not built in master :shrug:

agjohnson:
Actually master looks to not like latest sphinx release as well. These are not my monkeys.

Blendify:
Tests should now pass after I have rebased the patch.

agjohnson:
Looks like assets weren't rebuilt on master. Is this error in the failed check clear enough for now?

agjohnson:
Fixed the error: 
image

agjohnson:
Using job dependencies in the config now too, as `build` job sometimes ended up the last to build, which ends up backing up our builders needlessly. PR failure will now look like this:

image

.circleci/config.yml
```diff
@@ -13,8 +13,28 @@ commands:
...
+      - checkout
+      - run: npm ci
+      - run: npm run build
+      - run:
+          name: Ensure built assets are up to date
+          command: |
+            if [[ `git status sphinx_rtd_theme/ --porcelain` ]]
+            then
+              echo "ERROR: assets are out of date. Make sure to run 'npm run build' on your branch."
+              git status --porcelain
```
stsewd, line 28: should we add the same path here?

stsewd (approved):
wonder if the depending on the node version the generated files will be different...

.circleci/config.yml
agjohnson: Makes sense! I forked this from an existing CircleCI orb, so didn't pay too much attention here.

agjohnson:
In theory, the node version could cause assets could be different, but it's probably unlikely. It's not even clear why we hit this on our application assets, but also we were building with releases as ancient as node 8, so it's not too surprising. Something to watch for I'm sure -- we can either use a circleci node image with unspecified version of python, or pin the version of node allowed in package.json, to eliminate node 8 built assets/etc.

PR #1182, opened 2021-08-03
Fix #1177: Usability problem with sidebar on mobile devices

### LINKED ISSUE #1177: Usability problems with sidebar on mobile devices
Dear authors,

thank you for this wonderful theme! I am writing because of a usability issue that I see on mobile devices -- or on any device that is so narrow that the sidebar is hidden and the Hamburger menu becomes visible on the top-left.

I can open the sidebar by clicking on the Hamburger-menu, but it is not obvious to many of my users how to close it again. Here is what I observe:

- users try to close the sidebar with a "swipe left" gesture
- users try to close the sidebar by click to the right of it

Not everybody found out that you need to click the Hamburger icon a second time and this led to some frustration. People generally seem to expect that the Hamburger icon is not active because it has been moved away from its standard location. Perhaps to would be possible to honor swipe gestures and/or clicks on the right?

Best -- and thanks again,

Stefan.

nienn:
### Changes

* Added a close button inside the sidebar just above the toctree
* CSS: Made the button visible only on small screens
* CSS: Made the sidebar 100%, when open, on small screens
_(before it was set to 85% width so the original hamburger button that toggles the sidebar could still be accessed with the menu opened. now the close button was added inside the menu itself, so there's no need to access that button while the menu is opened)_
* CSS: Made the `.rst-versions` 100% width accordingly.

_See bellow screenshots of before and after changes, both displaying the full window with the menu open_

**Before:**
old-menu

**After:**
new-menu

Fixes #1177


sphinx_rtd_theme/layout.html
```diff
@@ -170,6 +170,12 @@
 
         {%- block navigation %}
         <div class="wy-menu wy-menu-vertical" data-spy="affix" role="navigation" aria-label="{{ _('Main') }}">
+          
+          {#- MOBILE NAV, TRIGGLES SIDE NAV ON TOGGLE #}
+          <div class="wy-close-top">
+            <i data-toggle="wy-nav-top" class="fa fa-close"></i>
```
ericholscher, line 176: Is there a way to make this more accessible? I wasn't really sure what hitting the X would do when I tried to use it -- perhaps even some text along with the `X` like `Close Sidebar X`? I'm guessing there's also an `aria-label` we should be using here for screen readers?

ericholscher: It looks like a close button is the example in these docs, so I think we should definitely include one: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute#examples

agjohnson: `aria-label` is the best solution here, yeah. Verbose text would fill out too much space and would need to be translated, which introduced other horizontal spacing concerns.

nienn:
Added `aria-label` and some extra padding to the close button so it's easier to click.

sphinx_rtd_theme/layout.html
ericholscher: I do wonder about something a bit more explicit that it's closing the menu, but I agree that translation point is a good one. Do the aria labels also need to be translated?

agjohnson (commented):
This is a great upgrade, the 85% menu always bothered me a bit.

Do you think there is anything we can easily to do put the close button at the top of the navigation pane?

image

The top left position would position the close button right over where the hamburger menu was before opening the menu, though top right also feels intuitive.

I don't feel like there are good options that don't either 1) bump the content below the button down awkwardly or 2) potentially overlap elements if we make the close button float or fixed position. Neither are great, really. This PR could be a good interim location for the close button, but might not be as discoverable.

Ultimately, I could see us moving towards a menu that includes the top navigation bar as well. It would be nice to preserve the project title in both views probably. Something like:

image

This would most certainly be a larger change though, and probably not something to jump right into unless it's easy. The HTML structure for the top bar and the side navigation probably makes this a bit more work.

We can continue discussion separately if we want to do this in multiple steps.

HYNNSDPG (approved):
785039E8

sphinx_rtd_theme/layout.html
agjohnson: Yeah, aria elements should also be translated. The theme actually has pretty active translations, so localization is definitely a bit more effective here.

agjohnson:
Going to wait to merge until 1.0 is tagged.

sphinx_rtd_theme/layout.html
```diff
@@ -170,6 +170,12 @@
 
         {%- block navigation %}
         <div class="wy-menu wy-menu-vertical" data-spy="affix" role="navigation" aria-label="{{ _('Main') }}">
+          
+          {#- MOBILE NAV, TRIGGLES SIDE NAV ON TOGGLE #}
```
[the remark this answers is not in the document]
nienn: **Triggle** definition from the [Urban Dictionary](https://www.urbandictionary.com/define.php?term=triggle): 
`An aggressive tickle on someone (or thing) as a show of exuberant affection. with possibility of light biting.`

Didn't invented it on the spot however, copied it over from the hamburger menu code a few lines bellow.
I also tried to correct it, then decided to look the meaning over, decided it actually made a lot of sense and left it there! :innocent:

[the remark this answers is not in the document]
agjohnson: Hah fair enough. In that case, retracted!

nienn:
@agjohnson 
> Do you think there is anything we can easily to do put the close button at the top of the navigation pane?

I agree that the top of the navigation pane would be the best place for the close button.
In it's current layout however, I fear it would cause more problems:

* It would make the top pane, which is already a bit too tall for a good user experience on mobile devices, even taller
* It would invite miss-clicks on the logo link

There are several ways to sove these but all involve more design decisions.

> We can continue discussion separately if we want to do this in multiple steps.

👍 Completely agree! A better UX could definitely be accomplished here by redesigning the top pane and integrating the close button.

agjohnson:
Sounds like we have a plan then. It definitely does seem like a larger project to do this the way we want, we can address this in stages though.

> In it's current layout however, I fear it would cause more problems

These are both good points. I actually almost posted a version with the logo removed. It might be worth considering removing the logo in the mobile menu. It's already hidden in the menu and it feels to me to really interfere with usability of the menu on mobile. We have a horizontal logo too. For projects with square or vertical logos the menu gets even more cramped on mobile.

Blendify:
I want to point out that I also made changes that work towards improving this as part of much larger refactors/rewrites. See: #1122

The PR listed above makes several changes to HTML structure, layout, and breakpoints. Making it a slightly bigger change but also a slightly different UX compared to this patch.

Can you two look at the UX and tell me what you think? I can extract those changes from that PR if we like that UX more than this one.

The one benefit of this patch is that it removes the hamburger icon, which is a plus because the icon is rather confusing and is being removed across global UI design patterns.

nienn:
@agjohnson 
> It might be worth considering removing the logo in the mobile menu

Definitely and option. Or maybe have a second logo version, different from the desktop menu.

@Blendify 
> See: [#1122](https://github.com/readthedocs/sphinx_rtd_theme/pull/1122)

Removing the hamburger button is a big plus! I don't think that making the sidebar a fixed width would be the best option here.
Moving to the new tags would be great thought, but because of Sphinx 3.5 lack of support for html5  I don't think we can do that yet (or could we use html5 on the sidebar, where Sphinx does not interfere?)

agjohnson:
I commented on #1122 -- changing the HTML structure of the navigation is tricky because we have customer and users restyling our theme and expecting the existing current HTML structure. We are already discussing a big backwards incompatible change with bootstrap, and users will have to already port their styles to new HTML and classes once, I'd hate to make that twice for users.

I think we'll be in a better place to make better UX on some of the underlying structure after we have dropped some older Sphinx versions, we only have HTML5 output to worry about, and we are done messing around with the main HTML structures with a Bootstrap port. Until then, I think we're mostly in maintenance mode to get a final release out supporting old Sphinx/etc.

benjaoming:
We'll be bumping the Sphinx requirements for 2.0, so perhaps that's a good milestone for this?

PR #1183, opened 2021-08-03
CSS: Fix oversized logo in IE11

nienn:
### Problem
Because the parent `<a>` tag didn't had a `max-width` set, the child logo, on IE11, was assuming the full image width.

### Changes
Added `max-width` to the `<a>` tag wrapping the logo.

**Before:**
Screenshot 2021-08-03 181944

**After:**
logo-after


agjohnson (approved):
Much better! There is probably an issue or two for this open, as it's been one of the reasons folks have discussed dropping support for IE11. We can probably still discuss dropping IE11 support in an upcoming release, but for now I think most of our issues with IE11 are solved with more CSS and maybe at worst some polyfills.

Blendify (changes requested):
Small issue I am noticing there is an issue with the indentation otherwise looks good.

Blendify (approved):

PR #1186, opened 2021-08-06
Cleanup: Remove clearfix

Blendify:
This mixin generates the css snipit: `*zoom:1` which is a fix for IE5.5 - IE7 which we no longer support.


nienn:
Be careful with this. The clearfix mixin also produces the following code:

```sass
=clearfix
  *zoom: 1
  &:before,
  &:after
    display: table
    content: ""
  &:after
    clear: both
```

Removing will break elements flow for modern browsers too.

nienn:
The `*zoom:1` could be removed safely from the mixin itself but that is on the **wyrm** side.

PR #1188, opened 2021-08-09
Added missing space char in footer

nienn:
Added back a missing space character to the footer that was probably lost among the translation lines.


agjohnson (approved):
Good catch, change looks good! :+1:

Blendify (approved):

PR #1191, opened 2021-08-09
Fix broken math link icon

### LINKED ISSUE #1189: Missing font on mathjax output
On tests from sphinx 1.8 to Sphinx 4.1, the paragraph link text is showing placeholder font.

Test set up: firefox 88 on arch linux, html5 output.

Sphinx 1.8:

image

Sphinx 4.1:

image

This doesn't look to be an issue with the previous release:

image

Blendify:
This likely was indirectly broken with da86e4dc62aeb9cc442ba8247c0fb3a7f09add2d

This commit addes our standard headerlink icon styles to math links aswell, before they had no styling.

Fixes #1189


src/sass/_theme_rst.sass
```diff
@@ -206,7 +206,7 @@
     @extend h2
 
   // This is the #href that shows up on hover. Sphinx's is terrible so I hack it away.
-  h1, h2, h3, h4, h5, h6, dl dt, p.caption, table > caption, .code-block-caption
+  h1, h2, h3, h4, h5, h6, dl dt, p.caption, table > caption, .code-block-caption, .eqno
```
Blendify, line 209: This CSS selector seems a bit too specific in the future we may want to use just `.headerlink`

agjohnson: I'm guessing this list is just for applying the opacity on the parent element, otherwise we might be able to use `.headerlink`, yeah. This seems fine for now.

We could also use the entire `.math` parent element instead here, so that hovering over the math element will give the anchor link. I'm not sure how folks normally use the equation number here, but it might be hard to discover this anchor link if the hover event is just on the equation number.

agjohnson (commented):
Looks better! It is a bit hard to find the link to the equation now, though not sure how useful the link was before. I could see however on the parent `.math` class working well too.

Blendify:
> Looks better! It is a bit hard to find the link to the equation now, though not sure how useful the link was before. I could see however on the parent `.math` class working well too.

I agree this is something I noticed but wasn't sure if we want to go out of the scope of the bugfix here.
One solution I found was to make `eqno` `display: block` which makes the icon appear when hovering anywhere next to the headerlink.

Making it appear when covering the whole math equation is also a solution though. Maybe this should be tackled in 1.1 as part of: #383

nienn:
This looks good!

Also :bump to # 383 as part of 1.1.

agjohnson:
Sounds good, let's address it more in 1.1

PR #1192, opened 2021-08-10
Fix #1190: Add classifier separation on html4 writer

### LINKED ISSUE #1190: Definition list classifier separation missing
This may be a regression, I remember a fix being put in place for this, but I'm not yet finding the solution. I'm noticing on Sphinx versions 2.0+ and HTML4 writer, for both the latest and stable versions of the theme. The issue is that the classifier is not separated:

image

HTML5 output works as expected.

Related to #746

nienn:
This fixes #1190. The missing separator exists only as a CSS pseudo element. The selector that creates it is very specific and explicitly excludes the html4 output:

```css
html.writer-html5 .rst-content dl dt span.classifier:before
```

### Solution
Make the selector global:
```css
span.classifier:before
```

### Details
I made the selector global, instead of adding the specific selector container structure required for the html4 case, for two reasons:
* Did not find `span.classifier` being used anywhere else, in a different situation that would do badly with this styling.
_Might I be wrong here?_
* I'm not sure about the html structure for the html4 writer.
_Ex1: html4 writer uses tables instead of `dl`s (I did not test this thought)_
_Ex2: v 0.5.2 does not add the class `.writer-html5`, even if it's using  `dl`s_


Blendify (approved):

src/sass/_theme_rst.sass
```diff
@@ -270,6 +270,12 @@
   .hlist
     width: 100%
 
+  // Global classifier class format
+  // the classifier class is used both inside the Html4 writer tables and the html5 writer definition lists
+  span.classifier
```
agjohnson, line 275: This could be too generic for user content. I could see user content like API docs (similar to the demo docs API docs) having overlap on this selector. Perhaps a bit more specificity here would be an okay middle ground, ie `dl > dt > span.classifier`

agjohnson (commented):
In trying to explain some more test cases, I discovered the underlying issue: Sphinx 2 changed the structure of this classifier to match what is output by the HTML5 writer (this isn't specifically an HTML5 feature).

The Sphinx 1.x HTML4 output for the classifier actually had a hardcoded separator:

```html
<dt>Term <span class="classifier-delimiter">:</span> <span class="classifier">classifier</span></dt>
```

Thinking about this more, perhaps the easiest solution is to always hide `dl dt span.class-delimiter`, just to keep our selectors more reasonable.

nienn:
Added the `dd dl` to make the selector more specific. The `.rst-content` was also already present. So now the proposed changes on that selector are just,

from:
```css
html.writer-html5 .rst-content dl dt span.classifier:before
```

to:
```css
.rst-content dl dt span.classifier:before
```

Also added the rules to hide the hardcoded separator from Sphinx 1.x and prevent duplication of that element:

```css
.rst-content dl dt span.classifier-delimiter {
	display: none !important;
}
```

agjohnson (approved):
Looks great!

PR #1195, opened 2021-08-10
Bump to 1.0alpha1

agjohnson:
I think we talked about this and never followed through on the next few 
release, not sure though. After we release a new version, I think we should
be updating the repository version to something other than the most current
release.

The effect would be that the version in the repository would be an 
alpha/dev release, and so would be possible to differentiate easily between
0.5.2 release on PyPI and 0.5.3alpha1 development release from Git. Right
now, we don't follow up with incrementing this version, and so it's harder
to tell the difference. This is mostly for our use in development.

It adds a small bit of overhead, but is nice and explicit.

This PR updates the version in the repository to 1.0.0alpha1. If we had 
adopted this workflow, the progression would have been:

* Release 0.5.2, open PR bumping to 0.5.2 (run `bump2version release` and open PR)
* Follow up 0.5.2 release by bumping to `0.5.3alpha1` (run `bump2version patch` and open PR)
* We add a new feature, bump to `0.6.0alpha1` (run `bump2version minor` and open PR)
* Oops, backwards incompatible change. We bump to `1.0.0alpha1` (run `bump2version major` and open PR)
* We're ready for an rc, bump to `1.0.0rc1` (run `bump2version release` and open PR)
* Maybe we cut another rc after some fixes
* We're ready for final release, bump to `1.0.0` (run `bump2version release` and open PR)
* Follow up the release PR by bumping to the next version (run `bump2version patch` and open PR)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,11 +2,10 @@
 Changelog
 *********
 
-master
-======
+Development version (|development_version|)
+===========================================
 
-v1.0.0
-======
+.. |development_version| replace:: 1.0.0alpha1
 
 Incompatible Changes
 --------------------
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -197,4 +197,8 @@ To release a new version of the theme, core team will take the following steps:
       $ python setup.py sdist bdist_wheel
       $ twine upload --sign --identity security@readthedocs.org dist/*
 
+#. Finally, open a new pull request updating the development release version to
+   the next patch by running ``bump2version patch``. Open a pull request with
+   this change.
+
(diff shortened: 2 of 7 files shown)


Blendify:
This does not update the release notes, should we add a specific heading for pre release versions?

On the topic the release notes may be missing some changes that were made to master, this should be double checked.

agjohnson:
Ah good point. I'm not great about updating the changelog, but we can make this explicit as well. I added an example in 
b74323c -- is this what you're thinking? I imagine at release time we'd just add a new heading under `.. replace` for the release and continue on.

Blendify (approved):

PR #1185, opened 2021-08-06
Add support for docutils 0.17

Blendify:
This should'nt be a breaking change as backward compatibility is preserved so this change could go into 1.0 or wait until 1.1.

I know a lot of users are asking for this so it might be nice to include this in 1.0.

Fixes #1115

PR #1199, opened 2021-08-12
Fix spacing issue with HTML4 images

### LINKED ISSUE #1198: Issue with image spacing still
While doing more complete QA, the issue noted in #1185 around image spacing is still visible on HTML4 output, any version of Sphinx:

image

Image spacing on HTML5 looks better:

image

It wasn't clear what bug the image spacing fix was trying to resolve.

agjohnson:
This bug was introduced in #1185, it looks like the HTML4 selector
changed to not include directly descendents of `.section` elements.

I believe the issue was the missing ` > ` introduced here:
https://github.com/readthedocs/sphinx_rtd_theme/pull/1185/files#diff-2ad748ac9ebc515229907afab210e3bb370986d3db3a377f052ee5048dcc5b6cR36

But I'm also not familiar with the element that these rules are trying to target.

Fixes #1198


agjohnson:
@Blendify it's not really clear what the fix in #1185 was addressing, so I'm not exactly sure what to test to verify this change. What was this selector supposed to target? Fix was easy, it seems just the direct descendant selector was missing.

Blendify (approved):
I am unsure exactly also, we really need to better organize and comment what styling goes to what markup

agjohnson:
Heh yeah, seems we might have some investigation to do

PR #1202, opened 2021-08-12
Fix spacing issue with html4 output

### LINKED ISSUE #1201: Spacing bugs with docutils 0.17 and toctree
Still digging into which is the correct bug, but noticed competing bad spacing on two versions.

On QA builds[1], I get no indentation at all:

image

1: https://readthedocs-static-dev.s3.us-east-2.amazonaws.com/qa/theme/sphinx4-theme1-html4/index.html#

On local build of `tox -e py3-sphinx18-html4`, I a spacing regression that we've seen before:

image

I'm guessing the latter is the most accurate bug, I can't explain why the QA tests have such a clearly busted toc tree listing.

agjohnson:
The toctree wrapper can exist outside a section on html4 output. This
resolves some of the issues in #1201, namely a bug introduced with docutils 0.17 support.

This adds the missing margin back in:
image

It also protects against the edge case where toctree can be outside a section completely.

Edit: Nevermind, this does not fix the underlying bug completely, I'm noticing other spacing bugs

Edit edit: I think this does now fix #1201 

Fixes #1201


Blendify (commented):
For now this solution works, in the future we may want to remove the ul styles out of the section headings, I don't believe sphinx itself does this so there is a possibility that some styling might fail.

agjohnson:
I switched the approach in dde72c3, this seems to affect both html4 and html5 output positively. I don't get the bug in `sphinx{18,41}-html{4,5}` now.

agjohnson:
I tested quickly with docutils 0.16 and 0.17 and both seemed fine.

Blendify (approved):

PR #1206, opened 2021-08-16
Update translations for 1.0.0rc1

agjohnson:
This was all automated from Transifex, so review on these files will probably not be super helpful here. 1.0.0rc1 will have new strings to translate (these are already updated to transifex with this PR), so hopefully we can get the locales fully translated for a 1.0 final.

I noted the translation updates in changelog.


agjohnson:
Wait. This did not actually get the new translations, a lot are missing :thinking:

agjohnson:
Bah! Okay, the transifex client was very annoying here. First, and maybe not necessary, I need to copy a translation for the new locales to trick transifex-client into recognizing the new locales to sync. After I fixed that, the transifex-client would skip all of the locales and not mention why. I finally discovered it's because of the minimum translation percentage, which had dropped due to new string. I bumped this down to 80% for now.

All translations should be updates now.

Blendify (approved):

agjohnson:
I rebased and gave `update_translations` another pass. I'll merge after as it's just more unreviewable po/mo file updates. I did drop unreview percentage down again to get polish to be included again. We can bump this back up later, but i expect a few of the translations to catch up.

PR #1208, opened 2021-08-17
Add development and usage documentation

agjohnson:
This finally defines our supported browsers and dependencies publicly,
and communicates our next several versions that we have planned so far.
This information is starting with 1.0.0, but I did try to capture some
historical information as well.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -90,7 +90,7 @@ version:
 
 .. code:: console
 
-    % pip install "sphinx_rtd_theme<=2.0.0"
+    $ pip install "sphinx_rtd_theme<=2.0.0"
 
 .. _semantic versioning: http://semver.org/
 
@@ -111,7 +111,7 @@ it resolves bugs with the latest release of Docutils, version 0.17.
 Starting with this release, several deprecation warnings will be emitted at
 build time:
 
-Direction installation is deprecated
+Direct installation is deprecated
     Support for direct installation through GitHub is no longer a suggested
     installation method. In an effort to ease maintenance, compiled assets will
     eventually be removed from the theme repository. These files will only be
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -6,6 +6,8 @@ planning. Documentation authors can decide which theme release works best for
 their project based on required browser/operating system combinations or
 dependency support.
 
+.. _supported-browsers:
+
 Supported browsers
 ------------------
 
@@ -46,6 +48,8 @@ Opera (any OS, any version)
     **Community support only.** We do not receive enough traffic with this
     browser to officially support it in testing and development.
 
+.. _supported-dependencies:
+
 Supported dependencies
 ----------------------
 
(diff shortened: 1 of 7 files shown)


astrojuanlu (commented):
Awesome to have this 🚀

docs/development.rst
```diff
@@ -0,0 +1,174 @@
...
+Opera (any OS, any version)
+    **Community support only.** We do not receive enough traffic with this
+    browser to officially support it in testing and development.
+
+Supported dependencies
+----------------------
+
+The theme officially supports the following dependencies in your Sphinx project:
+
+.. list-table:: Supported dependencies
```
ericholscher, line 54: This feels like we should have it in the README, not hidden away in a page only for developers. I think we want this to be front & center for all users.

ericholscher (approved):
This is great content, but I think we want a lot of this information in the README. At least the set of supported things in each released version.

The roadmap can stay here, but I think some of the 1.0.0 content needs to be pulled up into the README as well, in a way to warn users about the upcoming changes and what versions to pin in what cases.

docs/development.rst
```diff
@@ -0,0 +1,174 @@
...
+:ref:`0.5.2` release. It will drop support for Sphinx 1.6, which is a rather old
+release at this point.
+
+This version will add official support for the Sphinx 4.x release series and
+it resolves bugs with the latest release of Docutils, version 0.17.
+
+Starting with this release, several deprecation warnings will be emitted at
+build time:
+
+Direction installation is deprecated
```
ericholscher, line 108:
```suggestion
Direct installation is deprecated
```

docs/development.rst
agjohnson: I'll link it from the readme

docs/supported-browsers.csv
```diff
@@ -0,0 +1,5 @@
+Browser,Operating System,Versions
+Chrome,"Windows, MacOS, Linux, Android",>=90
+Firefox,"Linux, Windows",>=90
+Edge,Windows,>=90
+Safari,"MacOS, iOS",>=13
```
ericholscher, line 5: I see no IE here, this makes me happy 👍

agjohnson: Yeah, this is the change that I was describing. We are still testing IE11, but it is not supported and receives no new features.

agjohnson:
I pushed up more changes than i'd like to get the additional links in the readme, but this was due to the fact that we were just includeing `readme.rst` on the index page. Including the readme breaks doc links at the index, but always pointing to stable version. So, I also took on reworking the index page for the documentation to not reuse readme.rst, consider this a first pass and we can come back to really reworking that page next.

humitos (approved):

docs/development.rst
```diff
@@ -0,0 +1,180 @@
...
+
+The theme developers follow the guidelines below for development and release
+planning. Documentation authors can decide which theme release works best for
+their project based on required browser/operating system combinations or
+dependency support.
+
+.. _supported-browsers:
+
+Supported browsers
+------------------
```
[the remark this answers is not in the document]
agjohnson: Hah, well I was originally trying to keep scope down here. Could be a future update though!

PR #1218, opened 2021-08-24
Allow specific deactivation of table handling

### LINKED ISSUE #1179: Table handling for sphinx-extensions
### Problem
The rtd theme wrapps all tables in a ``div`` with class ``wy-table-responsive`` via javascript.
Unfortunately some html-tables in sphinx-projects are not coming from "normal" rst-tables, but are created by sphinx-extensions, which provide their own table setup/manipulation.

An example is the [Sphinx-Needs extension](https://sphinxcontrib-needs.readthedocs.io/en/latest/), and there especially the directives [need](https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/need.html) and [needtable](https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needtable.html).
See also bug report #767 from the past.

Here a table comparison for ``needtable`` in the themes alabaster and rtd:
image

The first table is the important one, which gets handled by the JS-lib [DataTables](https://datatables.net/).
With rtd, head and body are separated and the columns are not aligned.

### Reproducible Project

You can take a look into the documentation on rtd. I have activated a branch which is using the rtd theme:

[needtable docs with alabaster](https://sphinxcontrib-needs.readthedocs.io/en/latest/directives/needtable.html#columns)
[needtable docs with rtd](https://sphinxcontrib-needs.readthedocs.io/en/rtd_checks/directives/needtable.html#columns)

[Sphinx-Needs doc sources](https://github.com/useblocks/sphinxcontrib-needs/tree/master/docs)

### Problem source
As I'm the maintainer of Sphinx-Needs, I tried to provide my own css-configuration, which contains rules for RTD only.
This works in some cases, but not in all.

Problem is also that rtd changes the DOM by inserting a div-wrapper with class ```wy-table-responsive``.
Which may confuse other JS-code.

### Proposal / Fix
For sure rtd-theme can not handle all the  exceptions needed by other extensions in its code.
But maybe there is a way to exclude certain table-configurations from the default rtd table-handling.

``theme.js`` is already doing this for classes like ``citation`` or ``footnote``.
See: https://github.com/readthedocs/sphinx_rtd_theme/blob/3532ffc89bbf89ef7d2ac62d65b3683f588a9673/src/theme.js#L102

Maybe we can extend this line by a specific class-name ``.rtd_exclude``, which can be set by sphinx-extensions to avoid the rtd-specific handling for these tables:
```js
$("table.docutils:not(.field-list,.footnote,.citation,.rtd_exclude)")
            .wrap("<div class='wy-table-responsive'></div>");
```

There may be some more positions in the code, where such an exception-handling is needed (haven't checked everything yet).

Would this be a proper solution.
I'm happy to create a PR, but first please give me your feedback :)

danwos:
Setting the table class "rtd-exclude-wy-table"
deactivates the normal table handling for this
table.

So the table gets not wrapped into a
"div.wy-table-responsive" container.

Needed by other Sphinx extensions, which
care about their tables on their own.

Fixes #1179


danwos:
@nienn : I hope this PR is fine and solves #117 as expected.

I have also added some documentation and a changelog entry.
As this is my first PR on this project, please let me know if anything is missing.

Local tests together with Sphinx-Needs are looking great.

agjohnson:
Thanks for the contribution, we'll see about getting this in version 1.1. I'll block this PR for now, we're still wrapping up a 1.0 release.

danwos:
@agjohnson Thanks for your feedback and for the overall work on RTD.

The change itself is really small, one line.
And it does not have any side-effects to the theme or any code.
So really no change to get this small bugfix into 1.0?

Not having a possibility of deactivating the automatic table handling makes RTD-Theme incompatible with some other sphinx-extensions. I, as an affected extension developer,  would be happy to avoid this incompatibility as fast as possible.

The other solution would be to ask my users to use another theme or use the fork, which contains my changes.
Something, I really don't like.

agjohnson:
> So really no chance to get this small bugfix into 1.0?

We'll look to getting this in for 1.1, we don't want to keep holding back a 1.0 release.

twodrops:
This issue is extremely critical for us. The fix solves a major problem that is affecting 100s of tables that we have in RTD theme. Any chance to get this in soon?

danwos:
Any updates on this?
``1.1`` is on the way and this PR is set to ``blocked``
Anything I can do here?

PR #1220, opened 2021-08-25
Fix #1214: Sphinx 4 <pre> overflow

### LINKED ISSUE #1214: HTML formatting issues after updating sphinx
### Problem

I'm hosting the documentation for my python project on readthedocs, there are 2 formatting problems in the following section:

https://dpowers.readthedocs.io/en/latest/reference/power.html#Dpowers.KeyboardAdaptor.key

1st problem: The right margin is not respected (in this section of the documentation only).
2nd problem:  (for all sections) The font color of class and method names/arguments is entirely blue, although it should be black (with blue background). Other documentations on readthedocs look fine in my browser. --- similar to #1212

They occured after updating sphinx version to 4.1.2 from 3.4.3 - I can reproduce this locally on my computer.

### Reproducible Project

The source folder for the documentation is here:
https://github.com/dp0s/Dpowers/tree/master/docs

#### Error Logs/Results

Local build on my computer with sphinx 4.1.2

Screenshot at 2021-08-19 22-58-35

#### Expected Results

Local build on my computer with sphinx 3.4.3

Screenshot at 2021-08-19 22-32-24

### Environment Info

- Python Version: 3.7
- Sphinx Version: 4.1.2 / 3.4.3
- RTD Theme Version: 0.5.2

### Note
I recently introduced the sphinx hoverxref extension, which is still in beta mode. So I rebuilt the docs for an old version of my repository (v0.1.2) without hoverxref extension and the problem is the same.

nienn:
This adds a `max-width` to the Sphinx 4 `dl.property` to prevent the overflowing of the `pre` tag that was occurring when placed inside this `dl`.

This specific `dl` class is set to `display:inline-block`, which doesn't deal very well with the `white-space:pre` from the `pre` tag. Adding a `max-width` to the `dl.property` should fix it.

**Altered CSS**
```css
html.writer-html4 .rst-content dl:not(.docutils) .property, 
html.writer-html5 .rst-content dl[class]:not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple) .property {
  display: inline-block;
  padding-right: 8px;
  max-width: 100%;
}
```

Fixes #1214


dp0s:
Thanks :)

nienn:
Np. This will probably have to wait for the current release to move forward thought.

agjohnson:
This looks like a good change to include in 1.0, our first release supporting Sphinx 4.

agjohnson (approved):
Easy enough! I don't think there is a test case for this in our docs either way, but didn't notice any significant changes with our test docs from this :+1:

PR #1252, opened 2021-11-10
Fix a italian translation

fedeliallalinea:
Footer translation from
`Built with Sphinx using a theme provided by Read the Docs.`
to
`Realizzato con Sphinx e il tema tema fornito da Read the Docs.`
is wrong, the correct sentence would be
`Realizzato con Sphinx usando un tema fornito da Read the Docs.`


benjaoming:
Hi @fedeliallalinea 

Thanks for this! We do translations via Transifex: https://sphinx-rtd-theme.readthedocs.io/en/stable/contributing.html#translations

I've added your translation. If you wish to be a part of the translation team, have a look here: https://www.transifex.com/readthedocs/sphinx-rtd-theme/

PR #1269, opened 2021-12-28
Show version for Sphinx and this template in footer.

marxin:
Emits now:

Screenshot from 2021-12-28 15-08-29

### DIFF
--- a/sphinx_rtd_theme/__init__.py
+++ b/sphinx_rtd_theme/__init__.py
@@ -31,6 +31,9 @@ def config_initiated(app, config):
             _('The canonical_url option is deprecated, use the html_baseurl option from Sphinx instead.')
         )
 
+def update_context(app, pagename, templatename, context, doctree):
+    context["rtd_theme_version"] = __version_full__
+
 # See http://www.sphinx-doc.org/en/stable/theming.html#distribute-your-theme-as-a-python-package
 def setup(app):
     if python_version[0] < 3:
@@ -60,4 +63,5 @@ def setup(app):
     else:
         app.config.html_add_permalinks = "\uf0c1"
 
+    app.connect("html-page-context", update_context)
     return {'parallel_read_safe': True, 'parallel_write_safe': True}
--- a/sphinx_rtd_theme/footer.html
+++ b/sphinx_rtd_theme/footer.html
@@ -47,12 +47,12 @@
   </div>
 
   {% if show_sphinx %}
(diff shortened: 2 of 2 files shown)


marxin:
> MIT8

Can you please explain this? Is the community interested in the change, or should I close the pull request?

benjaoming:
@marxin I've really been wanting this, too :tada: 

We have some issue with legacy installation setups, though. So if we introduced it _as is_, it would unfortunately give the same errors as described here:

https://github.com/readthedocs/sphinx_rtd_theme/issues/1355

@agjohnson moving forwards, we should probably put something on the roadmap about making legacy setups unsupported and find a way to detect them and fail nicely? On of the steps needed is also unfortunately on readthedocs.org - https://github.com/readthedocs/readthedocs.org/pull/9654

marxin:
Please make a similar PR if you want the feature in the future. This PR conflicts, thus I'm closing it.

benjaoming:
@marxin I was going over https://github.com/readthedocs/sphinx_rtd_theme/issues/1355 and recalled that it only failed because an undefined context variable was used in a way that it cannot.

If you implement it in a way that anticipates that `rtd_theme_version` may be `None` then it can work. For instance `{%- if rtd_theme_version -%}(v{{ rtd_theme_version }}){%- endif -%}`

Btw that "MIT8" comment is bot spam that we're unfortunately seeing a lot of on GitHub nowadays.

marxin:
> If you implement it in a way that anticipates that `rtd_theme_version` may be `None` then it can work. For instance `{%- if rtd_theme_version -%}(v{{ rtd_theme_version }}){%- endif -%}`

The question is: How can it be `None` if I set it in:
`app.connect("html-page-context", update_context)`
?

benjaoming:
That's the longer story that is uncovered in https://github.com/readthedocs/sphinx_rtd_theme/issues/1355 and https://github.com/readthedocs/readthedocs.org/pull/9654

Basically, you cannot rely on `app.connect` to be executed in two cases:

* Legacy configurations that didn't add the theme the correct way
* On RTD `sphinx_rtd_theme` has to be added to `extensions` in `conf.py` which is both unnecessary and unconventional (due to an injection that RTD does, which is probably also related to legacy)

We are facing the same issue with the automatic enabling of sphinxcontrib-jquery which adds some priority to fixing the issue.

marxin:
Sure, I've just made a new PR in #1410.

Issue #1273, opened 2022-01-10
Unequal number of <div> and </div> in sphinx_rtd_theme/layout.html


thomas-nilsson-irfu:
### Problem

The included template, [sphinx_rtd_theme/layout.html](https://github.com/readthedocs/sphinx_rtd_theme/blob/d64dadf1ceec4f9ff6c1ca2d3ea4c3d0fdb0e8d2/sphinx_rtd_theme/layout.html), does not have an equal number of `<div>` and `</div>` tags making it effectively invalid as per html standard.

Counting the numbers of `<div>` tags I get 12 and counting `</div>` I only get 11...

### Reproducible Project

Including a simple shell script example:

```shell
# number of start tags:
grep -c '<div' sphinx_rtd_theme/layout.html
# number of end tags:
grep -c '</div' sphinx_rtd_theme/layout.html
```

#### Expected Results

Correctly start (`<div>`) and end (`</div>`) all tags in the template so it correctly follows the HTML standard, neither tag is omissible.

Ref:
https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element

### Environment Info

- Python Version: N/A
- Sphinx Version: N/A
- RTD Theme Version: master branch (as of 2022-01-10).

NaincyKumariKnoldus:
Hey @thomas-nilsson-irfu I want to resolve this error. Can I?

astrojuanlu:
Hi @NaincyKumariKnoldus ! Your contribution is more than welcome. We haven't yet assessed what `<div>` is missing a closing tag.

NaincyKumariKnoldus:
hye @astrojuanlu I will see that which <div> is missing closing tag. Can I?

NaincyKumariKnoldus:
Hello @astrojuanlu and @thomas-nilsson-irfu I have raised now a PR for this and fixed this issue. 
Please review it. Thanks!

astrojuanlu:
xref #1274

PR #1274, opened 2022-01-10
fixed the issue of Unequal number of <div> and </div> in sphinx_rtd_theme/layout.html 

NaincyKumariKnoldus:
I have fixed the issue of Unequal number of <div> and </div> in sphinx_rtd_theme/layout.html. Now its both equal.

Please review and merge my PR.
Thanks!

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -223,6 +223,7 @@
         </div>
       {%- endblock %}
       </div>
+     </div>
     </section>
   </div>
   {% include "versions.html" -%}


astrojuanlu:
I _think_ the CI failures are unrelated, see #1275.

nienn:
This fixes #1273

mitya57:
This change is wrong. Inside the `<section>`, all divs are closed properly.

Perhaps you were counting two kinds of `<div class="rst-content">` as two distinct tags? They are in if/else block, so should be counted as one.

Also, this file uses two-spaces indentation. Your closing tag uses is indented by an odd amount of spaces, and does not match any opening tag.

PR #1281, opened 2022-01-19
Fix failing CI tests

Blendify:
Issue was newer node versions are not compatible with our dev envorment.
This failed because the `cimg/python:3.9-node` defaulted to a newer version of node.
To fix this we can use multiple images to specify a specific node version.

I settled on `node:15.14.0` as it is the last working version.

After we get tests passing we can work on supporting newer node versions.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,7 @@ jobs:
           curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
           echo 'export NVM_DIR=$HOME/.nvm' >> $BASH_ENV
           echo 'source $NVM_DIR/nvm.sh' >> $BASH_ENV
+          nvm install v15.14.0
       - run-build: {}
   py27:
     docker:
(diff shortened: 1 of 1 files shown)


agjohnson:
Yeah, it's wise to take do this in steps. I'm not sure we need to go much higher out the Node version, given the issues we've had.

Looks like tests are failing still, but if you have trouble here, I did solve this on a separate repo using CircleCI workspaces, because there wasn't a good image available with configurable python and node version.

This might be something to try out. I was playing around with a dedicated repo for testing the theme at:
https://github.com/readthedocs/test-sphinx-theme/blob/main/.circleci/config.yml

Blendify:
@agjohnson I am having issues getting the installed node version to get recognized in the CI path.

Do you have any suggestions here?

benjaoming:
Hi @Blendify!

Using two different approaches, we seem to have managed to run Node 14 in an okay fashion but without NVM.

* Circle CI uses a so-called orb since which adds Node 14 to Circle's docker container #1314 
* Docker is introduced as an alternative development environment, which uses `node:14-alpine` in #1319 

I'm going to close this, as NVM currently isn't necessary, although might be useful for another type of local development setup :)

PR #1291, opened 2022-03-12
Fix typos

kianmeng:


nienn:
Thank you for fixing this @kianmeng.

stsewd (approved):
Thanks! BTW, the po files are updated automatically, they don't need to be updated on the same PR.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

Issue #824, opened 2019-10-03
Nav Bar dynamically moving


wsf11:
### Problem

The navigation bar dynamically moves when clicking on links. For large projects, this can be annoying as a user sifting through documentation and constantly having to move the navigation bar back to where it was before clicking. 

### Reproducible Project

[Cloud Custodian](https://cloudcustodian.io/docs/azure/resources/azure-common-filters.html#azure-common-filters-firewall-rules) uses RTD. When clicking around on the nav bar and pages of docs, it often jumps up and you have to constantly move the bar to where you were previous. 

Line 155 in [theme.js](https://github.com/readthedocs/sphinx_rtd_theme/blob/master/src/theme.js)

`link[0].scrollIntoView();` causes the link to be dynamically place at the top of the nav bar, where I would like the nav bar to not move at all when I click on an href. 

#### Expected Results

Navigation bar should stay in the same position while clicking around documentation. 

### Environment Info

- Sphinx Version: 2.2.0

jessetan:
This is not trivial to solve. Since the theme is not an SPA and it does not use an iframe, the navigation will always jump to the top by default. The `scrollIntoView()` was added so the correct menu option is at least in the viewport.
To achieve the requested behavior, we would have to store the current scroll position in a cookie or pass it in the URL. Restoring the scroll position is complex because it needs to take into account menu items that are (no longer) expanded.
This is not a definite no and I understand how this could improve UX, so I'll keep this ticket open. I just don't see this happening anytime soon.

sabinem:
I am struggeling with the same line of code: what I would like to have is a configuration option in the theme, that prevents that `scrollIntoView()` of the active item. I tried to overwrite this with a custom.js script, but since the theme.js is called after my custom.js I could not change the behaviour: 

So what would help be is a configuration option for either:
- preventing the scroll to the active item
- or being able to add custom.js after the theme.js

Would it be possible to implement one of these two options?
-

aiqc:
Logo and search bar look great, but the nav just jumps to the page in view.

PR #1293, opened 2022-03-29
Make the nav sidebar's scroll-to-current-page behaviour less jumpy

tsibley:
This avoids scrolling the sidebar when the link is already in view by
using "nearest" for the vertical axis ("block") instead of the default
of "start".  This behaviour is much better when clicking on items in the
current viewport, especially if the current viewport matches the initial
viewport.

It's still a little jumpy when the current link isn't visible in the
initial viewport, because the sidebar position still changes across page
loads.  But the new behaviour solves the big problem on initial home
page loads with a long sidebar where the logo and project title would be
immediately scrolled out of view.

Related to #824.

---

[Quick video](https://tsibley.net/paste/2022-03-295xz3AzGS-rtd-theme-scrollIntoView-cut.webm) demonstrating before/after.


nienn:
Hi @tsibley, thank you for submission!
This probably needs more in-depth testing with multiple platforms and edge cases, but I think it can greatly improve the user experience with the sidebar.

tsibley:
@nienn Yep, understood about testing! [Data on caniuse.com shows](https://caniuse.com/mdn-api_element_scrollintoview_options_parameter) pretty good support for `{block: "nearest"}` relative to browser-usage across the board, with the notable exception of Safari (on macOS and iOS).

image

I suspect (but haven't tested) that when the `options` object parameter isn't supported, as in Safari, it will be treated as a truthy value for the [`alignToTop` boolean parameter](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#aligntotop) and so will in effect act like the current behaviour.

Would be good to test all this empirically, of course. Does RTD have the resources to do this?

x157660503 (approved):
- [ ] #

tsibley:
Repushed after rebasing onto the current tip of `master` and also regenerating the built `theme.js` asset so it matches the source file.

Issue #1302, opened 2022-06-29
Support docutils 0.18.x


befeleme:
docutils 0.18.x has been with us for quite some time and it has finally started gaining support of major documentation projects, including Sphinx 5.x. Could you also start supporting that version?

befeleme:
In Fedora, I manually relaxed the docutils version requirement and built the package as RPM against the version 0.18.1 & Sphinx 5.0.2.
Please note, we patch our package (especially fonts), so my findings may not be a complete list.
I compared the rendered package documentation to the one build on docutils 0.17 and Sphinx 4.5.0 and found those issues:

1) Field list is rendered with two colons instead of one:
![image](https://user-images.githubusercontent.com/33810531/179508910-1abdf4ab-2f33-42fd-9fc4-fec1d27b7b65.png)

2) Footnotes and citations not visually rendered, functionality (references) without changes:
![image](https://user-images.githubusercontent.com/33810531/179509123-a7c9ccf6-605e-4b57-9d54-b8bd177d2c86.png)

n-peugnet:
@befeleme :
> 1. Field list is rendered with two colons instead of one:

This issue has been [fixed in Sphinx](https://github.com/sphinx-doc/sphinx/issues/10594) and will be [released in version 5.0.3](https://www.sphinx-doc.org/en/master/changes.html#release-5-0-3-in-development)

JasperCraeghs:
> This issue has been https://github.com/sphinx-doc/sphinx/issues/10594 and will be [released in version 5.0.3](https://www.sphinx-doc.org/en/master/changes.html#release-5-0-3-in-development)

The change to Sphinx' style sheet has been released in Sphinx 5.1.0 already. However, sphinx-rtd-theme' style sheet also needs a change like this. Two colons are still rendered with sphinx-rtd-theme and Sphinx 5.1.0.

nedbat:
I'm trying to update the dev dependencies for coverage.py, and get a conflict on docutils, partly because of the requirement in sphinx_rtd_theme:
```
Could not find a version that matches docutils<0.18,<0.20,<0.21,>=0.14,>=0.19,~=0.18.0 (from doc8==1.0.0->-r doc/requirements.in (line 10))
[...]

PR #1304, opened 2022-07-19
Support docutils 0.18

edgarrmondragon:


edgarrmondragon:
Build step is failing but I'm not sure how to fix that

kloczek:
docutils 0.19 has been released.

benjaoming:
Possibly a related bug in this PR:

> Two colons are still rendered with sphinx-rtd-theme and Sphinx 5.1.0.

https://github.com/readthedocs/sphinx_rtd_theme/issues/1302#issuecomment-1195290758

Please do say if you have any findings here :pray:

benjaoming:
I'd say that the "two colons" issue is important, but that it's a very little issue and shouldn't block us from the upgrade here, since it's causing issues for other important projects.

benjaoming (approved):
Thanks for fixing this @edgarrmondragon - it's very helpful to have comments confirming that it also has been tested and works :) I'll proceed to have a look that everything in our demo build looks okay and open up any issues I might find.

kloczek:
Any chance to make new release?
It is already quite long since last release has been made ..

benjaoming:
@kloczek that's definitely the intention

I'll have to investigate these current build issues though:

```
Exception occurred:
  File "/home/circleci/project/.tox/py27-sphinxlatest/lib/python2.7/site-packages/sphinx/ext/intersphinx.py", line 365, in missing_reference
    contnode[0] = nodes.Text(newtarget, contnode[0].rawsource)
AttributeError: 'Text' object has no attribute 'rawsource'
```

benjaoming:
Above error is fixed in https://github.com/sphinx-doc/sphinx/commit/961f5af0963055f0206a437765a087a2bf03bcca

This commmit is released with Sphinx 4.3.0. Hence, Sphinx releases prior to this version definitely have this issue and aren't compatible with docutils 0.18. What might be confusing is that the error isn't occurring 100% reliably.. but that's another story.

Should we handle this simply in our Tox matrix?

benjaoming:
It turns out that the build error was because of a missing lower bound on Sphinx in `tox.ini`, meaning that for Python 2.7, we always got Sphinx 1.8 for everything 2.0+. In combination with docutils 0.18, this gave the error.

After specifying the versions correctly in `tox.ini` and `.circleci/config.yml`, we are fine :+1:

edgarrmondragon:
hey @benjaoming, I've been really busy so thanks for taking on the PR 😄

benjaoming:
Not sure if https://github.com/readthedocs/sphinx_rtd_theme/pull/1338 might cause a need to solve conflicts, currently it has a clean cherry-pick.

spagh-eddie (approved):

twodrops:
I wanted to test this branch with 
`pip install git+https://github.com/edgarrmondragon/sphinx_rtd_theme.git@docutils-0.18`
but somehow the docutils seems to be still set as `< docutils 0.18`

See pip error:
`
The conflict is caused by:
    sphinx-rtd-theme 1.0.1a1 depends on docutils<0.18
`

The branch from @benjaoming however installs correctly, but I do not need docutils 0.19 yet (due to incompatibility with some other packages)
`pip install git+https://github.com/benjaoming/sphinx_rtd_theme.git@docutils-0.19` works.

Any clue why?

benjaoming:
@twodrops 
> I wanted to test this branch with
pip install git+https://github.com/edgarrmondragon/sphinx_rtd_theme.git@docutils-0.18
but somehow the docutils seems to be still set as < docutils 0.18

Can you verify the version of Sphinx and Python that was used for this?

benjaoming:
Failing to reproduce `conflict is caused by: sphinx-rtd-theme 1.0.1a1 depends on docutils<0.18` - docutils 0.18 is correctly installed in the tox environment.

Some Circle CI build output...

```
py310-sphinx51 installed: alabaster==0.7.12,attrs==22.1.0,Babel==2.10.3,certifi==2022.9.24,charset-normalizer==2.1.1,docutils==0.18.1,idna==3.4,imagesize==1.4.1,iniconfig==1.1.1,Jinja2==3.0.3,MarkupSafe==2.1.1,packaging==21.3,pluggy==1.0.0,py==1.11.0,Pygments==2.13.0,pyparsing==3.0.9,pytest==7.1.3,pytz==2022.4,readthedocs-sphinx-ext==2.1.9,requests==2.28.1,six==1.16.0,snowballstemmer==2.2.0,Sphinx==5.1.1,sphinx-rtd-theme @ file:///home/circleci/project/.tox/.tmp/package/1/sphinx_rtd_theme-1.0.1a1.zip,sphinxcontrib-applehelp==1.0.2,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==2.0.0,sphinxcontrib-httpdomain==1.8.0,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.5,tomli==2.0.1,urllib3==1.26.12
```

#1348 adds missing Sphinx versions to the test matrix, i.e. currently only Python 3.10 environment has Sphinx 5.x.

benjaoming:
Trying to move things forwards, so I'm going to propose the following assessment and if you agree @agjohnson I think we should merge both this PR and https://github.com/readthedocs/sphinx_rtd_theme/pull/1336 (because the reasoning for allowing docutils 0.19 is the same as below)

1. sphinx_rtd_theme has an upper bound on docutils because of previous fallout, see: https://blog.readthedocs.com/build-errors-docutils-0-18/
2. Previous incidents following the docutils 0.18 release are now largely handled since Sphinx is ultimately responsible for upper bounds on docutils - patched versions of Sphinx was released for 1.8.x and 2.4.x series, and subsequent Sphinx release series have upper bounds on docutils.
3. Going into the future, sphinx_rtd_theme _might_ maintain an upper bound on docutils since it may affect the HTML structure that Sphinx generates and thus what the theme is compatible with. We should discuss to completely remove our docutils dependency rely on the Sphinx dependency entirely in a separate issue.

Preferred solution: We bump the upper dependency and merge this PR as-is, same with https://github.com/readthedocs/sphinx_rtd_theme/pull/1336

### What happens then...

The upgrade to `docutils<0.19` may negatively impact deployments that have pinned some variant of `sphinx<1.8.6`, `Sphinx>=2.0,<2.4.5`.

**Example breakage:** A deployment with `sphinx==1.8.5` (this version has no docutils upper bound and is incompatible with docutils 1.8) and `sphinx_rtd_theme` (no version pinning) will get docutils 0.18 and then builds will break.

**Mitigation?** It's possible that we release sphinx_rtd_theme with `sphinx>=1.8.6,sphinx!=2.0.*,sphinx!=2.1.*,sphinx!=2.2.*,sphinx!=2.3.*,sphinx!=2.4.0,sphinx!=2.4.1,sphinx!=2.4.2,sphinx!=2.4.3,sphinx!=2.4.4` -- however, since the pip resolver was changed and had issues with endless backtracking, I'm not sure if this is the way forward? It will also introduce a very complicated pattern that will make our future dependency headaches worse.

**Restating the benefits of bumping docutils version:** This change is necessary (i.e. critical) to introduce because several projects have failing builds because of failure to satisfy dependencies. Docutils 0.18 is soon 1 year old, so it's due time to support it.

Refs: https://github.com/sphinx-doc/sphinx/issues/9807

twodrops:
> @twodrops
> 
> > I wanted to test this branch with
> > pip install git+https://github.com/edgarrmondragon/sphinx_rtd_theme.git@docutils-0.18
> > but somehow the docutils seems to be still set as < docutils 0.18
> 
> Can you verify the version of Sphinx and Python that was used for this?

@benjaoming I tested it with your branch again and I cannot reproduce it anymore. Sorry for the confusion.
Looking forward to the quick merge of this PR which has been blocking us for a while now.

benjaoming:
No worries! Thanks for weighing in! Looking forwards to resolving this!

agjohnson (changes requested):
Thanks again for the contribution! Unfortunately, this PR reintroduces the display issues with footnotes/citations that caused us to pin `docutils<0.18` initially:

https://sphinx-rtd-theme--1304.org.readthedocs.build/en/1304/demo/demo.html#references

image

Until we can address those in a backwards compatible way, `docutils<0.18` is the specifier we should use.

agjohnson:
A couple more notes:

> I'm going to propose the following assessment and if you agree @agjohnson I think we should merge both this PR

In the interest of sticking to our plan for faster release cadence, I don't think we should merge these just yet. These PRs will both require bug fixes for display bugs.

I noted on #1309 that the intention was to release 1.1 as a rather minimal release -- just pinning dependencies more accurately. We'd still try to iterate quicker and release 1.2/2.0 with new features like docutils 0.18/0.19 support however.

> Previous incidents following the docutils 0.18 release are now largely handled since Sphinx is ultimately responsible for upper bounds on docutils

Mostly, however our concerns are tighter than the concerns that Sphinx has. We have display issues from docutils to worry about as well, and have to support multiple versions of docutils :disappointed: 

> Going into the future, sphinx_rtd_theme might maintain an upper bound on docutils

Because of above, we can say fairly concretely that we'll always have docutils specifiers in both directions.

> Proposed solution: We bump the upper bound for docutils, i.e. merge this PR as-is, same with https://github.com/readthedocs/sphinx_rtd_theme/pull/1336

For next release, :+1: this sounds good. However, this is going to be more like "We bump the upper bounds for docutils to 0.19 and fix all display bugs from docutils 0.18 and docutils 0.19 changes."

> How many projects are affected?

This list is probably even more limited than you describe, as the project would also have to pin `docutils` for there to be a dependency conflict. As its a transitive dependency, this isn't common. I suppose more-so, the project would have to pin an impossible dependency as well, `docutils==0.17.1` will not introduce a build failure.

> Mitigation?

We can probably be rather relaxed here, as it's only projects that have `docutils` pinned that we really need to worry about. We also have some deprecations to enforce, including old release of Sphinx, so we can take the position that Sphinx 1.8 support is phased out enough that we aren't going to worry about supporting modern docutils and crust Sphinx combinations. That is actually the point of releases like 1.1, to give a snapshot that users *can* use before we start culling our dependencies.

benjaoming:
Going to look for more references about this issue - I would like to see the discussion about why docutils uses `<aside>` instead of `<dl>`, it seems wrong.

benjaoming:
@pradyunsg re: https://github.com/pradyunsg/furo/discussions/447#discussioncomment-2863617 I see that you ran into this and already fixed it in Furo. Do you have a quick take on your decisions here? Did you drop support for docutils 0.17 at the same time?

pradyunsg:
https://github.com/pradyunsg/furo/commit/d955850a89310a932ecab4e84cba0c20371a6a33

I tweaked the look, preserved docutils 0.17 compat, and filed https://github.com/sphinx-doc/sphinx/issues/10531.

I’ve not had bandwidth to tweak the styles in Furo since then, but it seems like it’ll work just fine to use the css in docutils with a tweak.

benjaoming:
Thanks for getting things improved with Sphinx @pradyunsg - it's really nice that the docutils 0.19 behavior got backported to Sphinx 5.1, I think we can reduce the scope of footnote DOM variants to just 2 versions.. the old `<dl>` and the new `<aside>` from docutils 0.19.

benjaoming:
@agjohnson the visual regression is now addressed in https://github.com/readthedocs/sphinx_rtd_theme/pull/1351

benjaoming:
Support for docutils 0.18 has been added in #1381 

In release notes, we might add some clarity on old versions of Sphinx potentially breaking because of docutils 0.18, see comment in https://github.com/readthedocs/sphinx_rtd_theme/pull/1304#issuecomment-1266924478

benjaoming:
Released in 1.2.0rc1 :+1:

kloczek:
In mean time `docutils` 0.19 has been released 😋 
It would be good to move to that new version as well 😄

benjaoming:
@kloczek - definitely, and if we can get some help with verification and testing, we might also be able to add the support quite quickly - the release notes indicate that they have changed the DOM for footnotes _again_ => https://docutils.sourceforge.io/RELEASE-NOTES.html#release-0-19-2022-07-05

benjaoming:
There is an issue open for docutils 0.19 support here: #1323

kloczek:
Thank you for update 👍 😄

Issue #1311, opened 2022-08-17
Updating node-sass


benjaoming:
node-sass 4.3.0 is super-old and uses a stack for Python 2. Since then, a bunch of newer support has landed for both Node and Python versions.

I suggest updating to 5.0.0, since prior releases just seem to be about very small bug fixes and support of many newer Node versions: https://github.com/sass/node-sass/releases

PR incoming...

benjaoming:
Same issue with bourbon and bourbon-neat.

benjaoming:
Oh game over, looks like we are stuck on Python 2 because of Wyrm.

```
bourbon-neat 1.9 is required, Wyrm is not compatible with Neat 2.0+.
The expected selector for the bourbon-neat dependency in package.json is "~1.9".
```

agjohnson:
Python 3 is definitely usable, but you cannot upgrade SASS, Wyrm, or Bourbon.

Upgrading these is a known issue, and the fix so far has be to drop the entire dependency chain and switch to Bootstrap. I would avoid trying to change these dependencies, as there was a fair amount of work that went into finding the correct configuration of dependencies that worked

agjohnson:
There is a good amount of information on the dependencies here:
https://github.com/readthedocs/sphinx_rtd_theme/pull/771

benjaoming:
Are we talking about the same thing? The Node stack needs Python 2 in order to work AFAICT. There are for instance `print "foobar"` statements in the executed code of the node packages, which fail hard.

agjohnson:
Yeah, I'm not quite sure where you are running into errors, but perhaps having more information on how you're noticing errors would help.

The theme definitely builds on Python 3, I am using 3.8 locally only because I haven't tried upgrading to 3.11 yet, but I'm sure that also works.

Python 3 tests are all passing here:
https://github.com/readthedocs/sphinx_rtd_theme/pull/1316

![image](https://user-images.githubusercontent.com/1140183/185246949-d0a38146-061f-443b-b0ad-b2bb080507f2.png)

benjaoming:
There is something smelly going on.

While I have definitely had issues with a ghost Node 18 installation (which was supposed to be Node 14), I am now also having error-free builds with Python 3 only available in the Docker container.

node-gyp 3.8 (which we are using) is NOT supposed to work with Python 3. That support landed in version 5: https://github.com/nodejs/node-gyp/blob/main/CHANGELOG.md#v500-2019-06-13

[...]

PR #1312, opened 2022-08-17
Build assets in Docker container + use same container setup in CI

### LINKED ISSUE #1310: Broken bumpversion rule touches package-lock.json
The package-lock.json has become invalid since https://github.com/readthedocs/sphinx_rtd_theme/commit/1b4263eadfa5dc13473cfa8797c579a78b3c175e

benjaoming:
Fixes #1310


Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM node:14-alpine
+FROM python:2-alpine
```
agjohnson, line 2: This should be Python 3, we don't support Python 2 anymore.

docker-entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+cd /project
+
+npm run build
```
agjohnson, line 5: You should be running `npm run dev` for development, not `npm run build`. `build` is for build the JS/CSS files, where `dev` runs Webpack development server and takes care of hot reload of the assets.

docs/contributing.rst
```diff
@@ -62,6 +65,27 @@ can be used to test built assets:
...
 
+
+_dockerized-build::
+
+Dockerized build
+================
+
+If you have Docker available on your platform, you can get started building CSS and JS artifacts a bit faster and won't have to worry about any of the setup spilling over into your general environment.
+
+When building with Docker, we create an image containing the build dependencies, such as `SASS`_ , `Wyrm` and `Webpack`_ in the anticipated versions. Some of these are quite outdated and therefore ideal to isolate a container. The image is tagged as ``sphinx_rtd_theme:latest``.
```
agjohnson, line 76: This is true without Docker as well, everything is installed via NPM. Just to be clear to the contributor, I think this should be removed.

```suggestion
```

docs/contributing.rst
```diff
@@ -62,6 +65,27 @@ can be used to test built assets:
...
+
+Inside the running docker image, we mount the working copy of the repository, build the artifacts and finally copy out those artifacts into your external environment.
+
+Use the following steps:
+
+.. code-block:: console
+
+    $ make docker-images      # Builds an updated version of the docker image
+    $ make docker-run         # Runs the docker environment and builds the assets. The container exits after completing the build.
+    $ make docker-copy-assets # Copies out the assets from the most recent docker-run
```
agjohnson, line 86: This process doesn't quite feel ready for contributors, it's a bit more complicated than the normal development pattern. Is there a reason to not just mount this path as a volume?

Dockerfile
agjohnson, line 2: This should be Python 3, we don't want to support Python 2 anymore

.circleci/config.yml
```diff
@@ -17,8 +17,12 @@ commands:
...
     steps:
       - checkout
-      - run: npm ci
-      - run: npm run build
+      - setup_remote_docker
+      # - run: npm ci
+      - run: docker build -t sphinx_rtd_theme:latest .
+      - run: docker run --cidfile=.container_id --mount type=bind,source="${CIRCLE_WORKING_DIRECTORY}/src",target=/project/src,readonly sphinx_rtd_theme:latest
+      - run: docker cp "$(cat .container_id):/project/sphinx_rtd_theme" .
+      - run: docker cp "$(cat .container_id):/project/package-lock.json" .
```
agjohnson, line 25: Noted already, but the node orb should be used here instead of Docker.

agjohnson (commented):
I hesitate a bit about changing the development and packaging process this much. Is there a strong reason to switch CI and/or local builds to use Docker?

Are you hitting specific issues with local development otherwise? Perhaps it helps to raise those issues first, before changing things too much.

Generally speaking, giving contributors multiple conflicting paths for development is a bit confusing, so worth being careful here.

agjohnson:
Also, I should note that I raised issues on this PR, but I think there are simpler alternatives to this and it is probably more helpful to discuss some build/development issues before going too much further.

agjohnson:
I addressed issues with CI building in #1314, issues with a merge error in #1315, and dependency issues with #1316

Dockerfile
benjaoming: Please see #1311

agjohnson: Yeah, I already commented there as well. Python 3 is supported currently, Python 2 is not required and might be documented as not support at all even.

benjaoming: This is not true AFAICT. We are using a version of `node-sass` that uses a version of `node-gyp` that doesn't support Python 3. I finally hit the same issue in `bourbon-neat` and gave up, since Wyrm didn't allow an upgrade. See #1311

.circleci/config.yml
benjaoming: This is now replaced with a CircleCI env that builds our Docker image like we specify and it doesn't take a lot of build time :+1:

Dockerfile
agjohnson: The theme has been building on Python 3 for some time. This PR has passing builds for all Pythons: https://github.com/readthedocs/sphinx_rtd_theme/pull/1316

docs/contributing.rst
benjaoming: Since the Node version and Python versions expected for the build environment weren't previously made explicit, this comment felt very spot on :) I would always advocate having a Dockerfile in order to manifest the build environment. Not trying to make it mandatory, though, that's why it's just an additional development suggestion.

agjohnson: This seems like a change for the contributing doc instead. The docker environment isn't going to be any different than the current build environment.

Dockerfile
benjaoming: I haven't found an explanation for that TBH, but I'm sure it's good :) When I change the Dockerfile to use ` FROM python:3-alpine`, I start getting errors that I never resolved, but I really did try to do it without touching our node stack. The "internet" seems to have similar issues.

https://github.com/sass/node-sass/issues/2877
https://github.com/nodejs/node-gyp/issues/1687

benjaoming:
For the record: This approach worked fine and builds were :heavy_check_mark:. I'd like to contest that Dockerized build environments *are* the way forward :wink:

Dockerfile
agjohnson: Odd, the errors don't necessarily surprise me, but I certainly can't offer an explanation. `node-gyp` apparently has python 3 support since 4.0.0 -- I hit this issue back in #771 -- https://github.com/readthedocs/sphinx_rtd_theme/pull/771#issuecomment-512689735

agjohnson: We figured this out, seems Alpine is somehow using Node 18 and also cannot find the `/usr/bin/python3` binary (or Alpine's python package only installs a `/usr/bin/python` binary

agjohnson:
From chat, I think this PR should still continue as an additional piece to local development. This seemed close, but using Python3 and Node 14 would be issues to resolve.

I would still hold off on altering local development workflow or our CI builds though. Switching to Docker has side effects that require a bit more consideration, and we'd rather avoid disparate development workflows for now.

PR #1314, opened 2022-08-17
Unbreak CI config

agjohnson:
Build with Node 14 installed through orbs.

Fixes #1275

PR #967, opened 2020-06-26
Nicer styles for <kbd>

Blendify:
Fixes #1224

It seems #677 never actually got merged into master

PR #1315, opened 2022-08-17
Fix SASS syntax errors from #967

agjohnson:
This introduced syntax errors that were not picked up in review, the
`:not()` psuedo selector can't lead a selector in SASS.

Based on top of #1314, merge that first.


agjohnson:
Annnd these CI failures are from the broken Jinja release, I won't fix those in this PR. This PR was only addressing the `ci/circleci:build` failures.

benjaoming:
FYI, this was already on my radar :) https://github.com/readthedocs/sphinx_rtd_theme/pull/1312/commits/5d8fbbaf65a2ac15bd4870fb50a73c6289f26a1d

agjohnson:
Sure, but any reason that needs to be in such a large PR? Seems we could merge a few of these and isolate your PR to add a docker development environment.

benjaoming:
> Sure, but any reason that needs to be in such a large PR?

Nothing more or less than builds not working, so kind of hard as a new contributor to reach the bar of :heavy_check_mark:

PR #1318, opened 2022-08-19
Bump to a valid version scheme, bump node-sass for Node 14 support

benjaoming:
* [x] Add package-lock.json...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sphinx_rtd_theme",
   "main": "js/theme.js",
-  "version": "1.0.1alpha1",
+  "version": "1.1.0-alpha.0",
   "scripts": {
     "dev": "webpack-dev-server --open --config webpack.dev.js",
     "build": "webpack --config webpack.prod.js",
@@ -30,7 +30,7 @@
     "jquery": "^3.6.0",
     "lato-font": "^3.0.0",
     "mini-css-extract-plugin": "^0.6.0",
-    "node-sass": "^4.13.1",
+    "node-sass": "^4.14.1",
     "optimize-css-assets-webpack-plugin": "^5.0.4",
     "roboto-fontface": "^0.10.0",
     "sass-loader": "^7.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sphinx_rtd_theme",
   "main": "js/theme.js",
-  "version": "1.0.1alpha1",
+  "version": "1.1.0-alpha.0",
```
agjohnson, line 4: Bumpversion and bumpver will put preference on the existing versioning pattern, I believe, so this will likely be overwritten. Bumpversion and bumpver follow PEP 440, which does not use the optional period in between:

`1.0.1alpha1`
`1.0.1alpha2`

https://peps.python.org/pep-0440/#pre-releases

I'd say lets stick with the pattern described in PEP440, both are equally valid as far as I know.

package.json
```diff
@@ -30,7 +30,7 @@
     "jquery": "^3.6.0",
     "lato-font": "^3.0.0",
     "mini-css-extract-plugin": "^0.6.0",
-    "node-sass": "^4.13.1",
+    "node-sass": "^4.14.1",
```
agjohnson, line 33: I haven't tested this one yet, so can't say what to look for other than build failures. Seems okay, but probably worth doing a thorough check on style output.

package.json
agjohnson: Also, setuptools normalizes this automatically anyways:

```
/home/anthony/secure/dev/readthedocs/theme/.direnv/python-3.8.12/lib/python3.8/site-packages/setuptools/dist.py:463: UserWarning: Normalizing '1.0.1alpha1' to '1.0.1a1'
  warnings.warn(tmpl.format(**locals()))
```

agjohnson: Also note, changing this string makes bumpversion not work, unfortunately, as it won't try to change this string anymore

benjaoming: We'll resolve this in #1326

PR #1319, opened 2022-08-19
Supplemental development process for Docker

benjaoming:
Updated the approach from #1312 

* Docker environment no longer used in CircleCI
* Added support for using `npm run dev` by proxying commands to `make docker-run command=<npm-command>`
* Documentation updated
* Found an approach to install Python package in editable mode without writing to the external environment (that's the intention!)


agjohnson:
So, similar to the site-community change to the workflow, isn't this what you'd use docker compose for? Seems this is reproducing much of what that gives us already -- defined volumes, ports, image building.

docs/contributing.rst
```diff
@@ -62,6 +65,39 @@ can be used to test built assets:
...
 
+
+_dockerized-build::
+
+Dockerized development
+======================
+
+If you have Docker available on your platform, you can get started building CSS and JS artifacts a bit faster and won't have to worry about any of the setup spilling over into your general environment.
+
+When building with Docker, we create an image containing the build dependencies, such as `SASS`_ , `Wyrm` and `Webpack`_ in the anticipated versions. Some of these are quite outdated and therefore ideal to isolate a container. The image is tagged as ``sphinx_rtd_theme:latest``.
```
agjohnson, line 76: Noted on the last PR, this seems confusing as SASS, Wyrm, and Webpack are all pinned by the package dependencies. Docker is not doing anything special here.

benjaoming: Removed it :+1:

benjaoming:
> So, similar to the site-community change to the workflow, isn't this what you'd use docker compose for? Seems this is reproducing much of what that gives us already -- defined volumes, ports, image building.

Yes, I have noted the choice several places: Either you have docker-compose as a dependency or you write a rather long `docker run` command. I was happy about this one in particular because I didn't have to add docker-compose in CI (when I was doing that). But it also saves at least a line of text in the instructions for setting this up :)

agjohnson:
Aye. We're already using docker compose in other repos, so seems that should be safe to include as a dependency here too. A homebrew approach is almost certainly bound to cause more problems than introducing docker compose would cause, and docker compose is way more approachable.

I still don't think we want to go as far as to worry about docker in CI, as we aren't using that pattern in any of our repos.

benjaoming:
> Aye. We're already using docker compose in other repos

I disagree. This repo is much more public-facing.

benjaoming:
Anyways, this works, and since you're not using Docker in your development environment and this is just a supplemental change, I would like to conclude it without adding docker-compose. I'm totally open to adding it later :+1:  But perhaps we can wait and see if there's a CI integration coming up as well :crystal_ball:

agjohnson:
> I disagree. This repo is much more public-facing.

Regardless of who is using it, I don't think asking for docker-compose is that high of a barrier. I would say it's a much lower barrier, as I can work with docker compose, but would not feel comfortable mucking around with a homebrew replacement for docker compose.

Dockerfile
```diff
@@ -0,0 +1,52 @@
+# python:3-alpine contains node 18 so has to go first
+# in order to get overwritten
+FROM python:3-alpine
+FROM node:14-alpine
```
agjohnson, line 4: And from a similar PR on site-community: the feedback there, which I shared as well, is that this would probably be better off as a Ubuntu image, as the team has more experience with Ubuntu and it's our standard elsewhere.

The double `FROM` is also a point that came up in the other pull request. I haven't seen this either. From the docs, it's not really clear if this creates a single image or multiple images:

> FROM can appear multiple times within a single Dockerfile to create multiple images or use one build stage as a dependency for another. Simply make a note of the last image ID output by the commit before each new FROM instruction. Each FROM instruction clears any state created by previous instructions.

Makefile
```diff
@@ -0,0 +1,19 @@
...
+                --network host \
+		--mount type=bind,source="$(CWD)",target=/project-readonly,readonly \
+		--mount type=volume,dst=/project-readonly/sphinx_rtd_theme.egg-info,volume-driver=local \
+		--mount type=bind,source="$(CWD)/src",target=/project/src,readonly  \
+		--mount type=bind,source="$(CWD)/docs",target=/project/docs \
+		sphinx_rtd_theme:latest $(command)
+
+docker-copy-assets:
+	docker cp "$(shell cat .container_id):/project/sphinx_rtd_theme" .
+	docker cp "$(shell cat .container_id):/project/package-lock.json" .
```
agjohnson, line 19: And discussed elsewhere already, but just noting here, this homebrew configuration seems to overlap with what docker-compose provides. I'd put some preference on using standard tooling, as it's more approachable.

@humitos would have the most input/opinions here

Dockerfile
benjaoming: I agree, the multiple `FROM` is not desirable.

I'll try something based on a minimal Ubuntu...

agjohnson: If the double FROM works, that's okay, I think we're just not familiar enough to :+1: it

benjaoming: Hm, the Ubuntu image will require either running `su -c 'curl -sL https://deb.nodesource.com/setup_14.x | bash -'` or otherwise some NVM stuff. I think it's better to use the offical Node images and put Python 3 in them. They also have Debian images.

Makefile
benjaoming: I don't know anything about `homebrew`, it's a MacOS thing right?

benjaoming: The idea of `docker cp` is to avoid a write operation on a bind mount because file system permissions are usually troublesome, so it can be better to be explicit about what you expect to get from the docker runtime.

I'm already fixing this on your request @agjohnson

Dockerfile
agjohnson: Seems like a good place to use `asdf` actually.

Makefile
benjaoming: This is now removed, the docker-compose command mounts the current git repo and builds inside that one.

Dockerfile
benjaoming: Understanding now how multiple `FROM` commands work! The former pattern was wrong, so need to be more explicit. Have included a comment for now that 3.10 is implied by `node:14-alpine`. Will figure out something that's easier to maintain.

benjaoming:
This additions is high priority for working through the remaining PRs for 1.1 to be fast-tracked. Since it's an optional and supplemental part of the development/build process, I'll merge it.

I'm quite sure that I'll be back with some adjustments, since this isn't really perfect.

@humitos as you can see from the current setup, there's a bit of simplification work left in the separation logic between docker image, docker runtime and docker host. I'll happily introduce this on a call :)

https://github.com/readthedocs/sphinx_rtd_theme/blob/ba059c7970ad71fe5bcbf1e358b74b52474b29e0/Makefile#L7-L13

Makefile
agjohnson: > I don't know anything about homebrew, it's a MacOS thing right?

I meant this in the sense that this is a homebrewed process -- it is a manual, one-off creation, as opposed to a standardized, repeatable process using docker-compose

Dockerfile
```diff
@@ -0,0 +1,58 @@
+# This implicitely includes Python 3.10
+FROM node:14-alpine
```
humitos, line 2: I'd prefer to use Ubuntu here if possible. Mainly to keep consistency with the rest of the repositories we have. Also, because I've found different weird compilation issues when working with Alpine and the saving in space is not really huge.

Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+    npm audit fix &&\
+    npm install
+
+# This is strictly speaking not necessary, just makes
+# running the container faster...
+# Install dependencies, then uninstall project itself
+COPY setup.py README.rst /project-minimal-copy/
+RUN cd /project-minimal-copy &&\
+    pip install ".[dev]" &&\
+    /usr/bin/yes | pip uninstall sphinx_rtd_theme
```
humitos, line 46: Nice trick! 😄

Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+
+
+# Copy in files that we need to run the project. These files
+# will not be mounted into the runtime from external environment
+# so we copy them during build.
+COPY webpack.common.js webpack.dev.js webpack.prod.js /project/
```
humitos, line 52: Is there a reason why we don't mount them via docker compose instead of copying them here?

Makefile
humitos: I'm not sure to understand what this step is for. Why are we copying these files from inside the container into the host?

Makefile
```diff
@@ -0,0 +1,18 @@
+SHELL := /bin/bash
+CWD := $(shell cd -P -- '$(shell dirname -- "$0")' && pwd -P)
+
+docker-images:
+	docker-compose build
+
+docker-npm-build:
+	rm -f .container_id
+	docker-compose run -d sphinx_rtd_theme build > .container_id
```
humitos, line 9: nit: there is no need to do `rm -f` since you are redirecting the output with `>` in the next command. If it exists, it will overwrite it with the new output.

docs/contributing.rst
```diff
@@ -62,6 +65,36 @@ can be used to test built assets:
...
+
+Use the following steps:
+
+.. code-block:: console
+
+    # Builds an updated version of the docker image
+    $ docker-compose build
+
+    # Runs the development webserver
+    $ docker-compose run sphinx_rtd_theme dev
```
humitos, line 88: Wouldn't be better to use `docker-compose up` here which is the standard and call whatever is required behind the scenes?

docs/contributing.rst
```diff
@@ -62,6 +65,36 @@ can be used to test built assets:
...
+    
+    # If you want to copy stuff out of the Docker environment, run this make
+    # target or read the actual Makefile to see what is going on.
+    # We suggest running this command every time that you want to quickly build
+    # new CSS/JS assets
+    $ make docker-build-all
```
humitos, line 94: We are using `invoke` to handle this type of extra commands in other repositories. I'm not opposite to use Makefile, tho, just saying 😄

humitos (commented):
I think this is great. I just asked a bunch of questions to understand the workflow a little more.

Dockerfile
benjaoming: You can mount stuff at runtime. This is build time, so we need to copy into the image. The goal is to have node packages built together with the image.. however, if I copy the ENTIRE repo, then Docker's cache will be invalid everytime I build the image, so I have to build it over and over again.. so I copy the minimal set of files.

docs/contributing.rst
agjohnson: I'd also prefer to have a singular pattern to use, as much as I like make.

Dockerfile
agjohnson: I think we're on the same page here and Ubuntu images would be the best way to go.

Dockerfile
agjohnson: I'm still not sure I follow the need for avoiding mounts. Could you clarify what issue are you trying to solve? There is probably prior art to reuse in our application Docker configuration, whether that is permissions issues or avoiding host filesystem artifacts.

I would look at what we do in the application docker processes, as we don't seem to have the issues that this docker config is trying to solve. I would put strong preference on reusing patterns that are working for us already.

benjaoming: > Could you clarify what issue are you trying to solve?

Issue is to create a reusable and containerized environment for development. As the development currently also has responsibility for building the production assets, then that's also solved, but I understand that we want to remove these artifacts from Git as soon as possible.

I'd like to propose that the development container specification can be reused for CI. I can open an issue for that.

> I would put strong preference on reusing patterns that are working for us already.

Or perhaps developing and improving those patterns? :)

The reason I put docker-compose with a bind mount was to echo what's in readthedocs.org, although I'd always recommend to keep the mounted set of files to an absolute minimum, especially so that files from the container aren't dumped on the host.

Dockerfile
benjaoming: From how I've been working with the project, it's been very helpful to have everything built in the container and just attempting updates at runtime.

I think the best container architecture is to have all dependencies built into the container so the entire containerized environment ideally is readonly at runtime.

Makefile
benjaoming: haha yes, will remove it.... 'twas because `docker run` could write to a file - now I need to detach in order to get the ID because `docker-compose run` doesn't have the same option that `docker run` has.

Makefile
benjaoming: I had to backtrack again from the approach where a host directory was mounted and written to by the container.

The reason was that I had issues with a mix of files owned by the user and files owned by root. I also needed to run `npm cache clear` and I never understood why.

But the fundamental reason was that I was mounting host system stuff into the container and writing irrelevant files there.

The best approach IMO is to keep everything containerized and copy out the artifacts directly from their known location. If you can separate out a bind mount and ensure that ONLY intended artifacts are written there, then that's also a fine approach.

docs/contributing.rst
benjaoming: `docker-compose run` is doing what `docker run` does but reading the docker-compose configuration - I start the service that I want and give an argument `dev` to the entrypoint. `docker-compose up` is a different type of command.

docs/contributing.rst
benjaoming: Containers provide a quick way to get started. `make` exists on most systems and doesn't sit on top of a large stack that's customized by the project.

With `invoke`, we'd have to firstly bootstrap a Python environment with the correct version of `invoke` - or as a new contributor understand what `invoke` even is. I think there's a better chance that people can copy-paste `make` commands without worrying about setting up an environment for `make`.

If the feeling is "let's not have Make", then I would rather have a shell script for running the `docker-build-all` target, since that's the only command that chains a few things together and should be scripted.

Dockerfile
agjohnson: > Issue is to create a reusable and containerized environment for development.

This seems like the overall *goal* of your change, but I'm more curious what *issue* you are hitting that requires this.

The main concern here is that we're making a complex solution for a problem we don't really have currently. The initial goal is to give contributors easy environment creation, which this setup does. But absolute isolation isn't currently a strong roadmap concern.

A longer term discussion point should be describing why isolation of build files is necessary for us. There are general arguments for this, but I'm not sure they are enough to bump up the priority of this work to an immediate team priority.

> I'd like to propose that the development container specification can be reused for CI

What issue are you hitting that requires this change? The reasons to stick with the current pattern are that it's already providing isolation and deterministic builds, and there is more prior art and information on using CircleCI without custom Docker images.

So switching this feels a bit like sideways progress, but might also lead to us spending more time scratching our heads.

> Or perhaps developing and improving those patterns?

It depends on the problems you are trying to solve, what design decision we come to, and where we place the work on our roadmap. The place to start changes like this is with discussion first, and I would suggest opening an issue on meta. From there, we'd find a place for the work on our roadmap if we all feel it's a priority.

Whatever we decide, sticking to a *singular* pattern is the main concern. One-off processes will add confusion and friction to maintenance.

humitos: @benjaoming 

> You can mount stuff at runtime. This is build time, so we need to copy into the image

I don't think there is need to copy files into the container that _are not used at build time_. If these files are used at run time, they should be mounted at run time. Otherwise, modifying any of these files would require rebuilding the Docker image when it's not if they are mounted.

Dockerfile
humitos: Yeap, I understand that and makes sense to me. I was referring to the usage of `/usr/bin/yes` for the `uninstall` command 👍🏼

Dockerfile
benjaoming: The files are used at build time, they are required for `package.json` and `npm install` to work. We could perhaps mock them.

humitos: Hrm... in that case, I don't follow it. These files are copied _after_ running `npm install` in the Dockerfile: https://github.com/readthedocs/sphinx_rtd_theme/pull/1319/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R36-R38

Also, the comment right above copying these files says they are required to run the project.

Following these two things, I'd think they are not required at build time.

Makefile
humitos: Gotcha! Yeah, creating files _inside_ the container that then are required _outside the container_ usually generates permission problems. The only way to do that correctly is mapping the UID/GID between the host _and_ the container and that is where things get more complicated. We've been there and I didn't enjoyed it 😄 

I like the pattern you used here with `docker cp` for this requirement, where files generated inside the container are required to be distributed/used outside the container.

We currently have this problem when creating migrations on Read the Docs, for example, where the resulting `.py` files is owned by root and left in a directory owned by "your user" in the host. Then you have to run `sudo chown` to fix the permissions on that file.

Dockerfile
benjaoming: I'm giving some pretty bad off-hand answers right now. I'll have to go over this again and give a proper answer. As I said, the idea is to avoid having anything mounted into the runtime environment to avoid making a mess in the host system's files.

I think this is easier to go through over a meeting than async.

docs/contributing.rst
agjohnson: I'd still agree that we should stick with a single pattern and use Invoke here. Both Python and Invoke are a normal part of our development workflow and already in use in other repositories, so are reasonable requirements for contributors. The goal is not to make this repository universally accessible as possible, it's just to make a repeatable environment that is easy for the core team to maintain.

PR #1320, opened 2022-08-19
Fix <kbd> CSS selector

### LINKED ISSUE #1224: <kbd> tag doesn't style as expected
### Problem

As was raised in this issue: https://github.com/readthedocs/sphinx_rtd_theme/issues/669 it's nice if the \<kbd> tag has better styling to indicate that it is intended to represent a button, not just code.

My understanding of it was that this should have been fixed in this pull request https://github.com/readthedocs/sphinx_rtd_theme/pull/677 but that does not appear to be the case. I can not find the code changed in the PR in the current iteration of the `src/sass/_theme_rst.sass`, nor does it seem to work in my project.

### Reproducible Project

```rst
To confirm the selection, hit :kbd:`ENTER`. To go back to the base mode, hit either :kbd:`ESC` or :kbd:`Q`.
```

#### Error Logs/Results

Here is the output of the above:

Confirm selection

#### Expected Results

I know nothing about sass, but I monkeypatched it with the little custom css I have, basing myself on the change in the linked PR with the following css:

```css
kbd, .kbd {
    color: inherit;
    font-size: 80%;
    background-color: #F0F0F0;
    border: 1px solid #D5D5D5;
    border-radius: 4px;
    box-shadow: grey 0px 2px;
    padding: 2.4px 6px;
    margin: auto 0;
}
```

Which produces the following output instead:

Confirm selection

I'm not sure that's quite right still, but it's a lot closer to the expected output.

### Environment Info

- Python Version: 3.8
- Sphinx Version: 4.0.1
- RTD Theme Version: 0.5.1

benjaoming:
Apparently, this change is a bit cursed. Tested that this works finally :crossed_fingers: 

image

Hopefully, this finally fixes #1224


agjohnson (approved):
Ah yes, this makes more sense.

PR #1329, opened 2022-08-24
Revert change to JS package

### LINKED ISSUE #1326: Align bumpversion version strings
Need to fix version string discrepancies between package.json and setup.py such that bumpversion can automatically search and replace

Subsequently open up an issue about later Node versions breaking on the package.json version scheme.

agjohnson:
This matches the version number in the Python package, and allows
bumpversion to continue working. We can revisit changing this version
later, but currently this version will never be updated again, as
bumpversion will not match this string anymore.

Fixes #1326

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "sphinx_rtd_theme",
-  "version": "1.1.0-alpha.0",
+  "version": "1.0.1alpha1",
   "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


benjaoming (approved):
Sounds good! The reason of the change was because the version syntax was invalid for later versions of node. See: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#version

PR #1330, opened 2022-08-24
Tune kbd selectors

agjohnson:
Exclude option lists from selectors for kbd styling.

In option list, this doesn't style the `kbd` elements:

image

Outside option lists, it does:

image

Untested across Sphinx versions and HTML4/5 writers though.


src/sass/_theme_rst.sass
```diff
@@ -520,10 +520,8 @@
     border-radius: $base-line-height / 6
     padding: ($base-line-height / 10) ($base-line-height / 4)
     margin: auto ($base-line-height / 12)
-  *:not(kbd) > kbd,
-  *:not(kbd) > .kbd,
-  *:not(.kbd) > kbd,
-  *:not(.kbd) > .kbd
+  *:not(dl.option-list) > :not(dt):not(kbd):not(.kbd) > kbd,
+  *:not(dl.option-list) > :not(dt):not(kbd):not(.kbd) > .kbd
```
benjaoming, line 524: This will not work for `.rst-content > p > kbd` ?

benjaoming: Maybe that doesn't matter since everything is inside a `section`

benjaoming: Okay there are so many nested elements that `.rst-content > p > kbd` seems completely irrelevant

benjaoming (approved):
LGTM! Thanks for fixing this!

src/sass/_theme_rst.sass
agjohnson: Yeah, it seems to work fine here at least:

image

Without two/three levels of descendants between `.rst-content` and `kbd`, it could not match though. I can't say for certain that is a potential output, but there should be enough elements in between to avoid that scenario.

PR #1331, opened 2022-08-24
Work-around for .option-list regression of <kbd> styling

benjaoming:
After #1330, we have an issue that `.option-list` is affected. It's not too bad, and we cannot really create a more sophisticated negative rule for a parent path... because of logic :)

So either we create an inverse kind of positive rule for `.option-list dt kbd` or we accept that `<kbd>` doesn't style inside a dt.

Before:
image

After:
image


agjohnson:
> and we cannot really create a more sophisticated negative rule for a parent path... because of logic

I gave an example of this in #1330

PR #1335, opened 2022-08-24
Fixup a wrong merge conflict resolution from #1321

benjaoming:
Looking at #1321, I made a mistake handling a merge conflict.

* This re-removes stuff that is already in setup.cfg.
* Edit: This also re-adds the `python_requires` option to setup.cfg which we should keep having in order for #1116 to be a neutral change.

Originally fixed in #1116


setup.py
```diff
@@ -93,60 +93,4 @@ def run(self):
...
-        'static/js/*.js',
-    ]},
-    include_package_data=True,
-    # See http://www.sphinx-doc.org/en/stable/theming.html#distribute-your-theme-as-a-python-package
-    entry_points = {
-        'sphinx.html_themes': [
-            'sphinx_rtd_theme = sphinx_rtd_theme',
-        ]
-    },
-    python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
```
benjaoming, line 112: Great to get rid of this btw.

benjaoming (commented):
Since this is a technical change, and I've just cross-checked that everything removed here is still in setup.cfg, I will take responsibility.... together with the build verification :heavy_check_mark:

setup.py
benjaoming: I take that back. This option is entirely great and we should keep having it in setup.cfg. So I've added it again. This should also ensure that we don't have regressions from #1116

PR #1349, opened 2022-10-05
Additions to Tox and Circle CI matrix

benjaoming:
After #1348, we should adjust expected Tox environments. Sphinx 5.2 was missing in 2 places.

Update: Sphinx 4.2, 4.3, 4.4 and 4.5 also added.

The Tox environment list is mostly for reference, as the Circle CI configuration _is_ the applied test matrix.

**Reviewing note**

You can look at the build outputs from Circle CI to validate the new test matrix.

Background: The environment lists are long and annoying, here's the story short: Since 1.0.0 was released, a lot of Sphinx releases have come out and the list was incomplete. Moreover, the Circle CI configuration would simply launch Tox with its default environments, and this was change to be explicit for each Python version to cut test times.


benjaoming:
Learned that Sphinx 4.2, 4.3, 4.4 and 4.5 are missing in the test matrix.

benjaoming:
@agjohnson I got a bit worried of what might slip through the cracks in this or any subsequent patch releases for 1.1.x, so I'd appreciate getting this extra "safety net" included.

@humitos could you have a quick look here?

tox.ini
```diff
@@ -35,8 +37,12 @@ deps =
     sphinx50: Sphinx>=5.0,<5.1
     sphinx51: Sphinx>=5.1,<5.2
     sphinx52: Sphinx>=5.2,<5.3
+    # All these Sphinx versions actually break since docutils 0.18, so we need to add this upper bound
+    # Projects using these Sphinx versions will have to do the same
+    # See: https://github.com/readthedocs/sphinx_rtd_theme/pull/1304
     sphinx{16,17,18,20,21,22,23,24,30,31,32,33,34,35,40,41,42}: docutils<0.18
     sphinxlatest: Sphinx
+    sphinxdev: https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip
```
benjaoming, line 45: This on isn't included, but it's fun to see it fail :upside_down_face:

tox.ini
```diff
@@ -1,8 +1,10 @@
 [tox]
 envlist =
-    py{2,27}-sphinx{16,17,18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,43,44,45,50,51}{-html4,-html5,}{-qa,}
-    py{3,36,37,38,39}-sphinx{16,17,18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,43,44,45,50,51,latest}{-html4,-html5,}{-qa,}
-    py{310}-sphinx{42,43,44,45,50,51,latest}{-html4,-html5,}{-qa,}
+    # Python 2.7 support was removed in Sphinx 2
+    py{2,27}-sphinx{16,17,18}{-html4,-html5,}{-qa,}
+    py{3,36,37,38,39}-sphinx{16,17,18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,43,44,45,50,51,52,latest}{-html4,-html5,}{-qa,}
```
humitos, line 5: Why do we have `py3` and `py3x`? What py3 means in this context? (the same question applies for `py2`)

humitos (approved):

tox.ini
benjaoming: I actually have no idea! I left it there from earlier. I could imagine it means "run whatever is default on the system".

But it doesn't make any sense in the Circle CI setup, nor if running Tox locally, I'll remove it. Thanks for noticing :+1:

PR #1350, opened 2022-10-05
Eager upgrade of all dependencies in docker development environment

benjaoming:
This helped me understand issues caused by jumping between old branches. The eager upgrade is over in a matter of ~6 second in my end, so I think it's worth it. Because of the ability to run the NPM development server, it's possible to skip this if working on SASS or docs/.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,7 +2,7 @@
 
 # Update latest Python dependencies in case they have changed
 cd /project-readonly
-pip install --upgrade -e ".[dev]"
+pip install --upgrade --upgrade-strategy eager -e ".[dev]"
 
 # This helps a potential permission issue, but might be removed
 # pending some more investigation of docker host file system


benjaoming:
CC: @humitos just wanted flag you on this one in case you think the change is interesting for other purposes... or problematic :)

humitos (approved):
I think this is fine and I don't think it will be problematic for this particular case.

One thing that I'd like to see added here is a comment in this file explaining why we added `eager` and what problem it solves. This will avoid confusions in the future.

Nekimaliizgrada:
6th man

humitos:
@benjaoming is this PR ready to be merged? what's missing?

benjaoming:
In itself, the PR is fine so will merge :+1:. I have some other refinements for the Docker setup stashed. I think they will be on-topic once we start building the 2.0 release.

Issue #1309, opened 2022-08-16
Release 1.1 with upper bounds on dependencies


agjohnson:
The goal is to get 1.1 out that doesn't break when Sphinx 6 or docutils are released. The existing 1.1 milestone has a lot of extra bugs/bug fixes, which would be best off in a 1.2 release.

agjohnson:
I've repaired a number of issues building at CI. #1316 has working CI builds.

benjaoming:
I don't quite understand this issue, can you elaborate @agjohnson ? :pray:

agjohnson:
Pinged in chat. The note here is that we might delay a release if we continue going through the 1.1 milestone bugfixes. We'll want to do some testing on latest Sphinx and docutils, and ensure those pins do not introduce any display changes, build errors, or otherwise break the dependency chain for users.

benjaoming:
By pinning, do you mean introducing upper bounds `<` on more of the transient dependencies, i.e. not `==`?

agjohnson:
The current specifier for Sphinx is https://github.com/readthedocs/sphinx_rtd_theme/blob/master/setup.cfg#L35, so really wide open. We'll want to tune that down to at least avoid bringing in Sphinx 6 and docutils versions we can't support.

benjaoming:
Ah okay, then I get you, I understanding "pinning" as `==`

agjohnson:
Ah I see, yeah we use that term fairly loosely I suppose

benjaoming:
_Not sure if correct language_

benjaoming:
@agjohnson these are our current direct dependencies:

https://github.com/readthedocs/sphinx_rtd_theme/blob/e741681526e37bf8069c8864bbee9a5d754bc991/setup.cfg#L34-L37

small note: we'll bump docutils to 0.18 and maybe also to 0.19

Should we investigate transitive dependencies and potential of putting upper bounds of some of those or can this issue be considered solved?

agjohnson:
We'll use this issue to track an actual release going out. We can reassign this issue while you're away, but the actual work behind the release is indeed looking close.

agjohnson:
> these are our current direct dependencies:

@benjaoming To clarify, those are the current dependencies specified in `master`. Release `0.5.2` still has `docutils<0.17`, and no upper bounds on `Sphinx`:

https://github.com/readthedocs/sphinx_rtd_theme/blob/0.5.2/setup.py#L117-L120

[...]

PR #1352, opened 2022-10-05
Release 1.1.0b1 + 1.1.0b2 (bump versions with bump2version)

benjaoming:
Refs: #1309 

@agjohnson it seems that bump2version reformats setup.cfg -- we are perhaps using this for the first time since last release. I think it looks fine.

Once this PR is approved, I would complete these items:

1. [ ] Release on PyPy
2. [ ] Tag the released commit
3. [ ] Merge PR

I'm not - however - entirely sure of the tagging process since the `bumpversion` section in setup.cfg also seems to want a `tag` and `commit` value, creating a bit of chicken/egg situation.

Maybe that's not too important now, but I'd like to go over the entire release process for 1.2, splitting it up in a preparation check list (I like things that can be copy-pasted into issues or PRs) and a PyPi+version bumping section.


agjohnson (approved):
This looks correct :+1:

However, I don't remember bump2version reformatting setup.cfg. It should just be doing a string replacement :shrug: 

bumpver does not do this though, and we've settled on using that instead anyways. We can deal with that some point in the future though.

benjaoming:
> However, I don't remember bump2version reformatting setup.cfg. It should just be doing a string replacement

We migrated a lot of package properties from setup.py to setup.cfg, not sure if that's related.. bump2version wants to store data in setup.cfg, so most likely it parses and rewrites the file. Possibly it's because of that... or maybe my local ini parser :shrug: I found an issue about bump2version that indicated that setup.cfg support was added later, so maybe it's not fully mature. I think we should do the pyproject.toml thing eventually, so I can definitely live with this :)

benjaoming:
@agjohnson I think I got the "alpha1" prefix from package.json, but looking at previous releases, we do align with PEP440, so I think we should continue like that. I also bumped to "b1" instead.

benjaoming:
Pardons. The first 1.1.0b1 release will have to go down as a mistake, I hadn't merged master into the release branch with latest changes.

Merging and rebuilding assets.

benjaoming:
Merging because released :+1:

PR #1354, opened 2022-10-06
Translation updates: Pull translations from transifex, push new sources

benjaoming:
Some empty languages were added, a neutral change to `.tx/config` --- since this was all generated with `python3 setup.py update_translations`, I think it's all good as is.


agjohnson (approved):
Looks good!

Issue #1358, opened 2022-10-12
1.1.0b2 - Jinja2 upper bound causes conflict in existing deployments


benjaoming:
The current pre-release will not fly with existing deployments that have already pinned the latest Jinja2 release.

Needs a decision:

* Release as is `Jinja2<3.1`
* Bump to `Jinja2<3.2` and do a new pre-release

agjohnson:
The upper bounds were added with #1316, and replicate the fix that we instructed users to use to resolve the issue on their projects (pin `Jinja2<3.1`).

However, there is nothing on the theme that requires a specific Jinja2 version. Perhaps this is the wrong place to apply this fix entirely.

I'd instead suggest removing the dependency entirely, and fix the errors in CI by tuning dependencies with tox/etc instead.

benjaoming:
Brilliant @agjohnson !

It's nice that we didn't have to take these steps. The simplest solution is almost always the right one :) => https://github.com/readthedocs/sphinx_rtd_theme/pull/1360

PR #1359, opened 2022-10-13
Replaces the Jinja2 dependecy with a meta package 🤯

benjaoming:
This is a proposal for fixing #1358 

Notice that the sphinx-rtd-theme-meta-jinja2-dependency actually exists.

We only need this extra dependency for Python 3, since Jinja2 dropped Python 2.7 in version 3.0.0, so this will suffice as the upper bound for all Python 2 deployments.


agjohnson:
I noted on #1358 another way to resolve the issue you noted. My preference would be dropping the dependency, or keeping `Jinja2<3.1`.

PR #1360, opened 2022-10-13
Remove Jinja2 dependency

### LINKED ISSUE #1358: 1.1.0b2 - Jinja2 upper bound causes conflict in existing deployments

benjaoming:
Adding this dependency would cause conflicts with existing projects.

The dependency is already injected for legacy projects hosted on Read the Docs. All other environments will have had to address the issue since it was addressed in https://github.com/sphinx-doc/sphinx/issues/10291.

Fixes #1358

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,11 +17,11 @@ Dependency Changes
 
 Many documentation projects depend on ``sphinx-rtd-theme`` without specifying a version of the theme (unpinned) while also depending on unpinned versions of Sphinx. The latest version of ``sphinx-rtd-theme`` ideally always supports the latest version of Sphinx, but this is now guaranteed.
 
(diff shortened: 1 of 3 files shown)


agjohnson (approved):
:+1: but you'll need to add the pin back on the tox configuration. Tests are failing with the original Jinja2 bug on older Sphinx releases

docs/changelog.rst
```diff
@@ -38,7 +38,8 @@ Fixes
 * Definition lists rendered wrongly in api docs (#1052)
 * Citation not styled properly (#1078)
 
-Minor changes
+
+Minor Changes
```
benjaoming, line 42: I made a minor change here :see_no_evil:

PR #1365, opened 2022-10-20
Docs: Add upgrade instructions and simplify conf.py

benjaoming:
This adds a bit of guidance to upgrading, specifically the paragraph about pre-releases that is necessary to announce a pre-release with a direct reference to this Upgrade section.

* I made some changes to fix broken cross-references and make it clear whether something is from the roadmap or changelog (because that initially caused an ambiguous reference)
* Removed `import sphinx_rtd_theme` from `conf.py` in installation instructions, I don't know why it's there. We can potentially investigate this together with https://github.com/readthedocs/readthedocs.org/pull/9654


docs/installing.rst
```diff
@@ -1,18 +1,19 @@
...
 
     $ pip install sphinx_rtd_theme
 
-In your ``conf.py`` file:
+In your Sphinx project's ``conf.py`` file, add ``sphinx_rtd_theme`` to the list of enabled extensions and as the active theme:
 
 .. code:: python
 
-    import sphinx_rtd_theme
-
```
benjaoming, line 15: By importing this module, nothing happens. It's not used in `conf.py` and nothing is executed in the module's `__init__.py`.

I'm 99.9% sure that this is here for some kind of legacy reason.

benjaoming: I also tested it multiple times. I haven't used the theme recently with this line of code.

docs/development.rst
```diff
@@ -94,15 +94,15 @@ version:
 
 .. _semantic versioning: http://semver.org/
 
-.. _release-1.0.0:
+.. _roadmap-release-1.0.0:
```
agjohnson, line 97: These may or may not be linked externally. If there isn't a strong reason for this change, it might be best to leave the refs as they are. Probably not a huge loss otherwise.

docs/installing.rst
agjohnson: I believe you are right, and import here was only required because the conf.py used to require manual manipulation of the theme path.

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+      See: https://github.com/readthedocs/readthedocs.org/pull/9654
+
+
+.. _howto_upgrade:
+
+How to upgrade
+--------------
+
+.. tip:: 
+    **Read the Docs users:** To know which version of ``sphinx-rtd-theme`` you are running, you can always check your latest build outputs. One or more steps in the documentation build will install python packages, one of which is ``sphinx-rtd-theme``.
```
agjohnson, line 51: Perhaps this note should go in RTD docs instead, this isn't specific to just this theme.

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+
+Adding ``sphinx-rtd-theme`` to your project's ``requirements.txt`` will make pip install the latest compatible version of the theme.
```
agjohnson, line 53: I like to be inclusive for project types when discussing dependency installation -- a good chunk of projects use packaging mechanisms besides requirements files.

```suggestion
Adding ``sphinx-rtd-theme`` to your project's dependencies will make pip install the latest compatible version of the theme.
```

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+
+If you want to test a **pre-release**, you need to be explicit about the version in ``requirements.txt``.
```
agjohnson, line 55:
```suggestion
If you want to test a **pre-release**, you need to be explicit about the version you specify.
```

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+Otherwise, pip will ignore pre-releases. Add for instance ``sphinx-rtd-theme==1.1.0b3`` to test a pre-release.
+
+.. tip::
+    We always recommend that you pin the version of Sphinx that your project is built with.
```
agjohnson, line 59:
```suggestion
    We always recommend that you pin the exact version of Sphinx that your project is built with.
```

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+    We'll make sure not to release sphinx-rtd-theme that installs together with a Sphinx theme that it is incompatible with.
```
agjohnson, line 60: I don't quite understand this sentence, could you clarify what you mean here? It seems reducing would probably help this sentence.

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+    
+    It can never be guaranteed that sphinx-rtd-theme is compatible with everything in your project.
+    If your project is complex or needs reproducible documentation builds, you should pin or put an upper bound on ``sphinx-rtd-theme``.
+    For instance, you could add this to ``requirements.txt``::
+    
+        # Ensure that we do not automatically include changes from the next major series
+        sphinx-rtd-theme>=1,<2
```
agjohnson, line 67: We have some great documentation here already that would be better to point to:
https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html

docs/installing.rst
benjaoming: It's one of the few packages that we install implicitly, so I think it's important to let people know about understanding when it's installed and how to inspect that.

docs/development.rst
benjaoming: I got the feeling that it was too magical when I noticed it was overlapping with the changelog: https://github.com/readthedocs/sphinx_rtd_theme/blame/master/docs/changelog.rst#L60

So we _could_ remove the roadmap entry and "move" it to the roadmap, but that doesn't guarantee direct anchor links to not break.

I think we should just be explicit that this is a roadmap label vs a changelog label.

docs/installing.rst
```diff
@@ -27,21 +28,54 @@ In your ``conf.py`` file:
...
+    **Read the Docs users:** To know which version of ``sphinx-rtd-theme`` you are running, you can always check your latest build outputs. One or more steps in the documentation build will install python packages, one of which is ``sphinx-rtd-theme``.
+
+Adding ``sphinx-rtd-theme`` to your project's dependencies will make pip install the latest compatible version of the theme.
+
+If you want to test a **pre-release**, you need to be explicit about the version you specify.
+Otherwise, pip will ignore pre-releases. Add for instance ``sphinx-rtd-theme==1.1.0b3`` to test a pre-release.
+
+.. tip::
+    We always recommend that you pin the exact version of Sphinx that your project is built with.
+    We'll make sure not to release sphinx-rtd-theme that installs together with a Sphinx theme that it is incompatible with.
```
benjaoming, line 60: True. It was wrongly phrased and I find it a bit tricky. Is the meaning better here?

```suggestion
    We won't release sphinx-rtd-theme without marking its compatibility with Sphinx. So if you do not pin `sphinx-rtd-theme` itself, you will always get the *latest compatible* release.
```

docs/installing.rst
benjaoming: Ah yes, I also ref'ed that in the announcement: https://github.com/readthedocs/blog/pull/184

I'll add a reference :+1:

docs/development.rst
agjohnson: Yeah, don't worry too much here. I can't recall what would have linked here, but at worst it's probably some old blog posts.

docs/installing.rst
agjohnson: Well, I meant to say that this content should be replaced with a link to the page we have authored. This is generic Python dependency management advice, so no specific to the theme, and pur guide on reproducible builds is thorough.

Most important, this is slightly at odds with what we recommend in the reproducible builds guide and we should probably just communicate to users one, single vision for reproducible builds.

docs/installing.rst
agjohnson: That is much clearer :+1:

docs/installing.rst
benjaoming: Is there already a consensus about treating the theme as all other dependencies?

The advice was intended as being specific to the theme... I would take into greater consideration to advice about the effects of exact pinning vs. loose pinning. The benefits of a loose pinning or no pinning at all are greater be. I was thinking that it's up to the reader to decide and take responsibility for what level of "reproducibility" they want.

agjohnson: Yup, we have consensus here. Our guidance is that guide that I linked, and is that users should always explicitly pin every dependency.

This is partly self-serving, as reproducible builds don't inflict support time when some random, backwards incompatible, transitive dependency puts out a release.

Users can do whatever they want to, but we prefer to communicate a _singular_ opinion here: always explicitly pin every one of your dependencies.

benjaoming: Sounds good

From experience, I disagree - but it's a good thing. I have benefited from not having to bump versions of the theme or keep an eye out for updates. I just always get the latest version in my builds, no errors or issues or efforts made.

benjaoming: In addition to the self-serving part.. asking people to declare versions is also asking them to be aware of new releases from the beginning. So that can plant a seed for more engagement with theme releases.

If we're gonna tackle this in the future, it could be by having a second section called "Non-reproducible builds". A thought for now.

On the self-serving part: We also don't benefit from having issues that are fixed in later theme releases and then having to instruct to people to upgrade.

Anyways, blah blah blah, will replace it with a link to the generic chapter :+1:

agjohnson: Yeah I get that. The middle ground between free for all and using a lock file is covered in that guide, using pip-tools, and seems like a great approach so far.

You can still upgrade easily to the latest sphinx_rtd_theme based on your specifiers. But it's not an automatic process. Either way, every version you're installing is explicitly defined in a requirements file, so there is no guessing which version you're installing.

benjaoming: This is fixed now :+1:

docs/installing.rst
benjaoming: The purpose of a "tip" box is to be able to have a bit of information that isn't weaved into the surrounding parts? Isn't this a good case for a tip?

> "Check one or more steps in your build" is probably lost on a user that is not familiar with RTD.

Hence, the tip box was clearly labeled for "Read the Docs users:" :)

I moved it to the bottom of the chapter, maybe that's better?

benjaoming:
@agjohnson can this be approved?

docs/installing.rst
agjohnson: The directions here now aren't specific enough to help users that don't already know what they are looking for -- what steps are they looking at, what output are they looking for, etc? There's a lot more here, but it's fairly out of scope for these docs.

benjaoming: It's not out of scope. See for instance: https://stackoverflow.com/a/74355734/405682

By giving people some pointers to understand the RTD build, they can solve issues like this. I'm gonna go ahead and improve the tip box :)

agjohnson: I haven't been talking about a pointer being out scope. Covering how to use Read the Docs in the theme docs *is* out of scope.

Again, the docs you are describing belong in the application docs, not here. Adding a pointer to relevant docs in our application is absolutely fine, that's not in question though.

docs/installing.rst
```diff
@@ -27,21 +28,51 @@ In your ``conf.py`` file:
...
+.. tip::
+    We recommend that you pin the version of Sphinx that your project is built with.
+    We won't release sphinx-rtd-theme without marking its compatibility with Sphinx. So if you do not pin ``sphinx-rtd-theme`` itself, you will always get the *latest compatible* release.
+    
+    More information is available in Read the Docs' documentation on :doc:`rtd:guides/reproducible-builds`.
+
+.. tip:: 
+    **Read the Docs users:** If you are upgrading a project created before October 20, 2020 (January 21, 2021 for Read the Docs for Business), it's a requirement to specify the version of ``sphinx-rtd-theme``. Otherwise, ``sphinx-rtd-theme`` is pinned to a lower version to address legacy issues.
+    More information is available in Read the Docs' documentation on :doc:`rtd:build-default-versions`.
+    We recommend checking your build logs to verify which version of ``sphinx-rtd-theme`` you are running.
```
agjohnson, line 64: We shouldn't be communicating date based feature flags any more than strictly necessary. I do not give any exceptions for old projects in recommendations, *all projects should specify their dependencies*.

At this point, this tip is trying to say too much. And I'm not sure it's even helpful to the user to tell them how to find the version of the theme they have installed. We're trying to tell them how to install a *new version*, so linking them to our configuration file docs and the reproducible build docs is where I would point the reader.

In the interest of getting this PR merged, we should skip this tip and come back to this when we have the content ready to link to.

benjaoming: > all projects should specify their dependencies.

That is literally what the tip box above this one already says :wink: 

Not everyone have pinned their versions, for either good or bad reasons... or no reason at all. The tip box is there so that people who aren't getting the expected version of sphinx_rtd_theme on RTD have a pointer.

With your feedback, I don't have a clear picture of what sort of guidance (tip box) they should be given. It sounds like you would want a more generic "Troubleshooting your Read the Docs build? Try these articles..." tip box. But I don't get it really, I tried adjusting this a lot.

I agree we should give up, rather than continuing this comment thread :+1: I'd be interested to see your suggestion at a later point.

docs/development.rst
benjaoming: The blog now gives build errors when that happens :)

benjaoming:
Thanks for taking time to review @agjohnson - great to be working on the documentation of the theme :+1:

PR #1367, opened 2022-11-01
Bump to version 1.1.0

benjaoming:
There have been no reports since releasing 1.1.0b3 so we can bump and release our final 1.1.0 :tada:


agjohnson (approved):

benjaoming:
Tagged and released on PyPi.

Issue #1368, opened 2022-11-04
1.1.0 caused ugly looking internal references


dolfinus:
### Problem

I have a page with 2 tables - list of classes on the page, and list of methods in each class.

1.0.0 produced tables which looks fine:
![изображение](https://user-images.githubusercontent.com/4661021/199957134-9eaba88f-c1ec-4dd2-84df-e9baae6f1caf.png)
![изображение](https://user-images.githubusercontent.com/4661021/199957183-b67581c3-9144-4db1-850d-007ad929daa6.png)

But after upgrading to 1.1.0 internal links to classes and methods started to overflow:
![изображение](https://user-images.githubusercontent.com/4661021/199957343-caa7a941-03f6-4829-b78b-0c2b9dfe9d96.png)
![изображение](https://user-images.githubusercontent.com/4661021/199957397-8e4aaf1c-4cd0-4798-86e6-9392dc7b8856.png)

This is caused by #1193, after disabling CSS rule `overflow-wrap: anywhere` table appearance is normal again.

Same issue is observed in other projects, like https://github.com/PlasmaPy/PlasmaPy/issues/1784. To restore previous behavior, users either lock there version to `<1.1.0`, or uses some hacks like https://github.com/PlasmaPy/PlasmaPy/pull/1785.

### Reproducible Project

Sorry, my repo is not public.

Documentation is generated using `autosummary` and `autoclass` plugins.

```rst
.. currentmodule:: onetl.connection.db_connection.teradata

.. autosummary::

    Teradata
    Teradata.ReadOptions
    Teradata.WriteOptions
    Teradata.JDBCOptions

.. autoclass:: Teradata
    :members: check, sql, fetch, execute, close
```

Generated `href`s and other attributes for `<a>` tags are quite long:
```html
<tr class="row-odd">
  <td>
    <p><a class="reference internal" href="#onetl.connection.db_connection.teradata.Teradata"
        title="onetl.connection.db_connection.teradata.Teradata"><code
          class="xref py py-obj docutils literal notranslate"><span class="pre">Teradata</span></code></a></p>
  </td>
  <td>
    <p>Class for Teradata JDBC connection.</p>
  </td>
</tr>
```

```html
<tr class="row-even">
  <td>
    <p><a class="reference internal" href="#onetl.connection.db_connection.teradata.Teradata.execute"
        title="onetl.connection.db_connection.teradata.Teradata.execute"><code
[...]

PR #1369, opened 2022-11-04
Do not overflow object cross references

dolfinus:
Fix for #1368


benjaoming:
Thanks for this! I can see that this is quite a serious visual regression when it occurs. Let's get a patch out.

I've rebuilt the assets and added a demo.

benjaoming:
I'll have to change my demo, the approach of using the extension in the demo docs wasn't possible since the theme supports more versions of Python and Sphinx than the extension.

Maybe in the future, we can do tests like this.

docs/conf.py
```diff
@@ -29,6 +29,7 @@
     'sphinx.ext.viewcode',
     'sphinxcontrib.httpdomain',
     'sphinx_rtd_theme',
+    'sphinx_automodapi.automodapi',
```
dolfinus, line 32: I'm using 
```python
extensions = [
    "sphinx_rtd_theme",
    "sphinx.ext.autodoc",
    "sphinx.ext.autosummary",
]
```
Not `sphinx_automodapi`

benjaoming: Oh thanks! I picked this up from PlasmaPy's issues and documentation.

I'll add a demo for this :+1:

benjaoming: It would seem that sphinx-automodapi generates the same outputs as autosummary, so I will skip it entirely and just use autosummary :+1: 

Thanks for being quick!

tox.ini
```diff
@@ -49,5 +52,13 @@ deps =
...
+    # This is a short-term hack to allow us to have a Python 3-specific syntax
+    # and swap it out for a Python 2.7 version while we're building.
+    # This terrible solution is only acceptable because we will remove it again very soon
+    # See: https://github.com/readthedocs/sphinx_rtd_theme/pull/1369
+    py27: cp docs/demo/test_py_module/test.py docs/demo/test_py_module/test_py3.py
+    py27: cp docs/demo/test_py_module/test_py27.py docs/demo/test_py_module/test.py
     !html4: sphinx-build -b html -Dhtml4_writer=0 -d {envtmpdir}/doctrees docs/ {envtmpdir}/html
     html4: sphinx-build -b html -Dhtml4_writer=1 -d {envtmpdir}/doctrees docs/ {envtmpdir}/html
+    py27: cp docs/demo/test_py_module/test_py3.py docs/demo/test_py_module/test.py
+    py27: rm docs/demo/test_py_module/test_py3.py
```
benjaoming, line 64: @agjohnson I haven't had a proper look as to why this didn't break before now. I don't understand how our test builds got away with having a module that has invalid Python 2.7 syntax. The issue didn't come up until here.

The work-around is ugly. I created a Python 2.7 version of the same module, and during tests, I swap that python file with the Python 3 version.

It's ugly, but it works, and we can remove it very soon again.

agjohnson: Mm yeah that's quite fiddly and hidden. Any reason not to use conditional reST instead?

I've pushed up an example at https://github.com/readthedocs/sphinx_rtd_theme/compare/agj/pr-1369 and will see what CI says.

agjohnson: 27 tests are okay on my branch, and output looks correct:

image

agjohnson: And all tests are green and output seems to look okay. I'm going to proceed with bringing in my branch here and releasing

agjohnson (approved):
I've fixed up the pattern for conditional py27 testing. The cross reference fix itself looks good though. I'll get a release out next.

PR #1373, opened 2022-11-04
Update versioning docs to be explicit about our versioning workflow

agjohnson:
The versioning pattern we used last release didn't quite match our normal workflow, so I've made directions explicit. We ended up with a `b` release somehow (`1.1.0b3`), I'm assuming this was manually edited it.

Related to #1372

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -200,6 +200,33 @@ put up for review.
 
 .. _Transifex: https://www.transifex.com/readthedocs/sphinx-rtd-theme
 
+Versioning
+==========
+
+..
+    TODO make these instructions organization wide. We've talked about
+    standardizing packaging version and tooling here (bumpver instead of
+    bump2version).
+
+In between releases, the package version specified will always be a alpha
+release, matching ``1.1.1alpha1``.
+
+As pull requests are merged and the release impact changes, this version will
+incremented ahead of release. For example, if we merge a new feature pull
+request, we'd run ``bump2version minor`` to increment ``1.1.1alpha1`` to
+``1.2.0alpha1``.
+
+To put out a release for testing, we will upload a release candidate. We would
+run ``bump2version release`` to increment ``1.2.0alpha1`` to ``1.2.0rc1``.
+
+After a testing period, the release candidate can become a full release. We
+would run ``bump2version release`` again to increment ``1.2.0rc1`` to ``1.2.0``.
+
+After the release process below, we increment the version again, so that the
+development version -- the version found in the repository -- is always greater
+than the most recent release. We follow up with another pull request by running
+``bump2version patch``, which increments ``1.2.0`` to ``1.2.1alpha1``.
+
 Releasing the theme
 ===================
 
@@ -222,7 +249,9 @@ To release a new version of the theme, core team will take the following steps:
    version. But take care that it isn't also bumping versions of dependencies
    in ways that are risky to the release. *If* ``package-lock.json`` changes, you
    need to commit it to git and tag it together with your release.
-#. Commit these changes.
+#. Commit these changes to a pull request.
+#. Merge the pull request after review.
+#. Check out ``master`` branch locally for tagging
 #. Tag the release in git: ``git tag $NEW_VERSION``.
 #. Push the tag to GitHub: ``git push --tags origin``.
 #. Upload the package to PyPI:
@@ -235,7 +264,7 @@ To release a new version of the theme, core team will take the following steps:
 
 #. Finally, open a new pull request updating the development release version to
    the next patch by running ``bump2version patch``. Open a pull request with
-   this change.
+   this change. See `Versioning`_ above for more details.
 
 .. _PEP440: https://www.python.org/dev/peps/pep-0440/
 .. _semantic versioning: http://semver.org/


benjaoming:
Is this addressing the version string's format (`b` instead of `beta`) or maintaining version strings in-between releases (development versions)?

I have some vague experience of setuptools/pip automatically changing `1.1.0beta3` to `1.1.0b3`? I think there's some magick that makes things PEP440 compliant, not sure.

agjohnson:
It's addressing using `beta` at all actually, the current bump2version config does not use beta:

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/setup.cfg#L94-L100

Progression is `alpha` -> `rc` -> None (stable release)

This is fine, we only do one, *maybe* two release candidates before stable release.

benjaoming:
That's fine, I mean the label of the pre-release could appeal to make people want to test it and give feedback.. and I think we can be in a good position if we release more early alphas and switch to rc once the milestone is reached :+1:

docs/contributing.rst
```diff
@@ -200,6 +200,33 @@ put up for review.
 
 .. _Transifex: https://www.transifex.com/readthedocs/sphinx-rtd-theme
 
+Versioning
+==========
+
+..
+    TODO make these instructions organization wide. We've talked about
+    standardizing packaging version and tooling here (bumpver instead of
+    bump2version).
```
benjaoming, line 209: I'm not sure if I would make this organization-wide. It can vary for every project. Release processes also lean towards automation, although I'm not a big fan.

For this project, we could work more with dynamic generation of artifacts for each release.

A versioning guide should contain more details on the semantics of the versions. For instance, can a patch release contain translations? What can be changed in a minor release? How do we handle deprecation? Etc.

benjaoming (approved):

agjohnson:
> we release more early alphas and switch to rc

We don't/wouldn't release an `alpha` version to PyPI. These version numbers only exist in the repository. If we are pushing to PyPI, it should be an `rc` release.

benjaoming:
I think it's useful to have as many public releases as possible, as long as they are clearly labeled for testing purposes.

The hypothesis would be that if we can start to release development versions on PyPi as early as possible, we can hopefully have a bit more quality assurance :+1:

agjohnson:
I'm not proposing few PyPI releases. I'm saying our releases should be labeled `rc` releases, not `alpha`. The `alpha` label is reserved for development between releases. This is not my plan, it is the versioning scheme we added with the switch to bump2verison.

agjohnson:
What you're describing is releasing with a faster cadence. We don't need to release `alpha` releases and `rc` releases. In fact releasing both is probably a signal that our releases are too large in scope, frequent `rc` release should be sufficient.

If we have enough merged to put out a release, it should go out as an `rc`. The trap that we continue to get sucked into is creeping the scope of a release until it takes too much bandwidth to test and release, and then further building on top of that.

benjaoming:
A release candidate is normally something that has all intended changes ready and will get promoted to a stable release if no bugs are found.

I'm talking about making more releases from the development branch (what is currently `master`). So after a significant PR, we do an alpha. Once we have all PRs for a milestone, we do an `rc`. To make this conversation simpler, I'm trying to not bring in any new definitions, but to suggest something that seems to fit with the current state of affairs.

For instance for the 1.2 milestone, we want docutils 0.18.. after that feature is merged, I think we should release something. But we will also continue to work on Sphinx 6 support w/ sphinxcontrib-jquery before a 1.2 rc1 is released.

Does the example make sense or do you want an `rc` label even on releases that are going to significantly change?

agjohnson:
> Does the example make sense or do you want an rc label even on releases that are going to significantly change?

I am describing cutting smaller, more frequent releases using `rc` labeled releases. Releasing alphas after PRs is way too much overhead and will steal our attention and time away from putting out stable releases more often. This is all part of a discussion we've already had around our release procedure and we nixed the idea of frequent beta releases to PyPI.

benjaoming:
I understand that we could force more frequent releases by having fewer changes forcing out more `rc`s that we quickly promote to stable releases.

However, with the milestones implied by the roadmap, I think it's a little off to have `rc` labeling on all pre-releases that would follow from each PR generated by the roadmap.

We also have a large hump of PRs that would be nice to be able to have a good way to work on.

With a sign-off on closed issues and PRs like "Thank you for your contributions! The change is now out for testing in a pre-release. Please help us test this the latest `1.2.3rc5` in your documentation project." <= for the 1.2.0, 2.0 and 3.0 milestones, I want `alpha` or `beta` to come before `rc` so it's easy to ship pre-releases that are correctly labeled.

> Releasing alphas after PRs is way too much overhead and will steal our attention and time away from putting out stable releases more often.

I understand that I'm basically contradicting what you say has already been discussed and nixed. But look at the roadmap, it proposes a lot of changes and perhaps it'd be nice to leverage from the community that's around to help verifying releases.

Putting pre-releases out often is also supportive of putting stable releases out often, for instance new maintainers can get more practice :) I think having more pre-releases can be a good way to get more testing and feedback, and especially with the involvement of people in issues and PRs. I think that we already gone over the manual barriers of putting together releases, but that's also something we'll get to improve with a bit more automation :+1:

agjohnson:
> it'd be nice to leverage from the community that's around to help verifying releases.

There is nothing in our current plan that stops this. Aiming for alpha, beta, then rc releases is over-engineering this problem and is way more work than we can allocate to this project.

agjohnson:
I'm moving the organization wide discussion here https://github.com/readthedocs/meta/discussions/87

PR #1375, opened 2022-11-09
Docs: Remove "git install" method from instructions

benjaoming:
Motivation: I'd like to remove this already now to not cause anyone to be tempted by this installation method.

We should be able to put out more frequent releases so this should be entirely infeasible.


docs/changelog.rst
```diff
@@ -4,10 +4,15 @@ Changelog
...
+.. |development_version| replace:: 1.1.2alpha1
+
 Development version (|development_version|)
 ===========================================
 
-.. |development_version| replace:: 1.1.2alpha1
+Deprecated
+----------
+
+* A reminder to not install the theme directly from Git, we will remove support for this in 3.0
```
benjaoming, line 15: Currently, this section looks a bit empty, so it's nice to have a single item: https://sphinx-rtd-theme.readthedocs.io/en/latest/changelog.html

benjaoming: Maybe the text should be phrased a bit more "clearly"

```suggestion
* Reminder: We will remove support for installing this theme directly from its git repository. Instead, please use releases from PyPi.
```

humitos:
@agjohnson this PR is also a good candidate for 2.0 final -- It only removes the "Installing from Git" from the documentation, which will reduce the problem for the future when we drop support for this in 3.0

agjohnson (approved):

agjohnson:
Yup, this is safe to merge now. 2.0 was the plan for this deprecation as well.

Issue #1323, opened 2022-08-19
Support for docutils 0.19.x


benjaoming:
> docutils 0.19 has been released.

1. [ ] We should add support for and test docutils 0.19.x
2. [ ] report any issues, beyond the issues with docutils 0.18 reported in #1322 
3. [ ] remove brackets around footnote reference links https://github.com/readthedocs/sphinx_rtd_theme/issues/1323#issuecomment-1308926553

gabicca:
Hello, I wanted to note that there is a requirement incompatibility issue with the latest m2r2 package what was released on 11 Aug:

```
  docutils>=0.19 from m2r2==0.3.3
  docutils<0.18 from sphinx_rtd_theme==1.0.0
```

As you can see, they don't support docutils below 0.19 anymore, so if sphinx_rtd_theme could start supporting 0.19, that would be great.

Thanks!

benjaoming:
Thanks for this input! We'll be getting something out the door soon. First step is docutils 0.18, and after a few fixes for this support, we can see if it's feasible to add docutils 0.19.

If you are able to do testing with docutils 0.19, that'd be great. It's helpful to know your impressions. The type of testing that's needed is mostly around visual regression testing - it's a bit manual and tedious since it involves looking at the "correct"/old version and then compare to the proposed new version.

gabicca:
> Thanks for this input! We'll be getting something out the door soon. First step is docutils 0.18, and after a few fixes for this support, we can see if it's feasible to add docutils 0.19.
> 
> If you are able to do testing with docutils 0.19, that'd be great. It's helpful to know your impressions. The type of testing that's needed is mostly around visual regression testing - it's a bit manual and tedious since it involves looking at the "correct"/old version and then compare to the proposed new version.

Thanks @benjaoming for the quick reply and taking actions. Sorry for not getting back to you sooner. This week and next week will be busy for me and won't be able to do any testing. After that, unless the PR is already merged and the issue resoled, what I can do is use your branch to install sphinx_rtd_theme and see if our docs build properly. I could do minor comparisons testing as well, but won't be able to spend more than a few hours on it.

Paebbels:
Any progress on supporting latest docutils?

--------------
/cc @umarcor

benjaoming:
@Paebbels @umarcor

> Any progress on supporting latest docutils?

[...]

PR #1381, opened 2022-11-16
Add docutils 0.18 citation/footnote support using CSS grid

agjohnson:
This is a fix for #1322 that continues our pattern for HTML5 support
using CSS grids, introduced for support for docutils < 0.17 HTML5
support.

At normal widths, the columns try to stay at a consistent width, including secondary paragraphs in the footnotes/citations:

image

At narrow widths, the paragraphs for each row shrink to accommodate the extra room for the lable/backrefs:

image


agjohnson:
For comparison, here is the html5 writer output for docutils 0.17:

image

agjohnson:
One new test that is not displaying as I would prefer is if we add a lot more backrefs:

image

I'd prefer this to wrap nicer, which is possible by limiting the backrefs width and wrapping them.

benjaoming (approved):
This is definitely the most correct fix. I'll merge test fixes in a separate PR and close the other effort.

benjaoming:
If you are looking at this one more time, can you also expand the bottom margin of the `footnote-list` and `citation-list` to please Gestalt? :)

image

src/sass/_theme_rst.sass
```diff
@@ -313,9 +313,30 @@
...
+        grid-template-columns: auto auto minmax(0.65rem, 10%) 90%
+        & > span.label
+          grid-column-start: 1
+          grid-column-end: 2
+        & > span.backrefs
+          grid-column-start: 2
+          grid-column-end: 3
+        & > p
+          grid-column-start: 4
+          grid-column-end: 5
```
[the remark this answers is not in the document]
agjohnson: Haha, so CSS grids are _not_ approachable, and are _quite_ fiddly in addition. The solution here seems obvious in hindsight, but if you haven't had experience with CSS grid, it's quite hard to even know how to apply the concepts.

agjohnson:
Given my last note, I went back and tested docutils < 0.18 with lots of backrefs too, and display looks rather bad there too.

Lots of backrefs isn't unreasonable, I've done close to this in the past, but it certainly isn't a common issue of course. So, seems like something to fix with this implementation.

My solution so far requires a reasonable length for the backrefs column to balance the row widths. Once I start reducing the column width here though, display gets worse. The docutils/Sphinx output here is using a non-breaking, comma separated list of backrefs in between parenthesis. This does not wrap in a consistent way, so I'm going to try getting fancier here.

The fix I think I lean towards is dropping all of the unnecessary punctuation and making this a superscript list of links instead:
image
 
I would make the html5 output match for both docutils < 0.18 and >= 0.18 both use these styles, however the CSS grid structure has to be slightly different for docutils < 0.18.

This version will look good for 2-6 references at normal viewport sizes. Once it wraps, it looks less good, but better than other options, and much better than the comma separated list.

agjohnson:
I also think I can make the docutils < 0.18 html5 structure match visually here without much trouble, so might make this PR a bit farther reaching.

agjohnson:
Here is Sphinx 4.5 + docutils 0.17 styling with the above fixes:

image

And here is the docutils 0.18 version:

image

Note, the spacing between lines in slightly different between the two, but I can't explain why. Line height and related properties are similar :shrug:

agjohnson:
I can continue playing with this, I'm not 100% sold on using superscript here. But also, the default display with extra punctuation is both bad UX *and* displays poorly in a many cases. I'm not tied to it for such a minor use display element.

The backrefs have rather bad UX to them. There is no annotation or UI context as to what the links are. They are just a list of integers that happen to be links. Users that don't understand what that is won't be helped by extra punctuation.

The two options I'd prefer here are hiding the backrefs in a dynamic element -- popup or similar -- or putting the backrefs in the last row, under the footnote paragraph. Unfortunately, CSS grids does not support this on grids with implicit rows.

benjaoming:
> But also, the default display with extra punctuation is both bad UX and displays poorly in a many cases.

What is the extra punctuation? If it's the colon, I think that we agree to toss it :wastebasket:

benjaoming:
Looking at the clock ticking towards Sphinx 6, I'd suggest to stick to the former version*) and postpone any further improvements to "lots of backrefs" scenarios for later? Since we are slowly trimming legacy support, these improvements might also be easier to lift in 2.0 and 3.0.

*) The former version was when you initially had everything looking like this:

image

agjohnson:
> What is the extra punctuation? If it's the colon, I think that we agree to toss it wastebasket

The parenthesis and commas. These are causing trouble wrapping as the output has nothing to line break on.

benjaoming:
> The parenthesis and commas. These are causing trouble wrapping as the output has nothing to line break on.

Feature or bug? :)

agjohnson:
Hah yeah definitely a bug, the entire HTML structure of this output is deficient really. It's optimized for display without any styles, but that is not how anyone uses docutils, effectively. We can work with it, but it's not going to be good UX with the current structure.

The output that docutils 0.18 gives us:

`<span class="backrefs">(<a role="doc-backlink" href="#id3">1</a>,<a role="doc-backlink" href="#id17">2</a>)</span>`

Which, if it were optimizing for output that _can be styled_, and provides more context and a11y attention:

```html
<span class="backrefs">
  <ul>
    <li><a role=".." href=".." title="Back reference 1 for footnote 1, for section Foo Bar">1</a></li>
    <li><a role=".." href=".." title="Back reference 2 for footnote 1, for section Foo Bar">1</a></li>
  </ul>
</span>
```

Not something we can change though.

benjaoming:
Given a bit of creative well-meaning, the non-wrapping of backrefs can be seen as a feature. What I mean is that if forced onto the same line, parts of readability is maintained, while columns may stop aligning.

Very long lists of backrefs on narrow displays may not look nice, but the likelihood is low?

benjaoming:
Also agreed that the DOM is less-than-optimal. Might worth raising the issue upstream, although also annoying to have to implement yet-another-DOM, especially if implemented in a way that makes CSS selectors ambiguous.

benjaoming:
@agjohnson what do you think about maintaining the current behavior and postpone improvements and decisions around superscript to a separate issue?

agjohnson:
> What I mean is that if forced onto the same line, parts of readability is maintained, while columns may stop aligning.

Well, that's what I mean. The issue is more than just alignment, display is affected too. Multiple backrefs also bump the column out of the grid, because the backref column exceeds the display space that was allocated for it in the grid hack too:

image

As the screen narrows, it gets more exaggerated too (and clipped in addition), as the columns are percentage basis:

image

> @agjohnson what do you think about maintaining the behavior in [47516ea](https://github.com/readthedocs/sphinx_rtd_theme/pull/1381/commits/47516ea1f58714a575ee284f76cf191cab5f4375) and postpone improvements and decisions around superscript to a separate issue?

I'd like to resolve the bug above, but would agree we can come back to this at some point and spend more time on this. I can open up an issue to track the problem here as well.

agjohnson:
Dropping the last column minmax value down drastically didn't seem to affect display on Firefox and allows for rather narrow display before the columns start breaking out of the grid -- at ~400px wide.

agjohnson:
Okay, I feel this is as good as it will get, barring any more issues popping up on other browsers/etc. I introduced some pseudo class spacing in the backrefs and they at least wrap consistently now.

Sphinx with docutils 0.17 and HTML5 writer:

image

Sphinx with docutils 0.18 and HTML5 writer:

image

setup.cfg
```diff
@@ -45,7 +45,7 @@ packages = sphinx_rtd_theme
 python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*
 install_requires = 
 	sphinx >=1.6,<6
-	docutils <0.18
+	docutils <0.19
```
mitya57, line 48: Is there any reason to not allow docutils 0.19 as well?

AFAIK there are no significant differences in generated HTML between 0.18 and 0.19, so your changes will work fine with 0.19 as well, won't they?

Currently sphinx-rtd-theme is the only reason why I can't upgrade docutils package in Debian/Ubuntu from 0.17, and I would like to upgrade directly to 0.19, rather than from one old version to another old one.

benjaoming: https://github.com/readthedocs/sphinx_rtd_theme/issues/1323

mitya57: Okay, got it.

benjaoming:
LGTM!

I've done a run of visual regression testing on all supported Sphinx versions and with htmlwriter4 and htmlwriter5.

With Sphinx 5.1 and htmlwriter4

image

With Sphinx 4.5 and htmlwriter5

image

PR #1383, opened 2022-11-16
New release of sphinxcontrib-httpdomain==1.8.1 broke tests

benjaoming:
Cherry-pick from #1351

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,6 @@ extensions = [
     'sphinx.ext.autosummary',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
-    'sphinxcontrib.httpdomain',
     'sphinx_rtd_theme',
 ]
 
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,5 @@
 sphinx>=3.0
+
+# NB! sphinxcontrib-httpdomain dropped Sphinx 1.7 support
+# https://github.com/sphinx-contrib/httpdomain/issues/64
 sphinxcontrib-httpdomain
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,1 @@
 sphinx>=3.0
-
-# NB! sphinxcontrib-httpdomain dropped Sphinx 1.7 support
-# https://github.com/sphinx-contrib/httpdomain/issues/64
-sphinxcontrib-httpdomain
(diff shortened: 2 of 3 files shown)


agjohnson (commented):
We might also want to always use `sphinxcontrib-httpdomain<1.8.1`, so that we're limiting the dependency differences between builds. Installing differing releases of this could lead to output differences in our testing. We could always increase the pinned dependency after dropping python 2.7

benjaoming:
We have rules that target the output generated by sphinxcontrib-httpdomain, right (I haven't really checked)? I think it's better if we keep it unpinned so we can discover build failures or visual regressions, since that would affect users of the theme. It's also less work, because we won't have to monitor releases of sphinxcontrib-httpdomain and update our tests, instead we try to be just reactive.

The ultimate solution would be to write automated tests asserting that sphinxcontrib-httpdomain output looks as expected :)

agjohnson:
> We have rules that target the output generated by sphinxcontrib-httpdomain, right

We do not. And in fact, I don't think we have any styles for _any_ third party Sphinx extension. We have rules that _affect_ extensions like mathjax, autosummary, etc, but I'm not sure we have too many rules that even _directly target_ any of these extension's output.

Either way, there are no third party extensions that I would say we actually support.

So, for that, I already lean towards explicit pinning, as we don't need to distract ourselves with random package releases failing our builds.

> It's also less work, because we won't have to monitor releases of sphinxcontrib-httpdomain

If we supported httpdomain, or any other extension, and had some automated tests, this might be different. But I'd still say these tests should be a different test suite than our PR CI tests. If we aren't going to take any action on supporting a new release of a random extension, it's not helpful to fail our CI builds.

> The ultimate solution would be to write automated tests

Long term, yes. I agree here, and have proof of concepts for testing. I wouldn't take on official support for third party extensions until we have this set up though

agjohnson:
Hah. Okay, all of this is moot.

I went to see where we were actually displaying output from httpdomain and we aren't even using the httpdomain directives in this project. This extension be removed entirely from our dependencies and config entirely.

benjaoming:
:exploding_head: 

glad that you did that check!

I guess I'll just try to remove it entirely and see how it goes.

Blendify (approved):

agjohnson (approved):
:+1:

Nekimaliizgrada:
Check mate 🫡

PR #1380, opened 2022-11-15
Added aria-label to search box input field.

dmundra:
Fixes #1379

PR #1384, opened 2022-11-21
Added hidden text for home link in the breadcrumbs.

dmundra:
Fixes another a11y issue where the home link in the breadcrumbs needs a text.

Wave report https://wave.webaim.org/report#/https://sphinx-rtd-theme.readthedocs.io/en/latest/contributing.html

Wave report screenshot

Similar to #1379. Approach taken from https://fontawesome.com/v4/accessibility/. Let me know if I should also make an issue for this or if there is an existing issue.

### DIFF
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -22,7 +22,7 @@
 <div role="navigation" aria-label="{{ _('Page navigation') }}">
   <ul class="wy-breadcrumbs">
     {%- block breadcrumbs %}
-      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home"><span class="visuallyhidden">Home</span></a></li>
+      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home" aria-label="{{ _('Home') }}"></a></li>
         {%- for doc in parents %}
           <li class="breadcrumb-item"><a href="{{ doc.link|e }}">{{ doc.title }}</a></li>
         {%- endfor %}
--- a/sphinx_rtd_theme/breadcrumbs.html
+++ b/sphinx_rtd_theme/breadcrumbs.html
@@ -22,7 +22,7 @@
 <div role="navigation" aria-label="{{ _('Page navigation') }}">
   <ul class="wy-breadcrumbs">
     {%- block breadcrumbs %}
-      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home"></a></li>
+      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home"><span class="visuallyhidden">Home</span></a></li>
         {%- for doc in parents %}
           <li class="breadcrumb-item"><a href="{{ doc.link|e }}">{{ doc.title }}</a></li>
         {%- endfor %}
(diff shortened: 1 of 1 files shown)


sphinx_rtd_theme/breadcrumbs.html
```diff
@@ -22,7 +22,7 @@
 <div role="navigation" aria-label="{{ _('Page navigation') }}">
   <ul class="wy-breadcrumbs">
     {%- block breadcrumbs %}
-      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home"></a></li>
+      <li><a href="{{ pathto(master_doc) }}" class="icon icon-home"><span class="visuallyhidden">Home</span></a></li>
```
agjohnson, line 25: I'm guessing we probably want `aria-hidden="true"` on the icon still, and I would normally use `aria-label` here instead of a nested element with the text. What is the case for using the hidden `span` instead of `aria-label`? The interactive element example describe how I'd normally address this:

image

agjohnson (commented):
Good find :+1:

I noted a difference in pattern here. We used `aria-label` in similar places, rather than the (new to me) class `.visuallyhidden`, which I assume is from Bourbon/Neat/Wyrm/or some reset file.

sphinx_rtd_theme/breadcrumbs.html
dmundra: Good point. Replaced it with aria-label instead. Force of habit with icons.

dmundra:
> Good find 👍
> 
> I noted a difference in pattern here. We used `aria-label` in similar places, rather than the (new to me) class `.visuallyhidden`, which I assume is from Bourbon/Neat/Wyrm/or some reset file.

Ya it was a quick solution that I have used before and say it was available. `aria-label` is better since it is consistently used. It does look like there is a need to add 'Home' as a translatable text. Should that be added in this PR or a separate one?

agjohnson:
Aye, I've seen similar on other projects, though I've tended to use vanilla aria solutions too.

Great point on the translation, that is absolutely okay to include here. These changes will all be compiled and put up for translation prior to next release, but that is all mostly automated.

agjohnson:
Bleh, CircleCI is refusing to run checks again :clown_face:

Pushed up here to force checks: https://app.circleci.com/pipelines/github/readthedocs/sphinx_rtd_theme/864

agjohnson (approved):
Waiting on checks at https://app.circleci.com/pipelines/github/readthedocs/sphinx_rtd_theme/864, but looks good otherwise!

benjaoming:
@agjohnson this PR confirms your hypothesis in the other PR: https://github.com/readthedocs/sphinx_rtd_theme/pull/1380#issuecomment-1322447795

benjaoming:
Thanks @dmundra :100:

agjohnson:
> @agjohnson this PR confirms your hypothesis in the other PR: https://github.com/readthedocs/sphinx_rtd_theme/pull/1380#issuecomment-1322447795

Yeah, I had this back from CircelCI support today:

> We recently change how the config needs to be retrieved from the user's authentication rather than relying from follower of the project.
> As a workaround the best way for now is for the user is to login to CircleCI and commit to trigger or user who have the permission to commit on that branch to trigger the build.

benjaoming:
Ay, that's not a good move by Circle CI.

Circle CI is already a step behind GitHub Actions in terms of integration and popularity, and then they add this trip wire. I'm still all in favor of keeping the current setup. The easiest workaround is to amend the latest commit in a PR and force-push. It doesn't take much time and it amounts somewhat to the GitHub feature whereby a new contributor is allowed to have PRs built (albeit not 1:1 since we have to repeat for every PR).

PR #1388, opened 2022-12-15
Adds a remix of tox.ini used for visual/explorative regression testing

benjaoming:
I want to save this for another time. I didn't want to play around with a combination of the tox environment matrix, so I've put it in a separate file for now.

In a future version, we could be using Firefox/Selenium to grab screenshots and save them to an artifacts directory on Circle CI.

image


humitos:
This is a good idea. I'd propose some changes here:

1. update the PR with the latest changes after releasing 2.0
2. use `{-firefox}` and call it with `firefox: sh -c 'firefox -P sphinx-testing "{tox_root}/html/{env_name}/demo/demo.html" &'`

This way, the developer can call:

```
tox -e py310-sphinx{53,62,72,dev}-firefox
```

PR #1389, opened 2022-12-15
Add changelog entry for docutils 0.18 support

benjaoming:


docs/changelog.rst
```diff
@@ -9,6 +9,21 @@ Development version (|development_version|)
...
+
+Note for users of Sphinx<3
+--------------------------
+
+If you have pinned your documentation project to use versions of
+``sphinx<1.8.6`` or ``Sphinx>=2.0,<2.4.5``, you need to either
+add ``docutils<0.18`` to your requirements since newer versions of docutils are incompatible with these older Sphinx versions
+**or**
+update to a later release of Sphinx that has docutils correctly pinned.
+Read more in `this comment <https://github.com/readthedocs/sphinx_rtd_theme/pull/1304#issuecomment-1266924478>`.
```
agjohnson, line 25: For the future, basically no one looks at the changelog, so I don't like to spend much time here. But, at this point we should always guide users toward upgrading. There is a lot of Technically Correctness in the suggestion here that I think we can just skip entirely, and make our guidance "just upgrade Sphinx".

```suggestion
If you cannot use a more recent Sphinx release,
you should at least ensure you are using the most recent release for your major version.
Currently, these are Sphinx `1.8.6` and `2.4.5`.
Older releases may install unsupported versions of several dependencies, including Jinja2 and docutils.
```

benjaoming: > basically no one looks at the changelog

The ones that do have a good reason :)

benjaoming:
Is this approved?

docs/changelog.rst
agjohnson: Aye, and for anything major we should put this information somewhere more accessible, but agreed that for cases like this it's nice to have some artifact.

agjohnson (approved):

docs/changelog.rst
benjaoming: It's a good "read more" reference for blog posts and toots :+1:

PR #1390, opened 2022-12-15
Release 1.2.0alpha1

benjaoming:


agjohnson:
So, going back to our conversation on versioning, the release we are putting out should be `1.2.0rc1`, `alpha` releases are just for this repository and some development reasons.

benjaoming:
oh okay, I'll change that. I was thinking we could choose between "alpha" and "rc" and I chose "alpha" because we might still have some Sphinx 6 and jquery stuff in the horizon.

PR #1391, opened 2022-12-15
Release 1.2.0rc1

benjaoming:


benjaoming:
@agjohnson package-json.lock is updated but theme.js is unaffected, so it checks out :100:

PR #1397, opened 2022-12-30
1.2.0rc2

benjaoming:
Review the entirety of changes since 1.1.1 (last stable release) here:

https://github.com/readthedocs/sphinx_rtd_theme/compare/1.1.1...master


agjohnson:
Looks like table output format has changed here, for the worse:

image

agjohnson:
Also, the flyout menu doesn't work on our PR build demo docs because of the missing jQuery. I thought the plan was to load this automatically from our theme, using the sphinxcontrib-jquery extension. Did this change?

agjohnson (commented):
Changes in this PR all look good, but seems we need a few more fixes before we can cut a release with Sphinx 6 support.

agjohnson:
I just realized these are one in the same issue -- the extra table classes to control width come from JS code, which doesn't load because the theme is not including jQuery anymore.

benjaoming:
@agjohnson I was a bit quick to include some PRs, so if you want the best possible overview of code changes, perhaps this diff view is useful? https://github.com/readthedocs/sphinx_rtd_theme/compare/1.1.1...master

After publishing this pre-release, I suggest we do a bit more QA with the community and on docs.readthedocs.io (by enabling Sphinx 6 + 1.2.0rc2 there)

benjaoming:
Tables seem good again :+1: https://sphinx-rtd-theme--1397.org.readthedocs.build/en/1397/demo/lists_tables.html#giant-tables

agjohnson:
Bumping our own doc projects would be a good next step :+1:

agjohnson (approved):
Output looks good now, and confirmed the fix for jQuery is included :+1:

benjaoming:
Released! https://pypi.org/project/sphinx-rtd-theme/1.2.0rc2/

PR #1405, opened 2023-01-11
Sphinx 6: Use logo_url instead of logo

### LINKED ISSUE #1404: Sphinx 6: Broken Logos and favicons
Sphinx 6 removes the `logo` and `favicon` html template variables.
These have been deprecated since sphinx 4 in favor of `logo_url` and `favicon_url`.

This will need to be resolved before the 1.2.0 release.

benjaoming:
~Partly fixes #1404~

Fixes #1404 

Edit: added `favicon_url` so it's now fixing the entire issue.


benjaoming:
Tested that `docs/` build and appear correctly on Sphinx 1.8, 4 and 6.

benjaoming:
@Blendify or @agjohnson could you have a look here?

The git diff is pretty hard to read. But I do think I managed to make the template code simpler to read :upside_down_face:

benjaoming:
Added a favicon file and verified in Sphinx 1.8.5, 4.5.0 and 6.0.0.

sphinx_rtd_theme/layout.html
```diff
@@ -133,23 +135,32 @@
         <div class="wy-side-nav-search" {% if theme_style_nav_header_background %} style="background: {{theme_style_nav_header_background}}" {% endif %}>
           {%- block sidebartitle %}
 
-          {%- if logo and theme_logo_only %}
+          {%- if ( logo or logo_url ) and theme_logo_only %}
+
+          {%- if sphinx_version_info < (4, 0) -%}
```
agjohnson, line 140: This logic is a bit hard to follow, and I suspect a conditional on the Sphinx version is not needed. The following scenarios are possible:

* Sphinx < 4 + logo
* Sphinx 4<x<6 + logo -- this one seems missing from some of the logic below
* Sphinx 4<x<6 + logo_url

Instead, this decision tree can be reduced if we just set the variables we use ahead of time:

```
{% set _logo_url = logo_url|default:pathto('_static/' + logo, 1) %}
{% set _root_doc = root_doc|default:master_doc %}
```

Then just use those variables instead of using conditional logic in the templates. Does that have the same effect you're aiming for?

sphinx_rtd_theme/layout.html
```diff
@@ -133,23 +135,32 @@
...
+            </a>
+            {%- endif %}
+          {# Behavior since Sphinx 4 #}
           {%- else %}
-            <a href="{{ pathto(master_doc) }}" class="icon icon-home"> {{ project }}
+            {% if logo_url %}
+              <a href="{{ pathto(root_doc) }}">
+                <img src="{{ logo_url }}" class="logo" alt="{{ _('Logo') }}"/>
+              </a>
+            {%- endif %}
```
agjohnson, line 154: It seems like the conditional `if logo` is missing here now, correct?

sphinx_rtd_theme/layout.html
```diff
@@ -133,23 +135,32 @@
...
           {%- endif %}
 
-          {%- if logo %}
-            {#- Not strictly valid HTML, but it's the only way to display/scale
-                it properly, without weird scripting or heaps of work
-            #}
+          {%- else %}
```
agjohnson, line 157: It seems like `theme_logo_only` option wouldn't operate correctly with this logic. The conditional above is `if (logo or logo_url) and theme_logo_only`, but the `else` clause here doesn't output a logo.

That is, the current logic is meant to be:

```
<a href="...">
  {% if not theme_logo_only %}{{ project }}{% endif %}
  <img src="{{ logo }}" />
</a>
```

But if I'm interpreting the template logic here correctly, it's instead:

```
<a href="...">
  {% if theme_logo_only %}
    <img src="{{ logo }}" />
  {% else %}
    {{ project }}
  {% endif %}
</a>
```

sphinx_rtd_theme/layout.html
```diff
@@ -41,11 +41,13 @@
...
-    <link rel="shortcut icon" href="{{ pathto('_static/' + favicon, 1) }}"/>
-    {%- else %}
-    <link rel="shortcut icon" href="{{ favicon_url }}"/>
+  {%- if sphinx_version_info < (4, 0) -%}
+    {%- if favicon %}
+      <link rel="shortcut icon" href="{{ pathto('_static/' + favicon, 1) }}"/>
+    {%- endif %}
+  {%- else %}
+    {%- if favicon_url %}
+      <link rel="shortcut icon" href="{{ favicon_url }}"/>
```
agjohnson, line 50: This seems to miss the conditional `sphinx_version_info > (4,0) and favicon`. Similar to below, nested conditionals here can be reduce easily:

```suggestion
  {% set _favicon_url = favicon_url | default:pathto('_static/' + favicon, 1) %}
  <link rel="shortcut icon" href="{{ _favicon_url }}"/>
```

agjohnson (commented):
I haven't had my second coffee, so I might be interpreting the template logic wrong, but this read to me like it was missing on some parity with the current version.

I think dropping the Sphinx version conditional entirely is probably a good direction, as the logic is getting a bit complex

sphinx_rtd_theme/layout.html
benjaoming: I understand that it's possible to condense this into shorter statements, but I preferred the explicit statements that express version logic. It makes it easier to spot what may be eliminated once support of some version is dropped.

Or that was the intention.

If you prefer condensed expressions, I can rewrite it and test it again.

sphinx_rtd_theme/layout.html
benjaoming: The `logo_url` is a replacement helper function that works since Sphinx 4. So it's not necessary here.

sphinx_rtd_theme/layout.html
benjaoming: Same answer - I can change it to at the loss of the version logic (which makes it easier to remove in the future) and then test it again.

Your call :+1:

sphinx_rtd_theme/layout.html
benjaoming: That `a` tag has `class="icon icon-home"` when displaying `{{ project }}`

benjaoming: The logic that is "lost" from the **old** version is the ability to set `theme_logo_only=False` and get just the logo with `<a href="{{ pathto(master_doc) }}" class="icon icon-home"> {{ project }}` in the start. Which I think isn't desirable in any case.

Old version:

```
{%- if logo and theme_logo_only %}
  <a href="{{ pathto(master_doc) }}">
{%- else %}
  <a href="{{ pathto(master_doc) }}" class="icon icon-home"> {{ project }}
{%- endif %}

{%- if logo %}
  {#- Not strictly valid HTML, but it's the only way to display/scale
      it properly, without weird scripting or heaps of work
  #}
  {%- if sphinx_version_info < (4, 0) -%}
  <img src="{{ pathto('_static/' + logo, 1) }}" class="logo" alt="{{ _('Logo') }}"/>
  {%- else %}
  <img src="{{ logo_url }}" class="logo" alt="{{ _('Logo') }}"/>
  {%- endif %}
{%- endif %}
</a>
```

benjaoming:
@agjohnson is the summary that you'd like the version logic removed from conditionals in favor of a condensed set of statements using `|default` ? I can do that and test everything again. It was a conscious choice to avoid such, but I don't have a strong opinion.

sphinx_rtd_theme/layout.html
agjohnson: It's not as much about shortening this as it is about preserving the existing logic. The change here seems to drop support for `Sphinx 4<x<6` and usage of the `logo` variable. This is still valid usage, Sphinx 4 and 5 have both `logo` and `logo_url`. Relying on the Sphinx version in this logic is making it easier to break this conditional.

In the end, I don't think we care about what version of Sphinx the user is using, just what variables they are using.

sphinx_rtd_theme/layout.html
agjohnson: The version conditional seems to be causing more problems than it's solving, so I think it's fair to remove this.

We really only care about the variables being used, the Sphinx version doesn't help us in this determination.

sphinx_rtd_theme/layout.html
agjohnson: I'm aware of that change, but it is necessary to inspect that here as it's possible to use both the `logo` and `logo_url` variables in Sphinx 4 and 5.

sphinx_rtd_theme/layout.html
benjaoming: I know what you mean because that was also my first implementation draft until I realized that `logo_url` will exist always if `logo` is set. It's implemented like this in Sphinx 4-6:

https://github.com/sphinx-doc/sphinx/blob/c1f2a52d5d97f874d84e4d5421c99a566242af6b/sphinx/builders/html/__init__.py#L1252-L1257

sphinx_rtd_theme/layout.html
benjaoming: *) correction: not a helper function but just a context var: https://github.com/sphinx-doc/sphinx/blob/c1f2a52d5d97f874d84e4d5421c99a566242af6b/sphinx/builders/html/__init__.py#L1252-L1257

sphinx_rtd_theme/layout.html
benjaoming: Alright, that's up to you, I'll re-implement it with the suggestions that you gave and re-test it :+1:

sphinx_rtd_theme/layout.html
agjohnson: Your comments make more sense looking into the code:
https://github.com/sphinx-doc/sphinx/blob/v5.3.0/sphinx/builders/html/__init__.py#L1252-L1257

So support for Sphinx 4,5 + `logo` variable is indeed here. It's probably worth a comment here, as this conditional logic is already difficult to parse.

sphinx_rtd_theme/layout.html
agjohnson: [Mentioned below](https://github.com/readthedocs/sphinx_rtd_theme/pull/1405/files#r1068646958), I had to look at Sphinx's code to verify it was possible to use `logo` still. This at very least is worth noting in the logic.

Though I still feel this would be far clearer using variables instead of nested conditionals

sphinx_rtd_theme/layout.html
agjohnson: > Which I think isn't desirable in any case.

That is actually the intended use of that option though:
https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-logo_only

The options users have are:

- No logo, text only
- Logo and text (`logo_only = False`, the default)
- Logo only (`logo_only = True`)

benjaoming: Gotcha, it requires a bit of extra imagination :innocent:  but definitely a flexible configuration option that makes it possible to have all combinations :+1:

benjaoming:
I'll wrap this up some time tomorrow with the condensed logic and then test that. If we want to ship 1.2.0, I could pass the torch to you after the update, and then you can release 1.2.0 Monday or Friday @agjohnson 

If you want to review my only concern of releasing 1.2.0, it's a terrible investigation of how we may skip the inclusion of the jQuery asset through sphinxcontrib-jquery on RTD. See: https://github.com/readthedocs/readthedocs.org/pull/9654

sphinx_rtd_theme/layout.html
benjaoming: @agjohnson `html_logo` is a configuration value: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_logo and `logo_url` is the only context variable since Sphinx 6.

I've implemented the condensed versions, removing some Django template syntax -- I know you love it deep down inside :smile:

sphinx_rtd_theme/layout.html
benjaoming: The new version is a bit different, in some way more condensed than the former version wrt `theme_logo_only`.

benjaoming:
Haven't had a chance to test it with more than Sphinx 6, will test some more setups and get back. @agjohnson feel free to leave a review meanwhile, I can merge it in case manual tests turn out positive.

Blendify (approved):
Looks good to me, it is nice also that this cleans up the old code nicely.

benjaoming:
@agjohnson I think this can be merged and tested as 1.2.0rc3

sphinx_rtd_theme/layout.html
```diff
@@ -133,22 +133,15 @@
...
-            #}
-            {%- if sphinx_version_info < (4, 0) -%}
-            <img src="{{ pathto('_static/' + logo, 1) }}" class="logo" alt="{{ _('Logo') }}"/>
-            {%- else %}
-            <img src="{{ logo_url }}" class="logo" alt="{{ _('Logo') }}"/>
+          {# the logo helper function was removed in Sphinx 6 and deprecated since Sphinx 4 #}
+          {# the master_doc variable was renamed to root_doc in Sphinx 4 (master_doc still exists in later Sphinx versions) #}
+          {%- set _logo_url = logo_url|default(pathto('_static/' + (logo or ""), 1)) %}
+          {%- set _root_doc = root_doc|default(master_doc) %}
+          <a href="{{ pathto(master_doc) }}"{% if not theme_logo_only %} class="icon icon-home"{% endif %}>
```
agjohnson, line 140: Guessing we want to use the new variable here instead:

```suggestion
          <a href="{{ pathto(_root_doc) }}"{% if not theme_logo_only %} class="icon icon-home"{% endif %}>
```

sphinx_rtd_theme/layout.html
agjohnson: New version looks good! It was much faster to verify too, and I've only had one cup of coffee even :coffee:

benjaoming:
> I'm happy to take the release. It's feeling like the release candidate is about as ready as it will be, unless we dive very deep into testing or getting user adoption of the release candidate. We won't release on fridays, but otherwise, we can set a day based on how confident we are in the release. If we are hesitant about the release, 1.2.0 should probably go out Monday so we have a full week to put out a hotfix release if any issues pop up.

That sounds great! I agree that releasing a final version should be mon, tue, wed.. maybe thursday, if there is a good reason.

If I release 1.2.0rc3 after testing this on Sphinx 3+4, then we can have a look at how the candidate works out in some other projects and ship it tomorrow if successful. I'm happy to look at patches Thursday or Friday if - Egyptian Gods forbid - we broken anything.

agjohnson:
Sounds good, let's get an RC out whenever you'd like :+1: 

I'm itching to wrap up the RCs and get to a 1.2.0. Hopefully we're through most of the major issues with Sphinx 6 compatibility, but this issue is probably a good example of the sort of issues we'll see pop up when projects start using 1.2.0 all of a sudden

benjaoming:
This issue is because Sphinx didn't publish the removal in their section on Sphinx 6 changes. I saw an issue posted that it was missing and got distracted, so I forgot to review if it had an effect, which @Blendify noticed :100: 

I did go over other confvals that said "Changed in version 4.0" and noticed that `root_doc` was replacing `master_doc`. They have NOT added that in their [deprecation list](https://www.sphinx-doc.org/en/master/extdev/deprecated.html), so it's definitely fair to assume that we cannot guarantee that everything in Sphinx 6 is accounted for unless we go over everything that was changed in the release.

benjaoming:
Did manual testing on 3.5.4, 4.5.0 and 5.3.0 - logo, favicon and logo link are displaying well.

Issue #1303, opened 2022-07-11
Canonical URLs are being outputted as foo.html instead of foo/


dmalan:
### Problem

For https://cs50.readthedocs.io/, the source for which is at https://github.com/cs50/cs50.readthedocs.io, we're using **Sphinx HtmlDir** at https://readthedocs.org/dashboard/cs50/advanced/, so that our URLs are of the form, e.g., `https://cs50.readthedocs.io/code/` instead of `https://cs50.readthedocs.io/code.html`.

However, we noticed that `sphinx_rtd_theme==1.0.0` is still outputting 

```html
<link rel="canonical" href="https://cs50.readthedocs.io/code.html" /> 
```

per

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/layout.html#L52-L55

which we gather might be happening because `conf.py.tmpl` is setting `canonical_url` at https://github.com/readthedocs/readthedocs.org/blob/main/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl#L98. (Somehow, at least, `theme_canonical_url` is being set but not `pageurl`.) The result is that, e.g., Chrome on iOS, when you click the Share icon, copies that URL to the user's clipboard rather than the HtmlDir version thereof, so the recipient ends up with a 404.

The canonical `link` tag is not outputted when we build locally with `make dirhtml`, though.

### Reproducible Project

https://cs50.readthedocs.io/code/

#### Error Logs/Results

```
Read the Docs build information
Build id: 17409789
Project: cs50
Version: latest
Commit: 8cc4272ae83166ccdfa7f2ed4ebd175421d49038
Date: 2022-07-11T15:10:52.977814Z
State: finished
Success: True


[rtd-command-info] start-time: 2022-07-11T15:12:25.463926Z, end-time: 2022-07-11T15:12:25.926319Z, duration: 0, exit-code: 0
git clone --no-single-branch --depth 50 https://github.com/cs50/cs50.readthedocs.io .
Cloning into '.'...

[rtd-command-info] start-time: 2022-07-11T15:12:26.127245Z, end-time: 2022-07-11T15:12:26.213132Z, duration: 0, exit-code: 0
git checkout --force origin/main
Note: switching to 'origin/main'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

[...]

PR #1407, opened 2023-01-11
Respect that dirhtml builder does not have .html in canonical URLs

benjaoming:
Fix for #1406 

(will test manually since canonical URL does not seem to appear in local builds)


ericholscher (commented):
This seems like a reasonable approach, but I'd like to see a test if we're going to implement this. Seems enough enough to have a test case that sets the canonical URL via sphinx?

benjaoming:
@ericholscher yes, agreed -- I would think that setting `html_baseurl` in `conf.py` is enough to mimic the intended behavior. LMK if you had other thoughts.

I don't think this needs to be a roadmap item for 1.2.0 but would be nice to publish it sooner or later.

benjaoming:
I'm going to leave this for a little while - but I don't think it's hard to add a test dimension in Tox that parameterizes `conf.py` in a similar fashion to how we are doing with `-D writer=html4writer`. It could be something like `-D 'writer=dirhtml' -D 'html_baseurl=https://example.com'` and then a verification that the canonical URL is set correctly in the output HTML.

Blendify:
I am noticing that sphinx itself would have the same issue here: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html#L132

Probably best to use a mixture of https://www.sphinx-doc.org/en/master/development/templating.html#file_suffix
and https://www.sphinx-doc.org/en/master/development/templating.html#pagename

benjaoming:
@Blendify this is just a workaround :) You'll find the upstream issue here: https://github.com/sphinx-doc/sphinx/issues/9730

sphinx_rtd_theme/layout.html
```diff
@@ -51,12 +51,29 @@
...
+    <link rel="canonical" href="{{ theme_canonical_url }}{{ pagename }}{% if builder == 'dirhtml' %}/{% else %}.html{% endif %}"/>
   {%- endif -%}
 
-  {#- CANONICAL URL #}
-  {%- if pageurl %}
-    <link rel="canonical" href="{{ pageurl|e }}" />
+  {#-
+    CANONICAL URL
+    NB! pageurl is currently a non-documented template context variable!
+    pageurl implementation: https://www.sphinx-doc.org/en/master/_modules/sphinx/builders/html.html
```
humitos, line 60: I'd suggest a link to the GitHub source code at the correct line where this `pageurl` is injected in the context.

humitos (approved):
I think this is a good workaround for now. We should merge it and release a patch version.

sphinx_rtd_theme/layout.html
benjaoming, line 60:
```suggestion
    pageurl implementation:
    https://www.sphinx-doc.org/en/master/_modules/sphinx/builders/html.html
    https://github.com/readthedocs/readthedocs.org/blob/4b2eb75bebfce04390ac3f84d1c57ef7f7126fe8/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl#L151-L156
```

benjaoming: :+1:

dmalan:
Thanks for this, @benjaoming, @humitos. This would indeed help with https://github.com/readthedocs/sphinx_rtd_theme/issues/1303 for now!

PR #1408, opened 2023-01-11
Fix canonical in dirhtml with dirhtml builder

benjaoming:
Testing #1407 with `dirhtml` builder enabled


benjaoming:
The test indicates that `.html` is trimmed. I'm not sure if the Canonical URL on PR builds is supposed to point to the stable branch, but it seems like a good idea to avoid search indexing PR builds.

Blendify:
I am noticing that sphinx itself would have the same issue here: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html#L132

Probably best to use a mixture of https://www.sphinx-doc.org/en/master/development/templating.html#file_suffix
and https://www.sphinx-doc.org/en/master/development/templating.html#pagename

benjaoming:
@Blendify responded here: https://github.com/readthedocs/sphinx_rtd_theme/pull/1407#issuecomment-1385266816

PR #1409, opened 2023-01-13
Test: What happens when "sphinx_rtd_theme" is removed from extensions

benjaoming:
This is a POC of the issue in https://github.com/readthedocs/readthedocs.org/pull/9654

I am reproducing the issue locally: jQuery is no longer injected by sphinxcontrib-jquery in this setup.

The RTD injection needs to either be removed or be more intelligent.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,8 +27,16 @@ extensions = [
     'sphinx.ext.autosummary',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
-    'sphinx_rtd_theme',
+#    'sphinx_rtd_theme',
 ]
+
+
+# This same code is added by RTD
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -27,9 +27,17 @@
     'sphinx.ext.autosummary',
     'sphinx.ext.mathjax',
     'sphinx.ext.viewcode',
-    'sphinx_rtd_theme',
+#    'sphinx_rtd_theme',
```
benjaoming, line 30: This needs to be commented out to reproduce the error.

Or to state it differently: Because of the RTD injection, adding `'sphinx_rtd_theme'` to `extensions` is necessary.. otherwise `setup(app)` is not called.

PR #1414, opened 2023-01-17
Release 1.2.0rc3

benjaoming:
I added a "Known issues" section that's worth being aware of.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -7,7 +7,7 @@ Changelog
 Development version (|development_version|)
 ===========================================
 
-.. |development_version| replace:: 1.2.0rc2
+.. |development_version| replace:: 1.2.0rc3
 
 Dependency changes
 ------------------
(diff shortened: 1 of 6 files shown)


docs/changelog.rst
```diff
@@ -7,26 +7,41 @@ Changelog
...
 
+Known issues
+------------
+
+In some cases, jQuery is not loaded with Sphinx 6 on Read the Docs.
+Workaround: You need to add ``sphinx_rtd_theme`` to ``extensions`` in your ``conf.py``.
+See `readthedocs.org issue #9654`_ for updates.
+
+.. _readthedocs.org issue #9654: https://github.com/readthedocs/readthedocs.org/pull/9654
+
```
benjaoming, line 44: A note on this:

We will release rc3 with this known issue and then work to fix it in RTD before 1.2.0 because we cannot risk users building documentation that _looks_ like it's working but actually doesn't have jQuery and thus have specific small details broken (the RTD Flyout menu for instance)

agjohnson (approved):
We talked about the notes here and the blocking issue. The note now at least helps the rc3 go out, so seems :+1:

Issue #1412, opened 2023-01-17
Do not allow section anchor to wrap alone to next line


benjaoming:
We need a CSS rule that prohibits the Pilcrow (¶) from jumping to the next line. It's hidden so it looks weird.

![image](https://user-images.githubusercontent.com/374612/212976147-80c08e45-f3fa-406b-96f8-94009d8bf96d.png)

n-peugnet:
This can be done by adding the following CSS (only tested on Firefox for now):

```css
.rst-content a.headerlink {
  display: inline;
}
```

As no space is present between the headerlink and the previous word, it forces them to be wrapped together.

![Peek 2023-01-20 15-15](https://user-images.githubusercontent.com/23519418/213720530-f17e6276-b74f-4a2b-87d7-d7d6280a8b87.gif)

benjaoming:
> As no space is present between the headerlink and the previous word

Ah yeah, perhaps we can add a `&nbsp;` - I guess this is exactly what it is meant for :)

n-peugnet:
> > As no space is present between the headerlink and the previous word
> 
> Ah yeah, perhaps we can add a `&nbsp;` - I guess this is exactly what it is meant for :)

Not really, the problem comes from the fact that currently the `.headerlink` is set to `display: inline-block`, so it will always first wrap at the block boundary. With no CSS at all, the behaviour would be the expected one, as the `<a>` is not separated from the previous word with any space. Using a `&nbsp` here will only add complexity, adding space with `margin-left` as currently seems a better idea.

I never used SASS but it seems `display: inline` (or `display: initial`) can be added here: https://github.com/readthedocs/sphinx_rtd_theme/blob/6444ca893043d8a0ca704474f7cf7d5a148f6767/src/sass/_theme_rst.sass#L216-L226

benjaoming:
Roger! Thanks for the wonderful explanation!

If you would like to submit a PR, I can build the SASS and push to your PR :+1: Alternatively, there is a `Makefile` in our project with some Docker build stuff.

n-peugnet:
> If you would like to submit a PR, I can build the SASS and push to your PR

I think that I prefer this option. I will make the PR right now.

PR #1417, opened 2023-01-20
Prevent permalink icon from wrapping alone

n-peugnet:
`.fa` styles sets `display` to `inline-block`. We force the `.headerlink` elements to stay `inline`, to make it wrap together with the previous word.

This should fix: #1412 
But I will wait @benjaoming to push the built assets to see the result.

### DIFF
--- a/src/sass/_theme_rst.sass
+++ b/src/sass/_theme_rst.sass
@@ -219,6 +219,7 @@
       font-size: 14px
       font-family: FontAwesome
       margin-left: 0.5em
+      display: inline
       @extend .fa
       &:focus
         opacity: 1


benjaoming:
Tested a build with the same documentation that had the issue :heavy_check_mark: 

## Before

image

## After

image

benjaoming (approved):
@agjohnson any worries about this one?

From a test, this works. The placement of the pilcrow seems unchanged 
image

I also just discovered that it's not a pilcrow in later versions of Sphinx.

image

n-peugnet:
> I also just discovered that it's not a pilcrow in later versions of Sphinx.

I think the link icon comes from the RTD theme:
https://github.com/readthedocs/sphinx_rtd_theme/blob/6444ca893043d8a0ca704474f7cf7d5a148f6767/sphinx_rtd_theme/__init__.py#L72-L76

The strange part for me is : in which conditions do you have the pilcrow with Sphinx RTD theme ?

benjaoming:
> The strange part for me is: in which conditions do you have the pilcrow with Sphinx RTD theme?

LOL, I honestly didn't consider this the reason. So I'm actually seeing the new icon after we started building with a fix in https://github.com/readthedocs/readthedocs.org/pull/9654 -- long story short, then the `setup(app)` signal has been skipped in certain scenarios on Read the Docs itself.

PR #1426, opened 2023-01-27
:menuselection: style

### LINKED ISSUE #1392: Missing style for :menuselection:
Good start: Use same style as for `:guilabel:`

Even better? Use a distinct but similar variation of `:guilabel:` so it's clear to the reader that they are not identical.

Ref: https://github.com/readthedocs/readthedocs.org/pull/9801#discussion_r1050224197

benjaoming:
Fixes #1392

The intention is to be consistent with `:guilabel:` but differentiate the color.

Suggestion with gray

image

Suggestion with same style as guilabel - @humitos got to pick this one :heart: so he now has to let me pick the next color :wink: 

image


benjaoming:
@ericholscher @agjohnson you can see the rule in action here:

https://sphinx-rtd-theme--1426.org.readthedocs.build/en/1426/demo/demo.html#inline-markup

benjaoming:
There was already a demo in place, I just overlooked it. It's quite long, so it also demonstrates how the `:menuselection:` works in line breaks. I think it's acceptable.

image

benjaoming:
image

humitos (commented):
Looks good to me. However, I would use the same color for guilabel and menuselection. I don't think there is a need to introduce a new color here.

benjaoming:
Using 2 shades of `$gray` was an attempt to _not_ introduce new colors. I like that it's distinguished a bit from `guilabel`, otherwise they will look exactly the same.

humitos:
> I like that it's distinguished a bit from `guilabel`, otherwise they will look exactly the same.

Ah yeah, I was suggesting to use the same color because it makes sense to me to do not distinguish them since they are pretty related each other.

I don't think we are using this dark gray in any other visual element that's rendered in the content itself. I feel it a bit "disconnected" from the theme colors.

benjaoming:
It's because I sense that `guilabel` and `menuselection` will often appear right next to each other in the same how-tos that I like to distinguish them.

We use gray quite a lot in the theme actually :)

image

PR #1427, opened 2023-01-31
Bump docutils to include 0.19

phlax:
Signed-off-by: Ryan Northey <ryan@synca.io>

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ current_version = 1.2.0rc3
 commit = false
 tag = false
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)((?P<release>[a-z]+)(?P<dev>\d+))?
-serialize = 
+serialize =
 	{major}.{minor}.{patch}{release}{dev}
 	{major}.{minor}.{patch}
 
@@ -15,11 +15,11 @@ author = Dave Snider, Read the Docs, Inc. & contributors
 author_email = dev@readthedocs.org
 description = Read the Docs theme for Sphinx
 long_description = file: README.rst
-project_urls = 
+project_urls =
 	Homepage = https://sphinx-rtd-theme.readthedocs.io/
 	Source Code = https://github.com/readthedocs/sphinx_rtd_theme
 	Issue Tracker = https://github.com/readthedocs/sphinx_rtd_theme/issues
-classifiers = 
+classifiers =
 	Framework :: Sphinx
 	Framework :: Sphinx :: Theme
 	Development Status :: 5 - Production/Stable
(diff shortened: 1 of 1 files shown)


phlax:
apologies opened in error

benjaoming:
NP :+1: Does docutils 0.19 work well for you?

phlax:
> Does docutils 0.19 work well for you?

i didnt get so far as properly testing yet - it got rid of the invalid escape errors i was hitting with 0.18, and ci tests pass, but i didnt run a build - i can report back here as i do

benjaoming:
Thanks, it's great with feedback! What is the reason that you cannot use 0.18? Is that some bug in docutils 0.18 that's fixed in 0.19 or a dependency that isn't compatible?

phlax:
> Thanks, it's great with feedback! What is the reason that you cannot use 0.18? Is that some bug in docutils 0.18 that's fixed in 0.19 or a dependency that isn't compatible?

my motivation was 2-fold 

- i was immediately hitting invalid escape errors with 0.18 and didnt see any obvious solution
- wanting to be able to use the latest sphinx versions

reading through tickets it seemed like the pin is there more for backward compatibility - so it seemed like worth the effort to try running the newer version

benjaoming:
Sounds good! The reason that it's not supported in the sphinx_rtd_theme is because so many documentation projects are built without upper bounds on dependencies, so it's too risky to open for docutils 0.19 until it's well-tested. All feedback is appreciated :+1:

phlax:
So with updated/contrived deps:

```requirements
sphinx>=6.1.3
docutils~=0.19.0
sphinx-rtd-theme @ https://github.com/phlax/sphinx_rtd_theme/archive/11a587c63fd2237003e60a7359e8503c7ea094a1.zip#egg=sphinx_rtd_theme
sphinx-tabs @ https://github.com/phlax/sphinx-tabs/archive/338fda565dbf917650503b3cf0d35760f7fe07df.zip#egg=sphinx-tabs
```

all seems to be good - ci passes, no escape issues, and docs build

visually, clicky testing the built docs, im not seeing any differences - not sure exactly what to look for - we had issues with tabs <> docutils in the past - but all looks good 

the only issue im hitting with some custom jquery-based code - but i think that is just a regular upgrade type issue

phlax:
just following up on above - i do actually hit an issue with sphinx==6.1.3 - altho i think unrelated to either the theme or docutils

sphinx ticket is https://github.com/sphinx-doc/sphinx/issues/11163

benjaoming:
> the only issue im hitting with some custom jquery-based code

sphinxcontrib-jquery is a dependency of sphinx-rtd-theme and should ensure that jquery is loaded. However, as you can find in #1421, it doesn't get loaded on local builds.

PR #1433, opened 2023-02-07
Bump to next alpha for development

agjohnson:
I tried removing the words "development version" from the changelog header,
thinking it would work more seamlessly on release, but then the header ref link
is wrong. Leaving it for now.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,10 @@ Changelog
 
 .. seealso:: :ref:`howto_upgrade`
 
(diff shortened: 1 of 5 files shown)


benjaoming (approved):
LGTM :+1:

benjaoming:
When time is ripe, it'd be nice to decide via the theme's roadmap when we are switching to either a 1.3.x or 2.0.x series, so the PRs in the pipeline are easier to process.

PR #1445, opened 2023-03-13
Flyout: use the new integration point

humitos:
Use the documented integration point for our theme.
Read the official documentation at https://docs.readthedocs.io/en/stable/flyout-menu.html

I'm thinking about simplifying how all these integrations work and use only the ones that are documented. The idea is to inject the documented HTML tag that will be replaced by our integration at serve time.

Besides, I'm populating the template context with all the `READTHEDOCS` environment variables, so we can [stop passing them from the Read the Docs application at build time by injecting them into the `conf.py` file](https://github.com/readthedocs/readthedocs.org/blob/f83eee6964788a9a45cedf2ec48d211795a0362d/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl#L140).

Related: https://github.com/readthedocs/readthedocs.org/pull/10127

### DIFF
--- a/sphinx_rtd_theme/__init__.py
+++ b/sphinx_rtd_theme/__init__.py
@@ -4,6 +4,7 @@ Sphinx Read the Docs theme.
 From https://github.com/ryan-roemer/sphinx-bootstrap-theme.
 """
 
+import os
 from os import path
 from sys import version_info as python_version
 
@@ -35,6 +36,11 @@ def config_initiated(app, config):
 def extend_html_context(app, pagename, templatename, context, doctree):
      # Add ``sphinx_version_info`` tuple for use in Jinja templates
      context['sphinx_version_info'] = sphinx_version
+
+     # Populate the context with all the READTHEDOCS variables
+     for variable in os.environ:
+         if variable.startswith('READTHEDOCS'):
+             context[variable] = os.environ.get(variable)
 
 
 # See http://www.sphinx-doc.org/en/stable/theming.html#distribute-your-theme-as-a-python-package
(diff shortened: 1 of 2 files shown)


humitos:
> Besides, I'm populating the template context with all the `READTHEDOCS` environment variables, so we can [stop passing them from the Read the Docs application at build time by injecting them into the `conf.py` file](https://github.com/readthedocs/readthedocs.org/blob/f83eee6964788a9a45cedf2ec48d211795a0362d/readthedocs/doc_builder/templates/doc_builder/conf.py.tmpl#L140).

Most of these "context" variables should be converted into environment variables that all the doctool/theme authors can benefit from to integrate properly with their doctool/theme. Note this is something risen at https://github.com/readthedocs/readthedocs.org/issues/9423 as well.

sphinx_rtd_theme/__init__.py
```diff
@@ -36,6 +37,11 @@ def extend_html_context(app, pagename, templatename, context, doctree):
      # Add ``sphinx_version_info`` tuple for use in Jinja templates
      context['sphinx_version_info'] = sphinx_version
 
+     # Populate the context with all the READTHEDOCS variables
+     for variable in os.environ:
+         if variable.startswith('READTHEDOCS'):
+             context[variable] = os.environ.get(variable)
+
```
benjaoming, line 44: I wonder if we should undeprecate [the `canonical_url` option](https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html#confval-canonical_url) and use `READTHEDOCS_CANONICAL_URL` ?

(not for this PR, just in case you agree, we can open a separate issue)

humitos: I don't think so. I'd say the theme should probably define `base_url = os.environ("READTHEDOCS_CANONICAL_URL")` -- which will make everything work properly

benjaoming: Yes, except for DirHTML, but we're on it :)

humitos:
I'm closing this PR because this integrates "the old flyout" and we are moving forward with the "new flyout out" now, see https://github.com/readthedocs/sphinx_rtd_theme/issues/1523

PR #1464, opened 2023-04-30
Add Sphinx 7 support

### LINKED ISSUE #1463: Add Sphinx 7 support
### Problem

Support sphinx>=7

benjaoming:
* [x] ~Remove `-Dhtml4_writer=0` in build options~ that didn't seem necessary, builds are fine.

Fixes #1463


humitos:
Did you find any visual difference or we are good to move forward?

benjaoming:
@humitos That should be done, but I held back on systematic Q&A. We need a release plan, and I was hoping to have a more "combined" effort, so I opened up https://github.com/readthedocs/sphinx_rtd_theme/issues/1470

potiuk:
It Would be great to get support for Sphinx 7, it holds us back in Airflow a bit :(

mimi1vx:
fix for sphinx-7+ is pretty simple, but is incompatible with Sphinx older than 5.1

benjaoming:
@mimi1vx this change should not be required because it's guarded by a version conditional:

https://github.com/readthedocs/sphinx_rtd_theme/blob/cd6b429f875999ee48bd914da5369422aa55dd8f/sphinx_rtd_theme/layout.html#L27C1-L30

mimi1vx:
Ah, i saw only older code , so it's need to update version guard + add new code path for 7.x

setup.cfg
```diff
@@ -45,7 +45,7 @@ zip_safe = False
 packages = sphinx_rtd_theme
 python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*
 install_requires = 
-	sphinx >=1.6,<7
+	sphinx >=1.6,<8
```
ericholscher, line 48: I'd be 👍 on just dropping the upper bound, and letting users install the theme on newer Sphinx, and report display issues as they come up. It would stop this from blocking people's upgrades in the future. 
```suggestion
	sphinx >=1.6
```

agjohnson: I'm still hesitant to do anything like this. With docutils, Sphinx, and jinja releases, we've learned that having upper bounds would solve many dependency issues. Unfortunately, if there is an issue between our theme and Sphinx, it most likely will not result in an error or failure, but silently broken output.

These tend to turn into fires we have to put out too, as we can't easily/consistently retroactively apply these boundaries to dependencies for releases.

ericholscher: True, there's no obviously correct option, unfortunately :(

alex: FWIW, as a consumer:

We pin all of our dependencies, so we can easily roll-back if a Sphinx bump ends up conflicting with the theme. When the theme has the pin, we don't really have a choice, we have to wait, even if it would work fine.

That said, I think the best case would be if the theme had the cap but was able to update shortly after Sphinx releases. I realize that's kind of an unbounded commitment though :-)

lorengordon: Seconding this. As the consumer, it is my responsibility to pin the deps I'm using. Strongly prefer when libraries only have a minimum restriction on their own deps, and no max (particularly in the absence of the ability to update the max quickly and release).

> We pin all of our dependencies, so we can easily roll-back if a Sphinx bump ends up conflicting with the theme. When the theme has the pin, we don't really have a choice, we have to wait, even if it would work fine.

potiuk: I think @agjohnson -> it has to be done sooner or later, there many, many, many more problems with dependencies limited upper-bound for your consumers than not limting them. Basically by limiting Sphinx now you simply don't know of the problems and your users can't even help you with solving them by pointing out to issues. 

Suggestion: Release an RC  (or beta, whatever) version with Sphinx 7 removed and some basic testing done and ask all people commenting here to test it and report problems to you and even provide fixes. I would be personally super happy if there is one place where we coudl discuss those based on an RC.

I think maintainers of rtd_theme can observe enough of the willpower and community ready to make the move that if you simply ask for help on solving the problems, you will get it.

This is a great thing BTW. that you are releasing it for free and keeping it running - this is great and big thanks for years of doing it. And you have huge community that will help, if only they have a chance to.

humitos:
I pulled down this PR and did some local testing here. Basically, I build the theme's documentation with the current version from `master` and also using Sphinx 7.1.2 and ran `diff -U sphinx-6-2 sphinx7-1` (https://gist.github.com/humitos/54e35eaa896370803bec66bb845e46c5)

Differences found:

* script files have a hash on their URL (`?v=<hash>`)
* some changes on the `searchindex.js` that I can't realize exactly

I also looked side-by-side to the rendered pages following the whole "Demo documentation" section and I didn't find any change there.

Regarding the Sphinx upper limit, I'd propose to merge this PR with the upper limit (e.g. `sphinx >=1.6,<8`) and make a plan for v2.0.0 of the theme that removes all these restrictions and old supported versions. I will try to write down some thoughts in an issue.

ericholscher (approved):

setup.cfg
agjohnson: > We pin all of our dependencies
> Seconding this. As the consumer, it is my responsibility to pin the deps I'm using.

This is great! We really wish more projects did do this and it should be the default really. But unfortunately for us, many projects still do not pin dependencies -- both transitive and direct dependencies even. Some of this is general packaging awareness, but it also mostly stems from years of Read the Docs magically deciding dependencies for projects. Doing so definitely has led to projects not defining important dependencies, and relying on our build process to do this instead.

So while this is a annoyance for projects that do perform package maintenance well, strict pinning is meant to avoid build failures for all of the projects that don't define direct and transitive dependencies explicitly.

> I think @agjohnson -> it has to be done sooner or later

Yeah, I definitely agree with all of this too, I want to see this package limit it's dependencies a great deal. But there is indeed no correct answer for all users.

> Suggestion: Release an RC (or beta, whatever) version with Sphinx 7 removed and some basic testing done

Yup, absolutely. I'm currently on other projects, so it's a bit more up to other core team to cut releases right now, but we have been trying to aim for frequent, small stakes releases and more RC releases. Especially if we are avoiding taking on features on these releases, they are very low stakes, but still require an annoying amount of testing.

lorengordon: >it also mostly stems from years of Read the Docs magically deciding dependencies for projects. Doing so definitely has led to projects not defining important dependencies, and relying on our build process to do this instead.

I wonder if it would be reasonable for Read the Docs to manage the version pin in _its_ build process?

agjohnson: > I wonder if it would be reasonable for Read the Docs to manage the version pin in its build process?

Hah well this is exactly what we originally thought when we decided on this sort of magical approach to dependencies. But we're actually trying to move away from this pattern and stop holding user's hands through packaging configuration. Teach a person to fish and all that.

It's definitely a slow process though, and not exactly pain free, but does help projects (and surely our scope of concerns) in the long run.

agjohnson (approved):

agjohnson:
Anyone following along here, the new RC is out at https://pypi.org/project/sphinx-rtd-theme/1.3.0rc1/. You can install this with `pip install --pre sphinx_rtd_theme`

alex:
Thanks! Is there any place to follow along for the final release?

agjohnson:
@alex We'll mention anything major on our blog/newsletter and usually fediverse (we're not active on the bird/x site anymore).

Or for timely updates, PyPI has a great feature to monitor packages via RSS feed: https://pypi.org/rss/project/sphinx-rtd-theme/releases.xml

alex:
Thanks

potiuk:
Cool. On it :)

potiuk:
The rc1 has also docutils < 0.19 which effectively blocks sphinx to be upgraded. Can it be changed as well?

agjohnson:
@potiuk Not as easily, docutils is the one the source of HTML output changes. Docutils 0.19 support is tracked at:

- #1323 

But, you shouldn't need docutils 0.19 either. The specifier for Sphinx 7.0 and 7.1 is:

https://github.com/sphinx-doc/sphinx/blob/320d0b631cc7d9617f5167cc61c9284674f5b2df/pyproject.toml#L65

So you should be able to use docutils 0.18.1, unless there is something else going on.

potiuk:
> @potiuk Not as easily, docutils is the source of most HTML output changes that affect the theme normally. Docutils 0.19 support is tracked at:

Indeed - it was our limitation :) . On it, trying to see how far we are when we **just** build our docs with Sphinx 7 ... 🤞

potiuk:
We are on a good way to migrate after this change. We already have CSS fixes needed to get back the look & feel needed after docutils bump. All looks good :)

agjohnson:
@potiuk oh good, glad you were able to work around that. If you can share any changes you make on the CSS side for docutils 0.19/0.20, we'll be eventually making the same CSS changes here. It's usually obvious what changes, but it doesn't hurt to have some extra eyes on the style changes too.

potiuk:
Well have them up soon in our repo - then I will (waiting for the CSS expert :D to submit a pr. And I am not sure it will be directly applicable. It's a customisation of our custom CSS that we changed already long time ago. I think the main difference is that that docutils now produces a bitt different HTML for sections/headers etc. and the css that applied to the previous one did not apply to the new one.  But I will share it regardless.

agjohnson:
Ah gotcha. It might be useful still! No rush either way, we won't be jumping into docutils 0.2 support for a little bit.

Issue #1465, opened 2023-05-01
Theme error: UndefinedError("'style' is undefined")


kuripart:
### Problem

Getting following error when building  a Sphinx project:

```
#14 2.108 Theme error:
#14 2.108 An error happened in rendering the page about.
#14 2.108 Reason: UndefinedError("'style' is undefined")
```

`conf.py` includes the following:

```
# -- Options for HTML output -------------------------------------------------

# The theme to use for HTML and HTML Help pages.  See the documentation for
# a list of builtin themes.
#
html_theme = 'sphinx_rtd_theme'

# Add any paths that contain custom static files (such as style sheets) here,
# relative to this directory. They are copied after the builtin static files,
# so a file named "default.css" will overwrite the builtin "default.css".
html_static_path = ['_static']
```
Setting `html_style` doesn't fix the issue as well.

### Reproducible Project

Please give a link to a public reproducible project or provide an example of the Restructured Text that gives an issue.

```rst
Place RST code here (if any).
```

#### Error Logs/Results

#14 2.108 Theme error:
#14 2.108 An error happened in rendering the page about.
#14 2.108 Reason: UndefinedError("'style' is undefined")

#### Expected Results

Setting any other `html_theme` eg: 'bizstyle' works as expected.

### Environment Info

- Python Version: 3.9
- Sphinx Version: 7.0.0
- RTD Theme Version: 1.2.0

Happy to share the Dockerfile and Pipifile setting if the info provided is not enough.

dvzrv:
Hi! I am running into the same issue trying to build libcamera's documentation on Arch Linux.
This is with sphinx 7.0.0 and sphinx_rtd_theme 1.2.0 on Python 3.11.

benjaoming:
Hi! Thanks for stopping by and reporting this!

Do you have an open source repo that produces this error?

There's no variables or properties in the theme's Python code called `style`, so I wondered if it could be related to Jinja2 template code and found this:

https://github.com/readthedocs/sphinx_rtd_theme/blob/03a86b803cf7cf281748ef94345ba7c35953952a/sphinx_rtd_theme/layout.html#L27-L30

As you can see, this code is guarded to only be used on old versions of Sphinx. But could the variable substitution still be causing the error?

benjaoming:
Thanks! I'm wondering why some Sphinx 7 builds work fine. For instance, this build doesn't produce the error:

https://readthedocs.org/projects/test-builds/builds/20357244/

[...]

PR #1467, opened 2023-05-02
Switch to `styles` key from `style` key

polyzen:
`style` key removed in Sphinx 7:
https://github.com/sphinx-doc/sphinx/pull/11381

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -25,7 +25,7 @@
 
   {#- CSS #}
   {%- if sphinx_version_info < (4, 0) -%}
-    <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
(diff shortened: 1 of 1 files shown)


sphinx_rtd_theme/layout.html
```diff
@@ -25,7 +25,7 @@
 
   {#- CSS #}
   {%- if sphinx_version_info < (4, 0) -%}
-    <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto('_static/' + styles[-1], 1) }}" type="text/css" />
```
benjaoming, line 28: @polyzen - thanks for opening this. However, the line that you are changing is only be relevant for `if sphinx_version_info < (4, 0)`.

I think https://github.com/readthedocs/sphinx_rtd_theme/issues/1465 was closed because the bug was finally identified to have another origin?

polyzen: Brainfart.

PR #1476, opened 2023-05-18
Docs: Typo in recommended pinning

tapple:
The paragraph above recommends "below the next major version". But the example says `<=`. Make them match


benjaoming (approved):
Thanks for spotting that and fixing it :dart:

Issue #1434, opened 2023-02-08
Search not working in 1.2.0 in environments without web access


tmct:
### Problem

Hi,

When I upgrade to 1.2.0, the "search" feature stops working in airgapped environments. When a search is initiated, it "spins" but does not finish.

Also observed:
1) on load, the webpage tries to access https://ajax.googleapis.com/ajax/libs/jquery/3.6.0/jquery.min.js upon load, and this fails in environments with no internet access, as expected
2) "Uncaught ReferenceError: jQuery is not defined" happens when you initiate a search

### Reproducible Project

Cannot share, sorry - but it shouldn't matter

#### Error Logs/Results

See above

#### Expected Results

Search completes as expected

### Environment Info

- Python Version: 3.8 and 3.9
- Sphinx Version: 6.1.3
- RTD Theme Version: 1.2.0

benjaoming:
@tmct we forced the version of `sphinxcontrib-jquery` to exclude `3.0.0` because there were incoming reports that jQuery wasn't loaded in local builds because of an SRI integrity check. See: https://github.com/readthedocs/sphinx_rtd_theme/issues/1420

If you can force-feed pip `sphinxcontrib-jquery==3.0.0`, you should be able to serve an airgapped build.

This will be finally fixed by a new version of sphinxcontrib-jquery: https://github.com/sphinx-contrib/jquery/pull/14

tmct:
Thanks very much! I'll keep an eye on that `sphinxcontrib-jquery` change.

gherceg:
@benjaoming we are experiencing a similar issue, but after upgrading `sphinxcontrib-jquery` to 4.1 (with `sphinx==6.1.3` and `sphinx-rtd-theme==1.2`), the "Uncaught ReferenceError: jQuery is not defined" error still occurs when using the search bar.

benjaoming:
@gherceg thanks for the feedback!! do you have a build log, a link to the documentation and/or a git repo that I can have a look at?

gherceg:
Yeah of course. [commcare-hq](https://github.com/dimagi/commcare-hq) is the repository, and the docs are [here](https://commcare-hq.readthedocs.io/), but notably that docs build is still running `sphinxcontrib-jquery==2.0.0`, with the same versions for sphinx and sphinx-rtd-theme mentioned above.

[...]

PR #1479, opened 2023-05-23
Do not use theme as extension, this should be made redundant

benjaoming:
This seeks to demonstrate that our `setup(app)` block is called correctly and all subsequent functionality is called as intended. Including the call to `jquery_add_js_files` that was broken because themes cannot activate extensions.

https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/__init__.py#L41-L85


humitos:
If this is true, we should update our installing instructions as well https://sphinx-rtd-theme.readthedocs.io/en/stable/installing.html

humitos:
I don't 100% understand the Sphinx internals, but it seems we cannot remove the installation as an extension: https://github.com/readthedocs/sphinx_rtd_theme/issues/1434#issuecomment-2289443380

benjaoming:
Hi @humitos 

I recently had to deal with yet another instance of this bug, it keeps coming up :) It's because of defining `html_theme_path` in `conf.py` which should not be needed, but has been part of many many instructions and circulating copy-pasta.

In the case in question, it's defined here:

https://github.com/LGouellec/streamiz/blob/d1a0b73b2ebc5a82c8d00babf16b51f2a7423841/docs/conf.py#L61-L64

I tried to summarize the issue here: https://github.com/readthedocs/sphinx_rtd_theme/issues/1434#issuecomment-1472671651

Another thing is that maybe the need for `html_theme_path` needs to be understood better... but I think that the entire code block above should be possible to remove.

humitos:
Hey 👋🏼 . 

> It's because of defining `html_theme_path` in `conf.py` which should not be needed, but has been part of many many instructions and circulating copy-pasta.

Yeah, it seems we removed this from our installation page a while ago now.

We still have the function `get_html_theme_path()` declared in our code. So, it would be probably a good idea to remove it from there and make those projects defining `html_theme_path` to hard fail; so they can upgrade.

benjaoming:
@humitos that sounds like a good way forward because the missing JQuery causes confusion and can waste a lot of time in debugging and for the users that are impacted (negatively impacting for instance search). If possible, a potential compromise could be to have a DeprecationWarning lodged in an intermittent update, then remove it completely (or rename it, in case there's something unanticipated that really needs it).

humitos:
> If possible, a potential compromise could be to have a DeprecationWarning lodged in an intermittent update

I added a commit [`cc5067a` (#1576)](https://github.com/readthedocs/sphinx_rtd_theme/pull/1576/commits/cc5067af431abbc32f2601945f5cc6d1a543d67f) to our 3.0.0rc1 release 👍🏼 . Let me know if that works.

PR #1489, opened 2023-07-19
Fix code-block linenumbers highlight on Safari

asikeero:
## Issue
When using Safari and highlighting text in a code block with line numbers, the line numbers are also included. Line numbers are not included with eg. Firefox.

## How to reproduce
Open https://sphinx-rtd-theme.readthedocs.io/en/stable/demo/demo.html#emphasized-lines-with-line-numbers with Safari (I used v16.5.2) and highlight the code block. 
image

## Expected behavior
Line numbers should not be included when highlighting.

# What is done
- Added the `-webkit-user-select: none` field. This is needed on Safari to disable highlighting. See https://developer.mozilla.org/en-US/docs/Web/CSS/user-select for details.

### DIFF
--- a/src/sass/_theme_rst.sass
+++ b/src/sass/_theme_rst.sass
@@ -103,6 +103,7 @@
   div.highlight
     span.linenos, .gp
       user-select: none
+      -webkit-user-select: none  // Safari fallback
       pointer-events: none
 
     span.linenos


src/sass/_theme_rst.sass
```diff
@@ -103,6 +103,7 @@
   div.highlight
     span.linenos, .gp
       user-select: none
+      -webkit-user-select: none  // Safari fallback
```
agjohnson, line 106:
```suggestion
      // Safari 3+ requires vendor prefix still
      -webkit-user-select: none
```

agjohnson (commented):
Good catch! We're in the middle of a light code freeze while we get some releases out strictly focusing on dependency trimming. This can be planned for version 2.1 or later though.

asikeero:
Sounds good 👍

PR #1497, opened 2023-08-05
Make permalink indicators not user-selectable

felixfontein:
Bug description from https://github.com/ansible/ansible-documentation/issues/245:

> When I double-click on section header to copy it to my code I don't want to finish with permalink-emoji within my YAML. :-)

Making the permalink indicator (`.headerlink`) not user-selectable fixes this.


samccann:
Tested this fix locally on the Ansible docs and seems to work, thanks @felixfontein !

Issue #1493, opened 2023-08-03
Reduce supported versions and release plan


humitos:
While working a little on #1463 I found that **this theme supports every single version from Sphinx 1.7 to 7.1, including HTML4 and HTML5 writers and Python 2.7 to 3.10.** In my opinion, this is "_great, but insane_". The main point here is that we don't have enough time to support all these versions and we've been pretty blocked when trying to move forward even with small changes because it breaks in a weird way in some random old version or combination.

I want to propose reducing the matrix of supported versions to:

- Last 3 major Sphinx versions: 5.x, 6.x, 7.x
- Last 2 docutils versions supported by Sphinx: 0.19.x, 0.20.x
- Last 3 major Python versions: 3.9, 3.10, 3.11
- HTML5 only (Sphinx does not support HTML4 anymore since Sphinx 7: https://github.com/sphinx-doc/sphinx/pull/11385)

This proposal will reduce the matrix a lot and also the complexity involved when testing and QAing. I think we have a good opportunity here to do a 2.0 release that removes all these new unsupported versions and start from fresh again aiming to recover some development pace by keeping things updated at least -- I'm not even talking about adding new features or similar.

### Related:
* https://sphinx-rtd-theme.readthedocs.io/en/stable/development.html#supported-dependencies
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1463
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1323
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1470
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1364
* https://github.com/readthedocs/sphinx_rtd_theme/issues/1086

agjohnson:
See also: https://github.com/readthedocs/sphinx_rtd_theme/issues/1086 which has a lot more conversation on the deprecations

humitos:
> This is already close to our plan, though I don't think we should aim for all of these deprecations at once in a 2.0 release.

What is blocking us from doing all these deprecation at once?

> 2.0 hasn't happened yet, but the deprecations are listed here: [sphinx-rtd-theme.readthedocs.io/en/stable/development.html#roadmap-release-2-0-0](https://sphinx-rtd-theme.readthedocs.io/en/stable/development.html#roadmap-release-2-0-0). We can definitely expand that more, but I think even Sphinx 3 is probably fine to include as the API hasn't changed too drastically after 2.x.

[...]

PR #1500, opened 2023-08-10
2.0 milestone: drop supported versions

humitos:
The new policy is:

- Python >=3.6
- Sphinx >= 5
- HTML4 and HTML5 writer
- docutils >= 0.14, < 0.19

Related: https://github.com/readthedocs/sphinx_rtd_theme/issues/1493

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -143,36 +143,43 @@ like Sphinx 1.8, HTML4, or required support for IE11.
 2.0.0
 ~~~~~
 
-:Planned release date: 2022 Q1
+:Planned release date: 2023 Q3
 
-This release will mark the beginning of a new round of feature development, as
-well as a number of backward incompatible changes and deprecations.
+This release will drop support for old Python and Sphinx versions.
 
-Of note, the following backwards incompatible changes are planned for this
+Note the following backwards incompatible changes are planned for this
 release:
 
-Sphinx 1.x, Sphinx 2.x, and Docutils 0.16 will not be tested
+Python < 3.8, Sphinx <5 and Docutils < 0.18 will not be tested
     Official support will drop for these version, though they may still continue
     to work. Theme developers will not be testing these versions any longer.
-
-HTML4 support will be removed
-    Starting with this release, we will only support the HTML5 writer output,
-    and builds attempting to use the HTML4 writer will fail. If you are still
-    using the HTML4 writer, or have the ``html4_writer = True`` option in your
-    Sphinx configuration file, you will need to either remove this option or pin
-    your dependency to ``sphinx_rtd_theme<=2.0.0`` until you can.
-
-    This option was suggested in the past to work around issues with HTML5
-    support and should no longer be required to use a modern combination of this
-    theme and Sphinx.
 
 .. _roadmap-release-3.0.0:
 
 3.0.0
 ~~~~~
 
-This release is not yet planned, however there are plans to potentially replace
-Wyrm with Bootstrap in a release after 2.0.
+:Planned release date: 2024 Q1
+
+This release is not yet fully planned.
+However, we already know that HTML4 support will be dropped.
+
+HTML4 support will be removed
+    Starting with this release, we will only support the HTML5 writer output,
+    and builds attempting to use the HTML4 writer will fail. If you are still
+    using the HTML4 writer, or have the ``html4_writer = True`` option in your
+    Sphinx configuration file, you will need to either remove this option or pin
+    your dependency to ``sphinx_rtd_theme<=3`` until you can.
+
+    This option was suggested in the past to work around issues with HTML5
+    support and should no longer be required to use a modern combination of this
+    theme and Sphinx.
+
+4.0.0
+~~~~~
+
+This release is not yet planned.
+There are plans to potentially replace Wyrm with Bootstrap in a release after 3.0.
 
 Also tentatively planned for this release is finally removing built CSS and
 JavaScript assets from our repository. This will remove the ability to install
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -62,11 +62,20 @@ The theme officially supports the following dependencies in your Sphinx project:
     * - Dependency
       - Versions
     * - Python
-      - 2.7 or 3.6 or greater
+      - 3.6 or greater
     * - Sphinx
-      - 1.7 up to at least 4.1
+      - 5 or greater
     * - docutils
-      - Up to 0.17
+      - >= 0.14, < 0.19
+
+.. deprecated:: 2.0
+    Sphinx < 5 support removed
+
+.. deprecated:: 2.0
+    Python < 3.6 support removed
+
+.. deprecated:: 2.0
+    docutils < 0.14 support removed
 
 .. versionadded:: 1.0
     Sphinx 4.0 support added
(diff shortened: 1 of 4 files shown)


docs/development.rst
```diff
@@ -143,36 +143,43 @@ like Sphinx 1.8, HTML4, or required support for IE11.
...
-This release will mark the beginning of a new round of feature development, as
-well as a number of backward incompatible changes and deprecations.
+This release will drop support for old Python and Sphinx versions.
 
-Of note, the following backwards incompatible changes are planned for this
+Note the following backwards incompatible changes are planned for this
 release:
 
-Sphinx 1.x, Sphinx 2.x, and Docutils 0.16 will not be tested
+Python < 3.8, Sphinx <5 and Docutils < 0.18 will not be tested
```
agjohnson, line 153: This doesn't quite match what we talked about, though we also only talked about *what we'd support* and not what we'd *test against*.

But where we are testing Sphinx 5, we should look to the Sphinx 5 dependencies to dictate our support:

https://github.com/sphinx-doc/sphinx/blob/953002e6261bdd9314ee0a3314aae19479a88c7e/setup.py#L10-L31

So, Python 3.6 and above and docutils 0.14 and above.

However, we did not talk about what we would *test* against. This can certainly be different than what we want to support. `Python>=3.8, docutils>=0.18` would be valid if we say we test the last two versions of Sphinx, but seems odd to say we test against a different set of dependencies than we support too.

setup.cfg
```diff
@@ -19,18 +19,15 @@ project_urls =
...
 	Framework :: Sphinx :: Theme
 	Development Status :: 5 - Production/Stable
 	License :: OSI Approved :: MIT License
 	Environment :: Console
 	Environment :: Web Environment
 	Intended Audience :: Developers
-	Programming Language :: Python :: 2.7
 	Programming Language :: Python :: 3
-	Programming Language :: Python :: 3.6
-	Programming Language :: Python :: 3.7
```
agjohnson, line 33: Supporting Sphinx 5 means these are still valid, as Sphinx 5 supports Python>=3.6

setup.cfg
```diff
@@ -43,9 +40,9 @@ classifiers =
 include_package_data = True
 zip_safe = False
 packages = sphinx_rtd_theme
-python_requires = >=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*,!=3.5.*
+python_requires = >=3.8
```
agjohnson, line 43: Same feedback here as above, this should be

```suggestion
python_requires = >=3.6
```

tox.ini
```diff
@@ -1,14 +1,10 @@
...
-    # Python 2.7 support was removed in Sphinx 2
-    py{27}-sphinx{17,18}{-html4,-html5,}{-qa,}
-    py{36,37,38,39}-sphinx{18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,43,44,45,50,51,52}{-html4,-html5,}{-qa,}
-    # Python 3.10 working from Sphinx 4.2 and up
-    py{310}-sphinx{42,43,44,45,50,51,52,53,latest}{-html4,-html5}{-qa,}
-    # Sphinx 6+ has simplified docutils and Python support
-    py{38,39,10}-sphinx{60,61,70,71}{-html5,}{-qa,}
+    py{38,39,310}-sphinx{50,51,52,53,60,61}{-html4,-html5,}
+    # HTML4 is not support on Sphinx >=7.x
+    py{38,39,310}-sphinx{70,71,latest}{-html5,}
```
agjohnson, line 5: There is a conversation about what we really need to be testing here too. While it's helpful to run these through CI, I think visually testing 3.6 - 3.10 is not helpful. I'd love to have a better policy here and think we could say something like we visually test against the latest Python only? This reduces our support matrix a great deal.

Same feedback here though, we should be running all of our supported Pythons through CI at least.

tox.ini
```diff
@@ -44,14 +21,6 @@ deps =
...
     sphinx70: Sphinx>=7.0,<7.1
     sphinx71: Sphinx>=7.1,<7.2
-    # All these Sphinx versions actually break since docutils 0.18, so we need to add this upper bound
-    # Projects using these Sphinx versions will have to do the same
-    # See: https://github.com/readthedocs/sphinx_rtd_theme/pull/1304
-    sphinx{17,18,20,21,22,23,24,30,31,32,33,34,35,40,41,42}: docutils<0.18
-    # External environments are required to add this dependency for older versions of Sphinx
-    # because it didn't ship with this upper bound.
-    # See: https://github.com/sphinx-doc/sphinx/issues/10291
-    sphinx{17,18,20,21,22,23,24,30,31,32,33,34,35,40}: Jinja2<3.1
```
agjohnson, line 54: Glad to get rid of these one off rules :+1:

agjohnson (commented):
This doesn't quite match what Sphinx 5 supports, so I don't think we should make all of these changes. I *would* be in favor of more aggressive tuning on what we *visually test against* -- mostly Python versions really, but perhaps docutils versions too.

Once our support matrix drops Sphinx 5 these will get tighter, but I think 5 usage is still high enough to warrant support. We're just at an unfortunate stage where our dependency graph and Sphinx's dependency graph is very wide still.

humitos:
@agjohnson I updated this PR with the changes suggested. Let me know if you OK with this. I'd be happy to move forward and get closer to a 2.0 release 🚀

tox.ini
```diff
@@ -1,14 +1,11 @@
...
 envlist =
-    # Python 2.7 support was removed in Sphinx 2
-    py{27}-sphinx{17,18}{-html4,-html5,}{-qa,}
-    py{36,37,38,39}-sphinx{18,20,21,22,23,24,30,31,32,33,34,35,40,41,42,43,44,45,50,51,52}{-html4,-html5,}{-qa,}
-    # Python 3.10 working from Sphinx 4.2 and up
-    py{310}-sphinx{42,43,44,45,50,51,52,53,latest}{-html4,-html5}{-qa,}
-    # Sphinx 6+ has simplified docutils and Python support
-    py{38,39,10}-sphinx{60,61,70,71}{-html5,}{-qa,}
+    py{36,37,38,39,310}-sphinx{50,51,52,53}{-html4,-html5}
+    py{38,39,310}-sphinx{60,61,62}{-html4,-html5}
```
agjohnson, line 4: I suppose we haven't talked much about html4 just yet.

Technically, html4 is already deprecated in the current release, and was planned to throw a hard error in our next release. So, that could be part of this release too, and in both cases, I'd probably say it's fine to stop testing the html4 writer, visually and in CI.

We'd probably leave the code around for now, but it could be removed any time after this release and when we support Sphinx 7+.

Anyways, that's all to say, should we only test on html5 writer here?

agjohnson: Also, I'd leave the `-qa` postfix targets. The docs mention these targets in verifying pull request changes:

https://sphinx-rtd-theme.readthedocs.io/en/stable/contributing.html#testing

agjohnson (commented):
Looks really close. I noted a couple of minor tox tweaks for this release.

tox.ini
humitos: > Technically, html4 is already deprecated in the current release, and was planned to throw a hard error in our next release. So, that could be part of this release too, and in both cases, I'd probably say it's fine to stop testing the html4 writer, visually and in CI.

I'm happy to remove it in _this 2.0 release_ as it was planned, actually 👍🏼

humitos:
@agjohnson I updated this PR with some changes from your feedback.

agjohnson (approved):
This looks great. Lets plan on getting a final 1.3.0 out before we merge this here, perhaps on Mon/Tues?

PR #1503, opened 2023-08-17
Use `.filename` attribute for CSS and JS files for `pathto`

### LINKED ISSUE #1502: Incompatibility with Sphinx 7.2
Since Sphinx 7.2, we changed the internal representation of Javascript and CSS files (see https://github.com/sphinx-doc/sphinx/commit/1db8cba3396cbb627d000e9d6ef497e70d769295). 

In particular, the following fails:

https://github.com/readthedocs/sphinx_rtd_theme/blob/b5833585b25358be94918f13c50530e3e9237e7e/sphinx_rtd_theme/layout.html#L35-L35

Note that the above line 

https://github.com/readthedocs/sphinx_rtd_theme/blob/b5833585b25358be94918f13c50530e3e9237e7e/sphinx_rtd_theme/layout.html#L33

works because it uses `css.filename` and not just `css` (which is no more treatable as a string).

(I don't know if other places need to be changed)

cc @humitos

humitos:
Sphinx 7.2 changed how it handles the CSS and JS files internally and they cannot be treated as string anymore. We need to access to their `.filename` attribute to get the same behavior.

Closes #1502

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -32,12 +32,12 @@
     {%- if css|attr("rel") %}
       <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
-      <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+      <link rel="stylesheet" href="{{ pathto(css.filename, 1) }}" type="text/css" />
     {%- endif %}
   {%- endfor %}
 
   {%- for cssfile in extra_css_files %}
-    <link rel="stylesheet" href="{{ pathto(cssfile, 1) }}" type="text/css" />
+    <link rel="stylesheet" href="{{ pathto(cssfile.filename, 1) }}" type="text/css" />
   {%- endfor -%}
 
   {#- FAVICON
@@ -86,7 +86,7 @@
           };
       </script>
       {%- for scriptfile in script_files %}
-        <script src="{{ pathto(scriptfile, 1) }}"></script>
+        <script src="{{ pathto(scriptfile.filename, 1) }}"></script>
       {%- endfor %}
     {%- endif %}
     <script src="{{ pathto('_static/js/theme.js', 1) }}"></script>


AA-Turner:
@humitos I believe ``css_tag`` should cover the use-cases in this theme, is there a reason why you can't use it, so that we might improve it?

A

agjohnson:
@humitos Is this backwards compatible to Sphinx 5?

sphinx_rtd_theme/layout.html
```diff
@@ -32,12 +32,12 @@
     {%- if css|attr("rel") %}
       <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
     {%- else %}
-      <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
+      <link rel="stylesheet" href="{{ pathto(css.filename, 1) }}" type="text/css" />
```
agjohnson, line 35: CI seems stuck, so I can't tell, but is this change compatible with Sphinx 5?

AA-Turner:
``.filename`` was added in Sphinx 1.6.: https://github.com/sphinx-doc/sphinx/commit/e6f8cd6c78ee4b6ea060be9edf418bd9c9ebac42

humitos:
> @humitos I believe `css_tag` should cover the use-cases in this theme, is there a reason why you can't use it, so that we might improve it?

I wasn't aware of it. I understand you refer to this:

humitos:
If  we are OK using `css_tag`, there are more code we can update. For example:

https://github.com/readthedocs/sphinx_rtd_theme/blob/2b1d3faa5d9265ead66df1e57b0525e9c144e309/sphinx_rtd_theme/layout.html#L31-L41

agjohnson:
> @agjohnson pinging you in case you know if there is a blocker about using these functions.

Not off the top of my head. If there was a blocker, it would probably be projects with old/outdated configuration.

humitos:
It seems we already did something similar to this PR in the past: https://github.com/readthedocs/sphinx_rtd_theme/pull/1101. I think we should take a look at both and merge them into the final PR 😄

PR #1504, opened 2023-08-17
Pass `includehidden=` as string instead of boolean

humitos:
Sphinx 7.2 changed its internal API for `toctree()`.

See https://github.com/sphinx-doc/sphinx/issues/11607 Related https://github.com/readthedocs/sphinx_rtd_theme/issues/1463


humitos:
This change was reverted upstream. So we don't need it anymore.

PR #1507, opened 2023-08-20
Set data-content_root for Sphinx ≥ 7.2

mitya57:
This attribute is now required by searchtools.js:
https://github.com/sphinx-doc/sphinx/blob/v7.2.2/sphinx/themes/basic/static/searchtools.js#L65

And this is what Sphinx' own `layout.html` does:
https://github.com/sphinx-doc/sphinx/blob/v7.2.2/sphinx/themes/basic/layout.html#L105

When the attribute is not set, loading search results via AJAX will be broken because the JS code will attempt to load paths like `/undefinedfoo.html`.


agjohnson (approved):
@humitos I'd be +1 on merging just this change for a 2.0rc1. This shouldn't change anything for projects using older releases of the theme.

PR #1512, opened 2023-08-23
Remove Sphinx<5 compatibility leftovers

### LINKED ISSUE #1100: arbitrary add_css_file attributes are not supported

humitos:
- Remove blocks with logic for Sphinx<5
- Always use `writer-html5` CSS class
- Use `css_tag` for better compatibility with newer Sphinx versions

* Closes #1075
* Closes #1100 
* Closes #1101 
* Closes #1503 
* Closes #1507

@agjohnson this PR is a good candidate to be included in `2.0rc2` 👍🏼 . If you approve it, I can try to make a release myself.


humitos:
I tested this visually as well with Sphinx 5, 6 and 7 and I didn't find any issue so far.

sphinx_rtd_theme/layout.html
```diff
@@ -24,29 +23,17 @@
...
-      <link rel="{{ css.rel }}" href="{{ pathto(css.filename, 1) }}" type="text/css"{% if css.title is not none %} title="{{ css.title }}"{% endif %} />
-    {%- else %}
-      <link rel="stylesheet" href="{{ pathto(css, 1) }}" type="text/css" />
-    {%- endif %}
+    {{ css_tag(css) }}
   {%- endfor %}
 
   {%- for cssfile in extra_css_files %}
-    <link rel="stylesheet" href="{{ pathto(cssfile, 1) }}" type="text/css" />
+    {{ css_tag(css) }}
```
mitya57, line 31: I think you meant this:

```suggestion
    {{ css_tag(cssfile) }}
```

humitos: Thanks

agjohnson (commented):
The changes here all seem reasonable, however I'd probably say we should plan out release structure a bit more. At very least, it makes sense to focus 2.0rc1 on just changing dependencies, not changing code yet. Then at least projects can upgrade dependencies and transitive dependencies without us changing anything major about their project output.

So for that, I'd probably say a 2.0rc1 with dependencies upgraded and a 2.1rc1 following with actual changes to our theme. In that case, we'd block this for a bit longer.

If we need any of these changes specifically for Sphinx 7 support, they can be broken away from some of the cleanup changes here. But this looks mostly like housekeeping.

agjohnson:
With `2.0` released, I think we're ready to merge this and prepare a `2.1a` release. We haven't had any more feedback on the CSS changes in the `2.0rc{1,2,3,4}` releases, so I think we're safe to continue on.

@humitos thoughts here?

humitos:
Sounds good to me! I've updated the PR to resolve the conflicts.

sphinx_rtd_theme/layout.html
```diff
@@ -24,30 +23,18 @@
...
-    <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
-  {%- endif %}
-  {%- for css_file in css_files %}
-    {%- if css_file|attr("filename") %}
-      {{ css_tag(css_file) }}
-    {%- else %}
-      <link rel="stylesheet" href="{{ pathto(css_file, 1)|escape }}" type="text/css" />
-    {%- endif %}
+  {%- for css in css_files %}
+    {{ css_tag(css) }}
```
mitya57, line 27: Please keep the old logic for `css_files`. See #1537 for the rationale.

humitos: Oh, thanks for pointing this out. I was making the same mistake again 😅

humitos: 🤦🏼 -- I'll come back to this another day because it seems I'm pretty distracted today and I can't focus clearly. Thanks again for your time reviewing this 🙏🏼

humitos: Done!

humitos:
This PR is ready to re-review and merge. I kept the injection of `css_files` code as it's on `master` which is working properly. On the other hand, I added a small comment for `extra_css_files` so we don't get confused again with that.

sphinx_rtd_theme/layout.html
mitya57: Thank you. LGTM now.

agjohnson (approved):

PR #1526, opened 2023-09-13
Addons: integrate with new beta addons flyout

### LINKED ISSUE #1523: Integrate the new addons flyout into the theme
We are working on a new addons flyout (https://github.com/readthedocs/addons) that emulates the old flyout shown at the bottom right for all the themes that are not Read the Docs Sphinx theme:

Screenshot_2023-09-07_13-47-21

However, for our own theme, we are nicely integrating it on the bottom left together with the navbar:

Screenshot_2023-09-07_13-48-22

So, we would use the integration points from the new addons to write `<readthedocs-flyout></readthedocs-flyout>` in the navbar of this theme (or using the JSON response)  and display it integrated in the same way it currently is the _old one_.

* https://github.com/readthedocs/addons/issues/60
* https://github.com/readthedocs/addons/issues/13
* https://github.com/readthedocs/addons/issues/17
* https://github.com/readthedocs/addons/issues/14

humitos:
Initial experimentation to use the `CustomEvent` triggered by the addons called
`readthedocsdataready` event (from
readthedocs/addons#64) to build the Read the Docs
flyout being integrated into the theme keeping the original look & feel.

* Related: readthedocs/addons#64
* Closes #1523

### DIFF
--- a/sphinx_rtd_theme/versions.html
+++ b/sphinx_rtd_theme/versions.html
@@ -1,34 +1,47 @@
-{% if READTHEDOCS %}
-{# Add rst-badge after rst-versions for small badge style. #}
-  <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="{{ _('Versions') }}">
+<script language="javascript" src="http://localhost:8000/readthedocs-addons.js"></script>
+<script language="javascript">
+ document.addEventListener("readthedocsdataready", function(event) {
+     const config = event.detail;
+
+     const flyout = `
+  <div class="rst-versions" data-toggle="rst-versions" role="note">
     <span class="rst-current-version" data-toggle="rst-current-version">
       <span class="fa fa-book"> Read the Docs</span>
-      v: {{ current_version }}
+      v: ${ config.versions.current.slug }
       <span class="fa fa-caret-down"></span>
     </span>
     <div class="rst-other-versions">
       <dl>
-        <dt>{{ _('Versions') }}</dt>
-        {% for slug, url in versions %}
-          <dd><a href="{{ url }}">{{ slug }}</a></dd>
-        {% endfor %}
+        <dt>Languages</dt>
+        ${ config.addons.flyout.translations.map(
+          (language) => `<dd><a href="${ language.url }">${ language.slug }</a></dd>`)}
       </dl>
       <dl>
-        <dt>{{ _('Downloads') }}</dt>
-        {% for type, url in downloads %}
-          <dd><a href="{{ url }}">{{ type }}</a></dd>
-        {% endfor %}
+        <dt>Versions</dt>
+        ${ config.addons.flyout.versions.map(
+          (version) => `<dd><a href="${ version.url }">${ version.slug }</a></dd>`)}
       </dl>
       <dl>
-        {# Translators: The phrase "Read the Docs" is not translated #}
-        <dt>{{ _('On Read the Docs') }}</dt>
+        <dt>Downloads</dt>
+        ${ config.addons.flyout.downloads.map(
+          (download) => `<dd><a href="${ download.url }">${ download.name }</a></dd>`)}
+      </dl>
+      <dl>
+        <dt>On Read the Docs</dt>
           <dd>
-            <a href="//{{ PRODUCTION_DOMAIN }}/projects/{{ slug }}/?fromdocs={{ slug }}">{{ _('Project Home') }}</a>
+            <a href="//${ config.domains.dashboard }/projects/${ config.projects.current.slug }/">Project Home</a>
           </dd>
           <dd>
-            <a href="//{{ PRODUCTION_DOMAIN }}/builds/{{ slug }}/?fromdocs={{ slug }}">{{ _('Builds') }}</a>
+            <a href="//${ config.domains.dashboard }/projects/${ config.projects.current.slug }/builds/">Builds</a>
           </dd>
       </dl>
     </div>
   </div>
-{% endif %}
+          `;
+
+     document.body.insertAdjacentHTML("beforeend", flyout);
+
+     {# TODO: disable the flyout auto-injected by `readthedocs-addons.js` somehow #}
+     document.querySelector("readthedocs-flyout").remove();
+ });
+</script>
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,6 +68,9 @@ if not 'READTHEDOCS' in os.environ:
         range(1, 100)
     ))
 
+if 'READTHEDOCS' in os.environ:
+    html_context['READTHEDOCS'] = True
+
 html_logo = "demo/static/logo-wordmark-light.svg"
 html_show_sourcelink = True
 html_favicon = "demo/static/favicon.ico"
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -17,6 +17,7 @@
 <html class="{{ sphinx_writer }}" lang="{{ lang_attr }}"{% if sphinx_version_info >= (7, 2) %} data-content_root="{{ content_root }}"{% endif %}>
 <head>
   <meta charset="utf-8" />
+  <meta name="readthedocs-api-version" content="1">
   {{- metatags }}
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
   {%- block htmltitle %}


agjohnson:
Noted on a call today, we should start trying out these changes for some of our own projects to start. This would serve as a working example for other authors, and gives us something to test against.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,47 @@
...
       </dl>
     </div>
   </div>
-{% endif %}
+          `;
+
+     document.body.insertAdjacentHTML("beforeend", flyout);
+
+     {# TODO: disable the flyout auto-injected by `readthedocs-addons.js` somehow #}
+     document.querySelector("readthedocs-flyout").remove();
```
agjohnson, line 45: I thought that we had landed on giving theme authors the ability to include and element like:

`<readthedocs-flyout disabled></readthedocs-flyout>`

Or something similar. This probably makes the most sense: it doesn't require mucking with async code, and it doesn't give theme authors undue control over addons configuration.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,47 @@
-{% if READTHEDOCS %}
-{# Add rst-badge after rst-versions for small badge style. #}
-  <div class="rst-versions" data-toggle="rst-versions" role="note" aria-label="{{ _('Versions') }}">
+<script async type="text/javascript" src="http://localhost:8000/readthedocs-addons.js"></script>
+<script type="text/javascript">
+ document.addEventListener("readthedocs-addons-data-ready", function(event) {
+     const config = event.detail;
+
+     const flyout = `
```
agjohnson, line 6: Injecting HTML as a string template does work, however eventually I'd like to give theme authors a stronger example to work from. This does remove translated strings by pushing them into JS (it is possible to translate the strings in Python via template strings embedded in JS, but this will look especially gnarly).

That's not to say that I think this needs to change, but when we go to point users towards an example they should have something more standard to work from, like an example using `document.createElement` to populate a dropdown perhaps.

sphinx_rtd_theme/versions.html
humitos: I need to remove the `readthedocs-flyout` here because it's injected by default via the addons. We need to find a way to tell the addons library to not do that when the users want to use the JSON data by themselves.

We haven't talk about the `disabled` attribute before, but probably is a good idea to communicate exactly what I'm needing here 👍🏼 

I will explore this and hopefully have something working that does not requires removing an HTML element as I'm doing now.

agjohnson: > We need to find a way to tell the addons library to not do that when the users want to use the JSON data by themselves.

Yup, I understood that part. I thought we already talked about using our elements in this way though. Theme maintainers would use a definition of our elements to have some control of the flyout, which would include the ability to disable automatic injection the flyout.

> We haven't talk about the disabled attribute before

This attribute is just an example too, I don't have strong opinions on `disabled` specifically. I don't think we talked about the specific attribute, just that this would be the pattern for theme maintainers to use to disable some of our elements/features. I'm not finding the discussion though.

sphinx_rtd_theme/versions.html
agjohnson: > What do you mean with a "stronger example to work from"?

Using string interpolation to inject a blob of HTML like this is certainly the *fastest* solution. But because of that might this pattern not be super usable or appealing for serious applications. This isn't a commonly used pattern as it's easier to author HTML outside JS entirely, similar to how I'd want to use Jinja instead of Python string interpolation. Lit is special, as the entire implementation is in JS already.

A stronger example could be using JS to provide data to existing HTML structures defined in templates. For example, populating a version selector dropdown. Personally, if I wanted a custom flyout in a theme, I would author the HTML structure in my templates and hide the custom flyout until the data event fired -- instead of creating the whole structure in JS strings.

Again though, I don't think we need to reshape things here for our testing. But I also probably would want to create a more real scenario for our example rather than build on top of this pattern going forward.

> We do not have too many strings to translate.

I wasn't necessarily looking for a solution for our code, my point was more that this is a pattern that will complicate development for theme maintainers. Theme maintainers are already doing translations in their template and we should meet them where they are.

> What are the benefits of this over a string template?

The biggest benefit is separation between layers. HTML authoring is in a single place, not split between JS and HTML template code. JS is treated like more of a data layer in this case. This is the main draw of data binding libraries like Vue, Knockout, and friends.

I don't think we need/want to show an example using data binding, so we will need to add some elements with JS. This is fine, it's just creating a large nested blob of HTML in JS that I'd avoid.

It also doesn't require theme maintainers have new patterns for localization, etc.

sphinx_rtd_theme/versions.html
humitos: I removed this line for now. We won't be automatically disabling the _official_ flyout injected for now. This is the safest decision for now and we can continue discussing this in another issue/pr in the near future.

sphinx_rtd_theme/versions.html
humitos: I followed the pattern we already have here to manage translations `{{ _("String to translate") }}` in 04d79091091e0a9289ca274bb568dadda8164f68. It works fine:

Screenshot_2024-04-15_15-55-54

humitos:
I marking this PR ready for review so I can get a new review round here. I'm happy with the current pattern and I think we should start testing it out in our own docs first by installing it from this branch: https://github.com/readthedocs/readthedocs.org/pull/11279

humitos:
I tested this in https://docs--11279.org.readthedocs.build/en/11279/ and I found we need to hide the sections if we don't have content on them:

Screenshot_2024-04-16_17-51-52

humitos:
Getting close 👍🏼 

Peek 2024-04-16 18-43

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,87 @@
...
+            </dd>
+            <dd>
+              <a href="${ config.projects.current.urls.downloads }">{{ _('Downloads') }}</a>
+            </dd>
+        </dl>
+        <dl>
+          <dt>Search</dt>
           <dd>
-            <a href="//{{ PRODUCTION_DOMAIN }}/projects/{{ slug }}/?fromdocs={{ slug }}">{{ _('Project Home') }}</a>
+            <div style="padding: 6px;">
```
agjohnson, line 58: Styles should go in sass/css instead of HTML, but is there a reason why this flyout menu is visually different from the standard flyout in the first place?

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,87 @@
...
+              <form id="flyout-search-form">
+                <input
+                  class="wy-form"
+                  type="text"
+                  name="q"
+                  aria-label="Search docs"
+                  placeholder="Search docs"
```
agjohnson, line 65:
```suggestion
                  aria-label="{{ _('Search docs') }}"
                  placeholder="{{ _('Search docs') }}"
```

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,87 @@
...
+            </dd>
+            <dd>
+              <a href="${ config.projects.current.urls.builds }">{{ _('Builds') }}</a>
+            </dd>
+            <dd>
+              <a href="${ config.projects.current.urls.downloads }">{{ _('Downloads') }}</a>
+            </dd>
+        </dl>
+        <dl>
+          <dt>Search</dt>
```
agjohnson, line 56:
```suggestion
          <dt>{{ _('Search') }}</dt>
```

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,87 @@
...
-        {# Translators: The phrase "Read the Docs" is not translated #}
-        <dt>{{ _('On Read the Docs') }}</dt>
+      <div class="injected">
+        <dl>
+          <dt>{{ _('Languages') }}</dt>
+          ${ config.projects.translations.map(
+            (translation) => `
+            <dd><a href="${ translation.urls.documentation }">${ translation.slug }</a></dd>
+          `).join("\n")}
+        </dl>
```
agjohnson, line 28: You probably want conditionals around each of these sections. You can use Lit directives if you want to bring in Lit, or you can do this in JS with string appending or breaking this up into multiple functions/methods/etc.

agjohnson (commented):
Not sure how complete this is, I didn't test this locally. But this seems on the right path, just some tweaking needed to the template.

sphinx_rtd_theme/versions.html
humitos: I'm not sure why you are saying it's visually different from the standard flyout. What flyout are you comparing against?

In this PR I'm replicating the flyout that is injected as blob HTML from the backend and I don't want to make any style/html structure change compared to that one. This is to keep as much as we can backward compatibility at this step.

I plan to make CSS/HTML structure changes in a following iteration/release that performs breaking changes (e.g. integrate the versions/languages selector into the layout itself as you have mentioned in the past).

For now, I want to keep things simple and try to make the upgrade as smooth as possible (_if_ possible, 😄 )

sphinx_rtd_theme/versions.html
humitos: Yes. I need the conditionals to avoid showing the sections when there is no content.

I don't want to introduce Lit in this PR, but it's something I want to discuss more and consider in the following iteration. Related to my previous comment, https://github.com/readthedocs/sphinx_rtd_theme/pull/1526#discussion_r1568648856

humitos:
> Not sure how complete this is, I didn't test this locally. But this seems on the right path

It's great knowing that we are going in the right direction. The PR is not _finished_ or _ready to merge_ yet, but I wanted to have a review at this stage to know if we are aligned here and it seems we are 👍🏼

sphinx_rtd_theme/versions.html
humitos: I did this in ef0a57e8526d032534356b74862dcc34012f7c03

humitos:
With the latest changes, it looks great to me! I think we are ready to merge https://github.com/readthedocs/readthedocs.org/pull/11279 so we can start testing it out in our documentation.

Screenshot_2024-04-17_13-43-59

sphinx_rtd_theme/versions.html
agjohnson: I'm comparing the flyout in this PR to the flyout you get without this PR. I was wondering why we needed the extra `padding` here.

But, looking at the standard flyout, it seems we used inline styles there too. Neither flyout should use this pattern, but it's okay that it matches for now.

agjohnson:
Yeah lets merge the docs PR :+1: This seems close enough to start testing

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,132 @@
...
+    return versionsHTML;
+  }
+
+  function renderDownloads(config) {
+    if (!Object.keys(config.versions.current.downloads).length) {
+      return "";
+    }
+    const downloadsNameDisplay = {
+      pdf: "PDF",
+      epub: "Epub",
```
agjohnson, line 52: I don't know how acceptable the other variants are here, but the epub3 spec always uses this capitalization:

```suggestion
      epub: "EPUB",
```

humitos: He, I knew you were going to comment on this 😄 . I used "Epub" here to keep consistency with the previous flyout. I if you go to our docs you will see "Epub": https://docs.readthedocs.io/en/stable/

Screenshot_2024-04-18_12-10-07

agjohnson: Gotcha, yeah we got that one wrong there too of course. Matching is okay, though I'm also :+1: on fixing this any time we want too, it's minor.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,136 @@
...
-    </span>
-    <div class="rst-other-versions">
+<!-- This style is injected ONLY on Read the Docs and it's missing. -->
+<style>
+ .rtd-current-item {
+   font-weight: bold;
+ }
+</style>
+
+<script type="text/javascript">
```
agjohnson, line 8: Similar to above, we should no longer be using inline `script` for our code, it should go in our JS sources. Inline scripts and styles are one of the largest points for XSS, we should start a beta release from a solid base.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,136 @@
...
+                <a href="${ config.projects.current.urls.builds }">{{ _('Builds') }}</a>
+              </dd>
+              <dd>
+                <a href="${ config.projects.current.urls.downloads }">{{ _('Downloads') }}</a>
+              </dd>
+            </dl>
+            <dl>
+              <dt>{{ _('Search') }}</dt>
+              <dd>
+                <div style="padding: 6px;">
```
agjohnson, line 100: This should be moved to our CSS rules, hardcoded element styles should be avoided.

agjohnson (commented):
We discussed next steps here today, and I think to consider an alpha/beta release of this, we should use the existing patterns for JS/CSS instead. Besides that, this all seems fairly close.

We talked about testing this with more internal projects, which is a good step as well. I don't see to much changing besides finishing the JS/CSS.

sphinx_rtd_theme/versions.html
```diff
@@ -1,34 +1,136 @@
...
-      v: {{ current_version }}
-      <span class="fa fa-caret-down"></span>
-    </span>
-    <div class="rst-other-versions">
+<!-- This style is injected ONLY on Read the Docs and it's missing. -->
+<style>
+ .rtd-current-item {
+   font-weight: bold;
+ }
+</style>
```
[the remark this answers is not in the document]
humitos: 5e4c12c

sphinx_rtd_theme/versions.html
humitos: dbefecb

sphinx_rtd_theme/versions.html
humitos: d16d19a

humitos:
Failing tests are unrelated with the changes on this PR. It seems the latest Sphinx/docutils version changed something in the HTML structure and we will need to update our test case.

humitos:
@agjohnson can you take a look at this PR? I'd like to move forward with an alpha release, so we can start testing it with some particular users.

agjohnson (approved):
I haven't had time to do too deep a review here, but this looks like it should be close enough. :+1:

We'll likely need to use the beta period to debug issues with this a bit, as it can be hard to code defensively in string interpolations.

humitos:
Cool! I'm merging this PR now. We can coordinate the release of an alpha/beta version after that 👍🏼

PR #1528, opened 2023-09-28
Fix #1522: fix `'str' object has no attribute 'attributes'`

### LINKED ISSUE #1522: Error with 2.0.0rc2: AttributeError("'str' object has no attribute 'attributes'")
### Problem

```python
Theme error:
An error happened in rendering the page api/aac.
Reason: AttributeError("'str' object has no attribute 'attributes'")
```

With verbose logs:

```python
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/sphinx/builders/html/__init__.py", line 1098, in handle_page
    output = self.templates.render(templatename, ctx)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/sphinx/jinja2glue.py", line 196, in render
    return self.environment.get_template(template).render(context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/sphinx/themes/basic/page.html", line 10, in top-level template code
    {%- extends "layout.html" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/sphinx_rtd_theme/layout.html", line 36, in top-level template code
    {{ css_tag(cssfile) }}
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/jinja2/sandbox.py", line 393, in call
    return __context.call(__obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lazka/.cache/pypoetry/virtualenvs/mutagen-z6FbeZMH-py3.11/lib/python3.11/site-packages/sphinx/builders/html/__init__.py", line 1197, in css_tag
    for key in sorted(css.attributes):
                      ^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'attributes'
```

### Reproducible Project

https://github.com/quodlibet/mutagen/tree/main/docs

#### Expected Results

the docs build

### Environment Info

- Python Version: 3.11.5
- Sphinx Version: 6.2.1
- RTD Theme Version: 2.0.0rc2

leplatrem:
Fix #1522

### DIFF
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -32,8 +32,9 @@
     {{ css_tag(css) }}
   {%- endfor %}
 
+  {# "extra_css_files" is a theme option and it's always a string #}
   {%- for cssfile in extra_css_files %}
-    {{ css_tag(cssfile) }}
+    <link rel="stylesheet" href="{{ pathto(cssfile, 1)|escape }}" type="text/css" />
   {%- endfor -%}
 
   {#- FAVICON
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -29,19 +29,11 @@
     <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
   {%- endif %}
   {%- for css in css_files %}
-    {% if css|attr("filename") -%}
-      {{ css_tag(css) }}
-    {%- else -%}
-      <link rel="stylesheet" href="{{ pathto(css, 1)|escape }}" type="text/css" />
-    {%- endif %}
+    {{ css_tag(css) }}
   {%- endfor %}
 
   {%- for cssfile in extra_css_files %}
-    {% if css|attr("filename") -%}
-      {{ css_tag(css) }}
-    {%- else -%}
-      <link rel="stylesheet" href="{{ pathto(css, 1)|escape }}" type="text/css" />
-    {%- endif %}
+    {{ css_tag(cssfile) }}
   {%- endfor -%}
 
   {#- FAVICON
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -28,13 +28,13 @@
     <link rel="stylesheet" href="{{ pathto('_static/' + style, 1) }}" type="text/css" />
     <link rel="stylesheet" href="{{ pathto('_static/pygments.css', 1) }}" type="text/css" />
   {%- endif %}
-  {%- for css in css_files %}
-    {{ css_tag(css) }}
+  {%- for css_file in css_files %}
+    {{ css_tag(css_file) }}
   {%- endfor %}
 
   {# "extra_css_files" is a theme option and it's always a string #}
-  {%- for cssfile in extra_css_files %}
-    <link rel="stylesheet" href="{{ pathto(cssfile, 1)|escape }}" type="text/css" />
+  {%- for css_file in extra_css_files %}
+    <link rel="stylesheet" href="{{ pathto(css_file, 1)|escape }}" type="text/css" />
   {%- endfor -%}
 
   {#- FAVICON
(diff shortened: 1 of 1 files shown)


agjohnson:
Thanks for the fix here @leplatrem !

This was recently changed, in #1519, though it wasn't clear if there was going to be any side effects from the change. I guess there are side effects, which isn't too surprising.

This is reverting to what we had before, so should definitely be safe. But I will first try to understand what is happening in #1522, or why the `css_tag` fails here. It was my understanding the `css_tag` call would be safer, but I'm guessing it comes down to the Sphinx configuration.

I'll want to aim to merge this in, reverting that change, before next week. I'm planning on a 2.0 final release on Monday or Tues.

leplatrem:
Excellent, thank you! 💚

agjohnson (approved):
This is revert to the old logic, it should be okay. If I don't come up with another fix or reason to keep the current logic, I'll look to merge this early next week for the 2.0 release. We can revisit the reverted change again in another release if we need.

sphinx_rtd_theme/layout.html
```diff
@@ -29,11 +29,19 @@
...
   {%- for css in css_files %}
-    {{ css_tag(css) }}
+    {% if css|attr("filename") -%}
+      {{ css_tag(css) }}
+    {%- else -%}
+      <link rel="stylesheet" href="{{ pathto(css, 1)|escape }}" type="text/css" />
+    {%- endif %}
   {%- endfor %}
 
   {%- for cssfile in extra_css_files %}
```
mitya57, line 39: Right now it won't work, as the loop variable is named `cssfile`, but below you are using `css`.

Please either do this:
```suggestion
  {%- for css in extra_css_files %}
```

sphinx_rtd_theme/layout.html
```diff
@@ -29,11 +29,19 @@
...
-    {{ css_tag(cssfile) }}
+    {% if css|attr("filename") -%}
+      {{ css_tag(css) }}
+    {%- else -%}
+      <link rel="stylesheet" href="{{ pathto(css, 1)|escape }}" type="text/css" />
```
mitya57, line 43: OR this:
```suggestion
    {% if cssfile|attr("filename") -%}
      {{ css_tag(cssfile) }}
    {%- else -%}
      <link rel="stylesheet" href="{{ pathto(cssfile, 1)|escape }}" type="text/css" />
```

But not both, of course.

mitya57: Also, it looks like `extra_css_files` is not a standard feature of Sphinx, but just something that can be explicitly included in HTML context, like here in @lazka's project:
https://github.com/quodlibet/mutagen/blob/release-1.47.0/docs/conf.py#L45-L49

So, I believe, it will never have a `filename` attribute, so we can further simplify this code block to unconditional:
```html
<link rel="stylesheet" href="{{ pathto(cssfile, 1)|escape }}" type="text/css" />
```

humitos:
I updated this PR to take into consideration the review comments from @mitya57. It should work for all the Sphinx versions we support now. I'd be happy to merge this PR and release a new `rc` version to PyPI so people can continue testing it.

agjohnson (approved):
This fix looks good to me, I'll get an RC prepared this afternoon so we can test the fix further.

I pushed up a change just standardizing the variable names, will wait on that.

humitos (approved):
Looks good to me.

agjohnson:
2.0.0rc3 is now out with this fix. Thanks everyone!

PR #1549, opened 2024-01-17
Fix equation numbering

eebasso:
Modify the MathJax SASS file to have equation numbers appear to the right of displayed formulas. This resolves the long-standing issue #301 while still having the header links appear, which improves upon the solution given by https://github.com/readthedocs/sphinx_rtd_theme/issues/301#issuecomment-1205755904 . These changes also avoid the use of any margin manipulations, which can lead to odd behavior when hovering vs not.

An unresolved issue is the vertical alignment of the equation numbers relative to the displayed formulas. This PR leads to an upper alignment, but a middle alignment seems preferable to most users. Hopefully, PR #383 can follow up on this and resolve the vertical alignment problem.

Regardless, I think these changes are a strict improvement over the status-quo.


eebasso:
I did find one way to obtain the vertical alignment of the equation numbers using a Y translation trick of
```
span.eqno
  float: right
  position: relative
  transform: translateY(50%)

  .headerlink
    position: relative
    z-index: 1
```
but this lead to the header links disappearing and the equation numbers overlapping the formulas for sufficiently narrow viewing windows. Therefore, I decided against adding this to the PR.

eebasso:
Example from the documentation showing the header link. Note that equation formula remains completely unmoved even when hovering over it:
no-hover
with-hover

Another example with a more narrow window width:
narrow-no-hover
narrow-with-hover

wandadars:
Please let them merge this. The equation numbering default is just comically bad at this point.

SebastienGarciaAissa:
I also support the merging of this fix. 

The equation numbering is the only bad feature I found in the rtd-theme. But if not fixed I will need to change for another theme, even if I think that this theme is perfect on all other points.

PR #1573, opened 2024-07-22
Release 2.1.0rc1

humitos:
I wasn't able to update translations because it fails with non-zero exit when running `python setup.py update_translations`.


humitos:
BTW, I followed the steps from https://sphinx-rtd-theme.readthedocs.io/en/stable/contributing.html#releasing-the-theme

agjohnson (approved):

PR #1574, opened 2024-07-23
Docs: update releasing

humitos:


docs/contributing.rst
```diff
@@ -227,6 +227,7 @@ Releasing the theme
 
 To release a new version of the theme, core team will take the following steps:
 
+#. Install the required depedencies with ``pip install bump2version wheel twine``.
```
agjohnson, line 230: We already have these defined in the package configuration as well:

https://github.com/readthedocs/sphinx_rtd_theme/blob/68a89bb07e71fbe222718f680d581dfcd0359769/setup.cfg#L55-L59

Twine should be there though.

humitos: Done in e09da77

setup.cfg
```diff
@@ -57,6 +57,7 @@ dev =
 	sphinxcontrib-httpdomain
 	bump2version
 	wheel
+  twine
```
agjohnson, line 60: I don't think indentation is syntactic in cfg files, but in case it is, the indentation is different here.

agjohnson (approved):

setup.cfg
humitos: Meh, tabs vs. spaces 🔥

agjohnson: Speaking of bump2version, I think it rewrites this file each time, converting whitespace

PR #1575, opened 2024-07-24
Tests: `sphinxdev` tox environment installs Sphinx from `master`

humitos:
It seems we had a bug in the name and it wasn't installing Sphinx from `master`. I expect this test to fail now because we pin `Sphinx<8`, but that's fine for now.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ deps =
     sphinx71: Sphinx>=7.1,<7.2
     sphinx72: Sphinx>=7.2,<7.3
     sphinxlatest: Sphinx
-    dev: https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip
+    sphinxdev: https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip
 allowlist_externals =
     echo
 commands =


humitos:
It fails as expected:

```
The conflict is caused by:
    The user requested sphinx 8.0.0 (from https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip)
    sphinx-rtd-theme 2.1.0rc1 depends on sphinx<8 and >=5
```

tox.ini
```diff
@@ -24,7 +24,7 @@ deps =
     sphinx71: Sphinx>=7.1,<7.2
     sphinx72: Sphinx>=7.2,<7.3
     sphinxlatest: Sphinx
-    dev: https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip
+    sphinxdev: https://github.com/sphinx-doc/sphinx/archive/refs/heads/master.zip
```
agjohnson, line 27: Not sure I follow this change. Isn't the target of the `envlist` still `sphinxdev`?

That is: `py311-sphinxdev-qa` not `py311-dev-qa`

agjohnson:
> It seems we had a bug in the name and it wasn't installing Sphinx from master

Where is this occuring?

tox.ini
agjohnson: Ah I understand what is happening here now, this does look correct

agjohnson (approved):
I see the failure now, this looks good :+1:

PR #1580, opened 2024-07-31
Theme option `flyout` to enable/disable theme's flyout

### LINKED ISSUE #1579: Add theme option to disable theme's flyout
By default, when building documentation on Read the Docs, this theme will add a flyout at the bottom left integrated in the navbar. However, there will be people that will want to disable this flyout and _only_ use the Read the Docs one floating at bottom right.

We should provide a theme config to allow users disable/enable the flyout the theme controls.

humitos:
Add a theme option to enable/disable the theme's flyout integrated at the bottom left into the navigation bar.

Closes #1579

### DIFF
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -179,6 +179,8 @@ Miscellaneous options
 
     Specify how to display the flyout (language and version selector).
     This can be either ``attached`` or ``hidden``.
+    ``attached`` means that it will show the flyout in the bottom of the sidebar.
+    You will need to disable the default `Read the Docs flyout <https://docs.readthedocs.io/en/stable/flyout-menu.html>`_ in order to not have 2 flyouts showing.
 
     :type: str
     :default: ``hidden``
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -21,7 +21,7 @@ For example:
         'style_external_links': False,
         'vcs_pageview_mode': '',
         'style_nav_header_background': 'white',
-        'flyout_display': 'attached',
+        'flyout_display': 'hidden',
         # Toc options
         'collapse_navigation': True,
         'sticky_navigation': True,
@@ -181,7 +181,7 @@ Miscellaneous options
     This can be either ``attached`` or ``hidden``.
 
     :type: str
-    :default: ``attached``
+    :default: ``hidden``
 
 
 File-wide metadata
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -21,7 +21,7 @@ For example:
         'style_external_links': False,
         'vcs_pageview_mode': '',
         'style_nav_header_background': 'white',
-        'flyout': False,
+        'flyout_display': 'attached',
         # Toc options
         'collapse_navigation': True,
         'sticky_navigation': True,
@@ -175,13 +175,13 @@ Miscellaneous options
     :type: string
     :default: ``#2980B9``
 
-.. confval:: flyout
+.. confval:: flyout_display
 
-    Render the flyout (language and version selector) at the bottom left
-    integrated in the navigation bar.
+    Specify how to display the flyout (language and version selector).
+    This can be either ``attached`` or ``hidden``.
 
-    :type: boolean
-    :default: ``True``
+    :type: str
+    :default: ``attached``
 
 
 File-wide metadata
--- a/docs/configuring.rst
+++ b/docs/configuring.rst
@@ -21,6 +21,7 @@ For example:
         'style_external_links': False,
         'vcs_pageview_mode': '',
         'style_nav_header_background': 'white',
+        'flyout': False,
         # Toc options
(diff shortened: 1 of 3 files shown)


docs/configuring.rst
```diff
@@ -174,6 +175,14 @@ Miscellaneous options
     :type: string
     :default: ``#2980B9``
 
+.. confval:: flyout
```
agjohnson, line 178: A config option `flyout` feels ambiguous to me -- that is, does `flyout=false` turn off the flyout entirely?

Maybe we want this to be `flyout_position` or `flyout_display` to mimic the CSS property?

- `flyout_display: floating`
- `flyout_display: hidden`
- `flyout_display: docked`

Does that make sense?

agjohnson (commented):
Giving users this control makes sense :+1:

I think the configuration option name and values could be tuned to be more clear though.

docs/configuring.rst
humitos: > does `flyout=false` turn off the flyout entirely?

yes

> `flyout_display` to mimic the CSS property?

It makes sense, yes. For now, we will have only two options `hidden` (no flyout) and `navbar_bottom` (the current integration).

I suppose we will use this same config in the future to change the integration type: `navbar_top` or similar ones. Using a string will give us plenty of options here.

agjohnson: So I assumed this was `flyout=false` would give the addons floating right flyout and `flyout=true` is the addons docked flyout in the navigation bar.

I'd probably avoid mixing positioning with `flyout_display`. I think we probably want a separate option for where the flyout is positioned: `flyout_display = "attached"; flyout_position = "bottom"` and `flyout_display = "floating"; flyout_position = "top right"`.

And we're only implementing `flyout_display = (hidden|attached)` for now. Or floating (default addons behavior) too?

I don't know what our plan is to control this from addons configuration though, it feels like there is some overlap, especially with floating positioning.

humitos: You are right here, just with a small change, however. Note that with the theme configs we are controlling _only_ the flyout injected by the theme itself --we won't be controlling the Read the Docs flyout addons. That's controlled from the project's dashboard.

So, in the theme, we can implement as many display/position as we want, but those has to be implemented in the theme and won't be related with the different display/position we could eventually implement from Read the Docs addons.

Theme options won't have any effect on Read the Docs options. They are separated.

agjohnson: Yeah, though this is getting back to what we have talked about with some intermediate customization pattern. I'm seeing this becoming more important for us to define, especially given we're not seeing many maintainers jump into full addons reimplementation yet.

We should continue the discussion about intermediate patterns:

- https://github.com/readthedocs/addons/issues/51

Floating positioning is an example here, but this probably is not super useful for the theme to redefine configuration for.

humitos: Yeah, we should continue talking about that customization pattern. In this PR, I'm only wanting to have a setting to disable the "feature is enabled on Read the Docs automatically" and currently can't be disabled. So, building locally and building on Read the Docs leads to different results.

People wanting to use _only the flyout addons_ can't currently do that because the theme always injects the flyout in the navbar.

agjohnson: To clarify, I feel like this is still the plan:

> And we're only implementing flyout_display = (hidden|attached) for now.

We can discuss intermediate customization patterns in https://github.com/readthedocs/addons/issues/51 and decide whether that means there is a `floating` option here as well.

Or rather, if it feels like `floating` might be an option here at some point, the above plan is more future forward than `flyout: (true|false)`.

humitos: I updated the PR to use `flyout_display: (hidden|attached)` 👍🏼

humitos:
We talked about

* Making the default behavior to be `hidden` by default.
* Integrate the version and language selector in the theme itself -- outside the _flyout_  concept -- is a good goal for 4.0.

humitos:
I made this changes. @agjohnson let me know if this is ready to get merged. After merging it, I will release `2.1.0rc2`

docs/configuring.rst
```diff
@@ -174,6 +175,14 @@ Miscellaneous options
     :type: string
     :default: ``#2980B9``
 
+.. confval:: flyout_display
+
+    Specify how to display the flyout (language and version selector).
+    This can be either ``attached`` or ``hidden``.
```
ericholscher, line 181: Would be useful to define what `attached` means.

```suggestion
    This can be either ``attached`` or ``hidden``.
    ``attached`` means that it will show the flyout in the bottom of the sidebar.
    You will need to disable the default `Read the Docs flyout <https://docs.readthedocs.io/en/stable/flyout-menu.html>`_ in order to not have 2 flyouts showing.
```

ericholscher (approved):
Looks good to me 👍

agjohnson (approved):

PR #1581, opened 2024-07-31
Populate `html_context` with `READTHEDOCS_*` environment variables

### LINKED ISSUE #1578: Read `READTHEDOCS_PROJECT` from environment variables
This line doesn't work because Read the Docs pass `READTHEDOCS_PROJECT` as environment variable, which is not directly accessible from the HTML context:

https://github.com/readthedocs/sphinx_rtd_theme/blob/7b141ea90ac320bb926a6a8e14f930c9b56ca94e/sphinx_rtd_theme/static/js/versions.js_t#L105

humitos:
Closes #1578
Related: https://github.com/readthedocs/readthedocs.org/issues/11474

### DIFF
--- a/sphinx_rtd_theme/__init__.py
+++ b/sphinx_rtd_theme/__init__.py
@@ -41,7 +41,7 @@ def extend_html_context(app, pagename, templatename, context, doctree):
      # https://docs.readthedocs.io/en/stable/reference/environment-variables.html
      context['READTHEDOCS'] = os.environ.get("READTHEDOCS", False) == "True"
      if context['READTHEDOCS']:
-         for key, value in os.environ:
+         for key, value in os.environ.items():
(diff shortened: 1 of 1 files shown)


humitos:
I put this on 3.0, but it could definitely go into the 2.x series, tho. We haven't done the final release for our rc1 yet.

agjohnson (approved):
I don't have a preference whether this is in the 2 or 3 series release, but maybe it does make sense in 2, in case there is any reason why users can't use the 3 series release, or don't upgrade for any reason?

humitos:
> but maybe it does make sense in 2

Cool, I will fix the tests and do a `2.1.0rc2` in the following days.

humitos:
I'm expecting the tests for Sphinx 8.x to fail because we don't support it _yet_.

PR #1589, opened 2024-08-12
Docs: do not install as Sphinx extension

### LINKED ISSUE #1222: Docs: install theme as Sphinx extension may not be required
As @evildmp noted, the installation docs say:

https://github.com/readthedocs/sphinx_rtd_theme/blob/b07560bf97dad3a4266f6145bd4b662ac708ab00/docs/installing.rst#L11-L22

However, just doing `html_theme = 'sphinx_rtd_theme'` works for most cases. There's a note below:

https://github.com/readthedocs/sphinx_rtd_theme/blob/b07560bf97dad3a4266f6145bd4b662ac708ab00/docs/installing.rst#L25-L29

but I don't understand what it means. Perhaps projects that want to opt-out from localization can use the simplified procedure? What am I missing here?

humitos:
Closes #1222

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -10,14 +10,9 @@ Install the ``sphinx_rtd_theme`` package (or add it to your ``requirements.txt``
 
     $ pip install sphinx_rtd_theme
 
-In your Sphinx project's ``conf.py`` file, add ``sphinx_rtd_theme`` to the list of enabled extensions and as the active theme:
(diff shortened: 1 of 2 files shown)


humitos:
Test failure is expected, `sphinx-rtd-theme 2.1.0rc1 depends on sphinx<8 and >=5`

humitos:
I don't 100% understand the Sphinx internals, but it seems we cannot remove the installation as an extension: https://github.com/readthedocs/sphinx_rtd_theme/issues/1434#issuecomment-2289443380

PR #1595, opened 2024-08-19
Docs: update the installation note

humitos:
Translated string work with/without the theme added into the ``extensions``. However, it's still required to be inside the extensions because of ``sphinxcontrib-jquery``.

I'm keeping the note, but changing it content.

- Refs https://github.com/readthedocs/sphinx_rtd_theme/issues/1222


agjohnson (approved):
This matches my worries about removing the explicit extensions configuration. I'm guessing there is some way around this behavior, and perhaps it's load order, but the jquery extension specifically did benefit from explicit extension configuration.

PR #1601, opened 2024-09-17
Show version/language selectors below the title

humitos:
Allow users to decide whether or not showing a version and language selectors below the title of the documentation.

I'm adding two new configs here:

* `version_selector`
* `language_selector`

I'm re-using the code written some time ago in #438 for CSS styling.

**Ideally, we should include this in the 3.0.0 final version and release it on October 7th as planned**. That way, people using our own theme will get a nice and integrated version/language selector without the need of using `flyout_display=attached` which re-implements _all the flyout_ and requires disabling the floating one to avoid duplications.

### Example

Peek 2024-09-17 16-30

* Live example: https://test-builds.readthedocs.io/en/ver-lang-selectors/


docs/configuring.rst
```diff
@@ -192,6 +198,24 @@ Miscellaneous options
     :type: str
     :default: ``hidden``
 
+.. confval:: version_selector
+
+    Display a version selector below the title.
+    This feature makes usage of `Read the Docs Addons <https://docs.readthedocs.io/page/addons.html>`_ for this,
+    so it's required the documentation to be hosted on Read the Docs.
```
ericholscher, line 205: I do wonder about adapting this so it can use a JSON file as well in a future version, but seems fine for now.

ericholscher:
```suggestion
    This feature makes usage of `Read the Docs Addons <https://docs.readthedocs.io/page/addons.html>`_ for this,
    so it's required the documentation to be hosted on Read the Docs.
    It only appears when there are more than 1 active version. 
```

docs/configuring.rst
```diff
@@ -192,6 +198,24 @@ Miscellaneous options
...
+
+    :type: boolean
+    :default: ``True``
+
+.. confval:: language_selector
+
+    Display a language selector below the title.
+    This feature makes usage of `Read the Docs Addons <https://docs.readthedocs.io/page/addons.html>`_ for this,
+    so it's required the documentation to be serverd on Read the Docs.
```
ericholscher, line 214:
```suggestion
    so it's required the documentation to be served on Read the Docs.
    It only appears when there is more than 1 active language. 
```

ericholscher (commented):
I reviewed the Python & JS, which looks pretty solid. Not knowledgable enough on the CSS :)

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -123,3 +124,68 @@ function renderLanguages(config) {
...
+              ${ version.slug }
+          </option>`
+      ).join("\n") }
+    </select>
+  `;
+
+  dropdownSelectors.insertAdjacentHTML("beforeend", versionSelect);
+  const versionSelector = document.querySelector("#readthedocs-version-selector");
+  versionSelector.addEventListener("change", onSelectorSwitch);
+  versionSelector.innerHTML = versionSelect;
```
agjohnson, line 160: I'm having a bit of trouble figuring out why some of this is needed. It feels like adding, querying, then adding the elements again could be simplified into something easier to follow.

The reason I hardcoded multiple inner divs in `div.switch-menus` is so that we already have an easy target for this HTML:

```suggestion
  versionSwitch.innerHTML = versionSelect;
  versionSwitch.firstChild.addEventListener("change", onSelectorSwitch);
```

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -123,3 +124,68 @@ function renderLanguages(config) {
...
+
+
+  {%- if theme_language_selector %}
+  if (config.projects.translations.length) {
+    // Add the current language to the options on the selector
+    let languages = config.projects.translations.concat(config.projects.current);
+    languages = languages.sort((a, b) => a.language.name.localeCompare(b.language.name));
+
+    const languageSelect= `
+      <select id="readthedocs-version-selector">
```
agjohnson, line 171: Same note on id usage here:

```suggestion
      <select>
```

CSS selector is `div.switch-menus > div.version-switch > select` and has plenty of specificity.

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -123,3 +124,68 @@ function renderLanguages(config) {
...
+                  ${ language.language.name }
+              </option>`
+        ).join("\n") }
+       </select>
+    `;
+
+    dropdownSelectors.insertAdjacentHTML("beforeend", languageSelect);
+    const languageSelector = document.querySelector("#readthedocs-language-selector");
+    languageSelector.addEventListener("change", onSelectorSwitch);
+    languageSelector.innerHTML = languageSelect;
```
agjohnson, line 187:
```suggestion
    switchLanguage.innerHTML = languageSelect;
    switchLanguage.firstChild.addEventListener("change", onSelectorSwitch);
```

src/sass/_theme_layout.sass
```diff
@@ -231,10 +231,48 @@ html
...
+      font-weight: normal
+      color: rgba(255, 255, 255, .3)
+      // Allow for padding
+      appearance: none
+      -webkit-appearance: none
+      -moz-appearance: none
+      &:hover, &:active, &:focus
+        background: rgba(255, 255, 255, .1)
+        color: rgba(255, 255, 255, .5)
+      &:after
```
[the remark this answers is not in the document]
agjohnson: It definitely needs something visual, it doesn't look like you are supposed to click it yet. This is where most frameworks recreate a custom select, like you're pointing at.

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -123,3 +124,68 @@ function renderLanguages(config) {
...
+  const dropdownSelectors = document.querySelector("div.dropdown-selectors");
+
+  {%- if theme_version_selector %}
+  const versionSelect = `
+    <select id="readthedocs-version-selector">
+      ${ config.versions.active.map(
+          (version) => `
+          <option
+              value="${ version.slug }"
+              ${ config.versions.current.slug === version.slug ? 'selected="selected"' : '' }
```
agjohnson, line 149: A couple things this needs to support:

- Either don't show this menu on PR builds or show this menu with the PR build as the selected version, so that there isn't a conflicting menu on PR builds

image

- Default to the default branch. Otherwise it just shows the first option (`0.5.0` in this case, which is not good UX)

humitos: > show this menu with the PR build as the selected version

I followed this approach.

sphinx_rtd_theme/static/js/versions.js_t
humitos: I followed this approach and it worked fine.

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -1,3 +1,4 @@
+{%- if theme_flyout_display == "attached" %}
```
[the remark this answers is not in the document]
humitos: I followed this pattern and I passed `prettier` on this file as well.

humitos:
I've addressed all the feedback received. This is how it looks now:

Peek 2024-09-19 11-59

I think it's good enough to me. Let me know if you have any other feedback.

ericholscher:
This looks like a great start. I'd love to have some indication of what the strings are (eg. the branch and language icons next to them), but that can be added in the next iteration.

agjohnson:
This looks good. I have a lot of CSS fixes that would help this, it's probably easier to just show the change though.

The width and padding could use tuning, as the field width look unintentional stretched to the nav bar width

image

I wouldn't add icons here yet, I think I'd play around with making this a combination dropdown field instead of native select elements next. Making this into a custom dropdown gives us some room to explain default versions/languages, add icons for versions/prs/languages, and add more supporting context to the UI that is missing with just a `select`.

agjohnson:
@humitos how are you testing this locally?

humitos:
I'm not 😅. I'm just pushing and re building the test-builds version

agjohnson:
Roger. What would we need to do to make this locally reproducible? Remote testing seems okay for an initial test, but we don't want to hide these styles from other contributors or things could be prone to breaking

humitos:
Yeah, I didn't want to jump to deep in the mocking for this just yet.

I suppose we can follow the pattern we had for multiple versions on debugging but in this case fire the custom event with mocked data. Shouldn't be hard

agjohnson:
I threw up my changes at #1603 because of the issues with testing locally. It changes the fields to variable width instead of hardcoding them to be 10em.

humitos:
> Mock the custom event with placeholder data?

@agjohnson I've done this at https://github.com/readthedocs/sphinx_rtd_theme/pull/1606

humitos:
@agjohnson this PR is ready to reveiw and merge 👍🏼

agjohnson (approved):
Looks good for a test! I'm curious to see what this change does in the wild, I've wanted this for a while.

cristian64:
With `display_version` removed, what's the workflow now for showing a single, static version in the navigation bar for docs that are _not_ hosted on Read the Docs?

humitos:
@cristian64 Hi, another user also asked this question and opened an issue at #1624. Please, subscribe there get updates on this topic. We will be working in a solution for this case.

PR #1603, opened 2024-09-19
Tune version and language selector styles

agjohnson:
UI issues:

- [x] Support shrinking select down to variable width, so dropdown icon isn't floating off to the right
- [ ] Without using the dropdowns, it seems the padding below the version was lost
- [x] FOUC when loading the lists, only the dropdown icons show


agjohnson:
image

Inline looks nicer, but it will wrap if the options are long:

image

agjohnson:
fe58efc gets rid of the FOUC-ish effect while we are still fetching the API response:

[Screencast from 2024-09-19 12-09-21.webm](https://github.com/user-attachments/assets/6e8a74a5-a981-4864-94d4-db4dacd27d1e)

humitos:
Looks awesome 😎

agjohnson:
I can't reproduce the second issue there. I'm not sure why it was happening, but it was still displaying the version as a plain text element and it had no padding. This is an issue stemming from local development being a little hard at the moment maybe. We can maybe disregard it for now.

humitos (approved):
Looks great 💯

I'm merging this PR into my own branch so we can re-built it on production. After that, I will do a 3.0.0rc2 so we can install it other places as well to test it more broadly.

PR #1606, opened 2024-09-20
Debug: trigger Read the Docs Addons event to test/debug locally

### LINKED ISSUE #1605: Mock the Read the Docs `CustomEvent` to render the selectors
In https://github.com/readthedocs/sphinx_rtd_theme/pull/1601 we added a version/language selector below the title using the Read the Docs `CustomEvent`. We want to figure it out a  way to be able to test this locally.

humitos:
Use a mocked JSON to trigger the event that Read the Docs Addons triggers when serving docs on production.

You can build the docs via:

```
sphinx-autobuild -E docs docs/_build/html
```

and then open http://127.0.0.1:8000 locally to see it in action.

Closes #1605


agjohnson:
Are you not using `npm run dev`? There shouldn't be any manual commands necessary for development like that. The changes here look like they'll be usable either way, `npm run dev` doesn't do anything special during the Sphinx build.

docs/_static/debug.js
```diff
@@ -4,4 +4,40 @@ $(function () {
   $("[data-toggle='rst-debug-badge']").on("click", function () {
     $("[data-toggle='rst-versions']").toggleClass("rst-badge");
   });
-})
+});
+
+
+const EVENT_READTHEDOCS_ADDONS_DATA_READY = "readthedocs-addons-data-ready";
+
+fetch("_static/addons.mocked.json", { method: "GET" })
```
agjohnson, line 12: Seems this file is missing from this PR

agjohnson (approved):
Looks good besides the missing data :+1:

docs/_static/debug.js
humitos: Pushed!

humitos:
> Are you not using `npm run dev`?

Interesting. I was not using this. I just tested it and it works great 💯

PR #1610, opened 2024-10-01
Trigger search modal on "Search docs" input focus

humitos:
This is a regression introduced in
https://github.com/readthedocs/sphinx_rtd_theme/pull/1601

This PR triggers the search modal when clicking on the "Search docs" input from
- inside the flyout if present
- the top left navbar


ericholscher (approved):
Seems that we just moved this outside the if block? Is there any more context on why that fixes it?

humitos:
Yes. We moved it away from the first block because that block is only executed when `theme_flyout_display = "attached"`, which is not executed by default.

agjohnson (approved):

humitos:
I released 3.0.0rc4 with this work: https://pypi.org/project/sphinx-rtd-theme/3.0.0rc4/

PR #1616, opened 2024-10-08
Mention "Hosted on Read the Docs" in the footer

### LINKED ISSUE #1587: Split flyout integration into different selectors
We've been talking about splitting the integration of the flyout into different selectors/blocks and place them outside the flyout. In fact, the concept of the flyout may not apply anymore for Sphinx theme after moving this information outside the current flyout -- leaving the concept of flyout only to the one injected by Read the Docs addons.

Using CPython as an example, the version and language selectors could be placed at the top left, below the project's logo/name:

 
Screenshot_2024-08-09_16-12-54

Besides, the "On Read the Docs" section from the original flyout could be moved to the footer:

Screenshot_2024-08-09_16-13-09

These are just examples, but I think it reflects how theme authors will integrate with us in a better way without the following the flyout concept.

humitos:
Screenshot_2024-10-08_15-31-26

Closes #1587

### DIFF
--- a/sphinx_rtd_theme/footer.html
+++ b/sphinx_rtd_theme/footer.html
@@ -57,6 +57,10 @@
     {% trans %}provided by {{ readthedocs_web }}{% endtrans %}.
   {% endif %}
 
+  {% if READTHEDOCS %}
+    Hosted on <a href="https://about.readthedocs.com/">Read the Docs</a>.
+  {% endif %}
+
   {%- block extrafooter %} {% endblock %}
 
 </footer>


agjohnson:
Not sure if I understand the need for this link from the underlying issue, but it feels a bit redundant or maybe even spammy. Is there a reason we need a second link to us?

humitos:
We don't need a second link. I removed the old one in 14d07f9

agjohnson:
To clarify, it's not just the link, it's the text that feels redundant/spammy. This will still render as:

```
Built with Sphinx using a theme provided by Read the Docs. Hosted by Read the Docs.
```

This still feels redundant, maybe spammy -- it doesn't feel to me like we need our name in there twice on every page.

```
Built with Sphinx and hosted on Read the Docs.
```

That seems fine.

I'm also fine not mentioning hosting. I think it's a surprise to users when there is conditional content like this -- locally this string isn't there but our builds produce different output.

sphinx_rtd_theme/footer.html
```diff
@@ -48,13 +48,16 @@
...
     {%- trans sphinx_web=sphinx_web, readthedocs_web=readthedocs_web %}Built with {{ sphinx_web }} using a{% endtrans %}
     {#- Translators: "theme" refers to a theme for Sphinx, which alters the appearance of the generated documentation #}
     <a href="https://github.com/readthedocs/sphinx_rtd_theme">{% trans %}theme{% endtrans %}</a>
     {#- Translators: this is always used as "provided by Read the Docs", and should not imply Read the Docs is an author of the generated documentation. #}
-    {% trans %}provided by {{ readthedocs_web }}{% endtrans %}.
+    {% trans %}provided by Read the Docs{% endtrans %}.
+  {% endif %}
+
+  {% if READTHEDOCS %}
+    Hosted on <a href="https://about.readthedocs.com/">Read the Docs</a>.
```
agjohnson, line 60: This should probably be wrapped by `show_sphinx` or a specific configuration option. If the user selected `show_sphinx = False`, they will probably not like that we injected different content in the same space.

PR #1618, opened 2024-10-10
CircleCI: Add Python 3.13 to the testing

cclauss:
* https://devguide.python.org/versions/
* https://hub.docker.com/r/cimg/python/tags?name=3.13
* https://circleci.com/developer/images/image/cimg/python#image-tags

~Blocked by:~
* ~CircleCI-Public/cimg-python#252~


humitos (commented):
Thanks for this PR! 🙏🏼 

I think this also needs some updates in the `tox.ini` file to add Python 3.13 there as well, right?

PR #1622, opened 2024-10-18
Sidebar should not be floating on mobile

kartben:
When sidebar is set to 100% width on mobile, it should also be switched back to being non-floating.


kartben:
The CI error seems to be unrelated to my change?

kartben:
@humitos ping?

humitos:
I'm not sure to understand what issue this PR solves. Can you expand on that? Maybe attaching a screenshot / video will help here as well.

kartben:
Basically, if the aside keeps floating when the screen is narrow, the element that follows ends up "swallowing" it, whereas it should instead be back to a "non-floating" layout in this particular situation
https://github.com/user-attachments/assets/4be5582c-4940-478d-ab36-a9840b487d61

Depending on how the element that immediately follows the aside is styled, this can cause very visible glitches. It might not be directly reproducible in RTD theme proper, but it can impact people who tweak the theme downstream

For example: 

```css
section > p:first-of-type {
    margin-inline: -10px;
    border-inline-start: 5px black solid;
    padding-inline-start: 10px;
}
```

The incorrect layout would be:

image

While making sure the aside does not float anymore gives:

image

Again, I appreciate that this might not be directly visible/reproducible in RTD theme but IMO it would still be "typographically incorrect" to not implement the fix.

Thanks!

kartben:
@agjohnson could you help review? thanks!

agjohnson (approved):
This looks good to me. 👍 Thanks for the fix @kartben!

agjohnson:
Not sure what is going on with the test failure, it looks like CircleCI is misbehaving though. I think this is safe to merge.

PR #1626, opened 2024-10-31
Allow disabling selectors

### LINKED ISSUE #1620: Version selector appears even with 'version_selector': False
### Problem

The version selector appears on the top left even if there is only one version and version_selector is set to False in html_theme_options.

### Reproducible Project

Project documentation: https://esieabot.readthedocs.io/en/latest/
Project documentation conf.py file: https://gitlab.esiea.fr/esieabot/esieabot-docs/rtd/-/blob/main/docs/source/conf.py

#### Error Logs/Results

image

#### Expected Results

It think it was expected that the version selector shouldn't be there at all?

### Environment Info

- Python Version: 3.12
- Sphinx Version: 8.3.1
- RTD Theme Version: latest (3.0.1)

humitos:
There was a bug with how we render a Python bool value into JavaScript. This PR fixes that issue.

Closes #1620

### DIFF
--- a/sphinx_rtd_theme/static/js/versions.js_t
+++ b/sphinx_rtd_theme/static/js/versions.js_t
@@ -1,6 +1,6 @@
 const themeFlyoutDisplay = "{{ theme_flyout_display }}";
-const themeVersionSelector = "{{ theme_version_selector }}";
-const themeLanguageSelector = "{{ theme_language_selector }}";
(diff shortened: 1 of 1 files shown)


agjohnson (approved):
I didn't catch that either of these were boolean, I assumed they were query selector strings. This is where I prefer to use naming to illustration the variable use: `hasVersionSelector` and `hasLanguageSelector` are more obviously boolean.

PR #1629, opened 2024-11-11
Attached flyout: show the current translation

### LINKED ISSUE #1621: Flyout language selector doesn't show current language after upgrading to RTD Theme 3.0.0
### Problem

I am not sure if this issue is related to RTD Theme or new addon system. Right now, flyout language selector doesn't show current language. Before upgrading to 3.0.0 (I was using 2.0.0), all language choices were shown. 

### Reproducible Project

Project documentation: https://esieabot.readthedocs.io/en/latest/
Project documentation conf.py file: https://gitlab.esiea.fr/esieabot/esieabot-docs/rtd/-/blob/main/docs/source/conf.py

#### Error Logs/Results

image

image

#### Expected Results

Is there a way to display all available languages (including the one that is currently used) like before? On my screenshots, I only see French option when I am reading in English. In French version, I can only see English version.

### Environment Info

- Python Version: 3.12
- Sphinx Version: 8.3.1
- RTD Theme Version: latest (3.0.1)

humitos:
Closes #1621

### DIFF
--- a/sphinx_rtd_theme/static/js/versions.js_t
+++ b/sphinx_rtd_theme/static/js/versions.js_t
@@ -8,7 +8,7 @@ if (themeFlyoutDisplay === "attached") {
       return "";
     }
 
-    // Prepend the current language to the options on the selector
+    // Insert the current language to the options on the selector
     let languages = config.projects.translations.concat(config.projects.current);
     languages = languages.sort((a, b) => a.language.name.localeCompare(b.language.name));
 
--- a/sphinx_rtd_theme/static/js/versions.js_t
+++ b/sphinx_rtd_theme/static/js/versions.js_t
@@ -8,10 +8,14 @@ if (themeFlyoutDisplay === "attached") {
       return "";
     }
 
+    // Prepend the current language to the options on the selector
+    let languages = config.projects.translations.concat(config.projects.current);
+    languages = languages.sort((a, b) => a.language.name.localeCompare(b.language.name));
+
     const languagesHTML = `
       <dl>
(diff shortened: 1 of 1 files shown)


sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -8,10 +8,14 @@ if (themeFlyoutDisplay === "attached") {
       return "";
     }
 
+    // Prepend the current language to the options on the selector
+    let languages = config.projects.translations.concat(config.projects.current);
+    languages = languages.sort((a, b) => a.language.name.localeCompare(b.language.name));
```
agjohnson, line 13: I have a feeling `sort` is not the correct operation here. I'm guessing this might randomly sort the list of locales after the first (current) locale. Using sort to only target one element in the list will resort/unsort/? the rest of the elements.

Is there a reason you can't use `insert(0, ..)` here?

humitos: Meh, the comment is wrong, but the code is correct.

We want to sort the languages alphabetically here, no matter where we insert the current project.

sphinx_rtd_theme/static/js/versions.js_t
```diff
@@ -8,10 +8,14 @@ if (themeFlyoutDisplay === "attached") {
       return "";
     }
 
+    // Prepend the current language to the options on the selector
```
humitos, line 11:
```suggestion
    // Insert the current language to the options on the selector
```

agjohnson (approved):

PR #1635, opened 2024-12-04
Document settings for source code links

abitrolly:
Fixes https://github.com/readthedocs/sphinx_rtd_theme/issues/465


abitrolly:
Ouch. There is also https://docs.readthedocs.io/en/stable/guides/edit-source-links-sphinx.html

Perhaps it should be copied here.

PR #1640, opened 2025-02-17
Run tests and build docs with Sphinx 8.2

humitos:


humitos:
All the tests should pass once 8.2.0 is released. Now, it's fails when installing it because it doesn't find it. However, we are testing against `master` branch and tests are passing 👍🏼

humitos:
8.2.0 was released already. Triggering tests again.

humitos:
It seems that all the tests are passing properly with the new version of Sphinx, so we should be ready to merge this PR.

ericholscher (approved):

.circleci/config.yml
```diff
@@ -64,28 +64,39 @@ jobs:
...
-          sphinx-version: "60,61,62,70,71,72,73,74,80,latest,dev"
+          sphinx-version: "60,61,62,70,71,72,73,74,80"
   py311:
     docker:
       - image: 'cimg/python:3.11'
     steps:
       - run-tox:
           version: py311
-          sphinx-version: "72,73,74,80,latest,dev"
+          sphinx-version: "72,73,74,80,81,82,latest,dev"
```
mustafacryptolife, line 74: E

mustafacryptolife, line 74: E

mustafacryptolife: []()

mustafacryptolife (commented):
[]()

PR #1686, opened 2026-04-19
Enable keyboard navigation

abitrolly:


hugovk:
This was brought up in https://github.com/sphinx-doc/sphinx/issues/12227 but @humitos said in part:

> So, we won't add `navigation_with_keys = True` to the project documentation's `conf.py` file and we will leave this decision to the documentation authors as mentioned on this issue as well. They should be the responsible about enabling/disabling this setting.

https://github.com/sphinx-doc/sphinx/issues/12227#issuecomment-2042178181

I've given some reasons not to enable this at https://github.com/python/docs-community/issues/199#issuecomment-5096122878.

